PR #21, opened 2021-07-13
Drop class name analogies favoring technical names

rvhonorato:
This is a boring PR to make the code more accessible.

I do believe analogy is a great technique to match unrelated objects in order facilitate understanding and it is seen in many large projects out there in the wild. However me and @joaomcteixeira are getting lost in the ones that were present here since they were often actually quite metaphoric. 

Additionally, since we aim the project to have a wide reach, we need to also consider code accessibility and making it more boring can increase its readability. There will still be plenty of room for creative expression later on :)

With this PR I also linted to code to comply with [E501](https://www.flake8rules.com/rules/E501.html).
In this branch there are no tests yet, so I tested it by running the examples and checking if things were working.

### DIFF
--- a/haddock/defaults.py
+++ b/haddock/defaults.py
@@ -38,108 +38,11 @@ class Default:
     LINK_FILE = data_path / "toppar/protein-allhdg5-4-noter.link"
 
     TRANSLATION_VECTORS = {
-        "trans_vector_0": (data_path /
-                           "toppar/initial_positions/trans_vector_0"),
-        "trans_vector_1": (data_path /
-                           "toppar/initial_positions/trans_vector_1"),
-        "trans_vector_2": (data_path /
-                           "toppar/initial_positions/trans_vector_2"),
-        "trans_vector_3": (data_path /
-                           "toppar/initial_positions/trans_vector_3"),
-        "trans_vector_4": (data_path /
-                           "toppar/initial_positions/trans_vector_4"),
-        "trans_vector_5": (data_path /
-                           "toppar/initial_positions/trans_vector_5"),
-        "trans_vector_6": (data_path /
-                           "toppar/initial_positions/trans_vector_6"),
-        "trans_vector_7": (data_path /
-                           "toppar/initial_positions/trans_vector_7"),
-        "trans_vector_8": (data_path /
-                           "toppar/initial_positions/trans_vector_8"),
-        "trans_vector_9": (data_path /
-                           "toppar/initial_positions/trans_vector_9"),
-        "trans_vector_10": (data_path /
-                            "toppar/initial_positions/trans_vector_10"),
-        "trans_vector_11": (data_path /
-                            "toppar/initial_positions/trans_vector_11"),
-        "trans_vector_12": (data_path /
-                            "toppar/initial_positions/trans_vector_12"),
-        "trans_vector_13": (data_path /
-                            "toppar/initial_positions/trans_vector_13"),
-        "trans_vector_14": (data_path /
-                            "toppar/initial_positions/trans_vector_14"),
-        "trans_vector_15": (data_path /
-                            "toppar/initial_positions/trans_vector_15"),
-        "trans_vector_16": (data_path /
-                            "toppar/initial_positions/trans_vector_16"),
-        "trans_vector_17": (data_path /
-                            "toppar/initial_positions/trans_vector_17"),
-        "trans_vector_18": (data_path /
-                            "toppar/initial_positions/trans_vector_18"),
-        "trans_vector_19": (data_path /
-                            "toppar/initial_positions/trans_vector_19"),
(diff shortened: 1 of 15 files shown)


haddock/defaults.py
```diff
@@ -37,57 +38,108 @@ class Default:
...
-        "trans_vector_42": data_path / "toppar/initial_positions/trans_vector_42",
-        "trans_vector_43": data_path / "toppar/initial_positions/trans_vector_43",
-        "trans_vector_44": data_path / "toppar/initial_positions/trans_vector_44",
-        "trans_vector_45": data_path / "toppar/initial_positions/trans_vector_45",
-        "trans_vector_46": data_path / "toppar/initial_positions/trans_vector_46",
-        "trans_vector_47": data_path / "toppar/initial_positions/trans_vector_47",
-        "trans_vector_48": data_path / "toppar/initial_positions/trans_vector_48",
-        "trans_vector_49": data_path / "toppar/initial_positions/trans_vector_49",
-        "trans_vector_50": data_path / "toppar/initial_positions/trans_vector_50"
+        "trans_vector_0": (data_path /
```
brianjimenez, line 41: This is a classic example where I would not personally follow strict 80-lenght line except if committing a major number of lines is intended in order to inflate statistics.

haddock/modules/sampling/lightdock.py
```diff
@@ -27,34 +27,43 @@ def run(self, module_information):
         self.patch_defaults(module_information)
 
         # Get the models generated in previous step
-        models_to_score = [p for p in self.previous_io.output if p.file_type == Format.PDB]
```
brianjimenez, line 30: List comprehension is faster and IMHO more readable here.

haddock/modules/topology/default.py
```diff
@@ -56,33 +56,50 @@ def run(self, module_information):
...
             logger.info("CNS engine has finished")
 
-            # Check for generated output, fail it not all expected files are found
+            # Check for generated output, fail it not all expected files
+            #  are found
             expected = []
             not_found = []
             for model in models:
                 model_name = model.stem
-                if not (processed_pdb := self.path / f"{model_name}_haddock.{Format.PDB}").is_file():
```
brianjimenez, line 77: Why this? Using modern Python and walrus operator is not a crime...

haddock/ontology.py
```diff
@@ -29,7 +30,9 @@ def __init__(self, file_name, file_type, path='.'):
         self.path = str(Path(path).parent.absolute())
 
     def __repr__(self):
-        return f"[{self.file_type}|{self.created}] {Path(self.path) / self.file_name}"
+        rep = (f"[{self.file_type}|{self.created}] "
```
brianjimenez, line 33: Less efficient and unnecessary.

haddock/workflow.py
```diff
@@ -0,0 +1,98 @@
...
+        # Create a recipe from a TOML file
+        self.recipe = Recipe(path)
+
+    def run(self):
+        """High level workflow composer"""
+        for step in self.recipe.steps[self.start:]:
+            step.execute()
+
+
+class Recipe:
```
brianjimenez, line 33: If not using metaphors anymore, maybe there is a better name for this class too?

brianjimenez (commented):
Boring review too. I am not a fan of strict linting, specially of line-length rules, but your call. List comprehensions and walrus operator are more pythonic options though.

haddock/modules/scoring/default.py
```diff
@@ -51,11 +56,21 @@ def run(self, module_information):
         jobs = []
 
         # Get the models generated in previous step
-        models_to_score = [p for p in self.previous_io.output if p.file_type == Format.PDB]
+        models_to_score = []
```
joaomcteixeira, line 59: same as before.
if you want to be purist, these logic should be some kind of method of function. but for now just keep the list comprehension

joaomcteixeira (commented):
Just made general comments. Nothing critical. Somethings I personally dislike, but they are not mistakes either.

haddock/workflow.py
```diff
@@ -0,0 +1,98 @@
...
+                    self.steps.append(Step(stage,
+                                           num_stage,
+                                           content["stage"][stage],
+                                           recipe_path.parent))
+            except RecipeError as re:
+                logger.error(f"Error found while parsing course {stage}")
+                raise HaddockError from re
+
+
+class Step:
```
joaomcteixeira, line 67: Needs documentation

joaomcteixeira:
However, despite we change the class names to more technical ones we should keep the nice HADDOCK-related interface for the users. Phrases of sailors are nice :boat:

haddock/defaults.py
rvhonorato: ooh this is great @joaomcteixeira, thanks!

PR #22, opened 2021-07-14
Merging beta1 to master

joaomcteixeira:
Hi  @amjjbonvin @rvhonorato @brianjimenez 

I am starting a draft pull request so that we can prepare `beta1` to merge it to the `master` branch. The current `beta1` state can't be automatically merge to `master` lest try to work that around first.


rvhonorato:
what if you rename the `master` to `old` and them merge? :)

joaomcteixeira:
okay. if that is so I will rename to `main` instead to comply with the ethical concerns avoiding `master/slave` nomenclature.

PR #24, opened 2021-07-16
data_path not being found

joaomcteixeira:
After #21 I got this error when running `defaults.py`.

`data_path` was not being found within dictionary comprehension in `defaults.py`.

```bash
File "/home/joao/github/haddock3/haddock/defaults.py", line 42, in <dictcomp>
    f"trans_vector_{i}": Path(data_path, 'toppar', 'initial_positions', f'trans_vector_{i}')
    NameError: name 'data_path' is not defined
```

reverting to a for-loop solves the issue. But I am really puzzled by this issue. I don't understand why `data_path` was not being found inside the dictionary comprehension. Any ideas?

### DIFF
--- a/haddock/defaults.py
+++ b/haddock/defaults.py
@@ -36,7 +36,6 @@ class Default:
     TOPOLOGY_FILE = data_path / "toppar/haddock.top"
 
     LINK_FILE = data_path / "toppar/protein-allhdg5-4-noter.link"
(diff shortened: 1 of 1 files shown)


rvhonorato (approved):
looks ok!

joaomcteixeira:
Thanks. In the future I will refact that file. Likely no need for that class.

PR #25, opened 2021-07-16
Update installation instructions

joaomcteixeira:
Updates installation instructions after #21:

* updates commands to new `main` branch name
* more general `cns` executable name
* adds conda `yml`
* facilitates source env


bin/activate_haddock
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export HADDOCK3_NUM_CORES=4
 export HADDOCKPATH=`pwd`
 export PYTHONPATH=$PYTHONPATH:$HADDOCKPATH
 export PATH=$PATH:$HADDOCKPATH/bin
+export HADDOCK3_CNS_EXE="${HADDOCKPATH}/bin/cns/cns"
```
rvhonorato, line 7: is here not `bin/cns`?

requirements.yml
```diff
@@ -0,0 +1,13 @@
...
+    - defaults
+dependencies:
+    - python==3.8
+    - pip
+    - numpy
+    - toml
+    - jsonpickle
+    - pip:
+        - pdb-tools
+        - lightdock
```
rvhonorato, line 13: I would not list lightdock as a dependency, you don't need it to run haddock.

joaomcteixeira: It is listed as a dependency in the current version. I will not remove in this PR. We can discuss it in a separate issue.

bin/activate_haddock
joaomcteixeira: `bin/cns/cns` is created during the steps provided in the installation, see `README`. In a future PR I will change this as well. but for now we can keep that behaviour not to loose the scope. Here, I simply renamed the executable for MAC to a more general `cns` name. See `README`.

rvhonorato: fair point, it makes no sense to be `/bin/cns/cns`

rvhonorato (commented):
Still think we should remove lightdock from the dependencies; will all the supported modules be part of dependencies? what if next we add PatchDock, maybe ZDOCK, or even GROMACS, do they also needs to be installed so that you can run haddock? I don't think it makes much sense, but up to you.

amjjbonvin:
Probably makes more sense to have it as an option

bin/activate_haddock
joaomcteixeira: You are right. it can be just on `bin`.

joaomcteixeira:
I understand your point. Well, at the current state those questions are difficult to answer in my opinion. But I understand your point. I will set `lightdock` as an optional dependency in another pull request and create the infrastructure to accommodate other optional dependencies as well. But I won't change them now in this PR not to mix too many things. :+1:

PR #26, opened 2021-07-16
renames flavour to mode missing files

joaomcteixeira:
Some names `flavour` remained after #21. This broke any runs where flavour was not the `default`, for example in `lightdock`.


rvhonorato (approved):
missed a few :)

PR #33, opened 2021-07-22
Examples cleaning and reorganizing

rvhonorato:
This PR cleans and re-organizes the examples folder, also removes overlapping ones.


joaomcteixeira:
files `scoring/scoring.tolm`, `full_scoring/scoring.tolm`, and `double_scoring/scorting.tolm` are different. I am not so deep into HADDOCK3 yet to evaluate this decision. Why do want to remove `double_scoring` and `scoring`?

rvhonorato:
They do not have any value as examples, scoring is a simplification of full_scoring and double_scoring and is a demonstration of the code capability. 

My understanding is that `examples` should guide the use of the software not as showcase of the code, we can do that in a later time with the development documentation.

amjjbonvin:
These are just workflow examples it I am correct.


- double scoring will run two rounds of EM+scoring

Issue #37, opened 2021-07-22
Implement working directory as a user parameter


rvhonorato:
Now when `haddock3` is executed, the folder structure is generated in the current directory, each folder being one step it would be more organized if a working directory could be passed as argument in the .toml file.
```
scoring/
|_ topology
|_ step_1
|_ step_2
```

to
```
scoring/
|__ run<name>
    |_ topology
    |_ step_1
    |_ step_2
```

PR #38, opened 2021-07-23
Refactor scoring pipeline

rvhonorato:
This PR is a small refactoring to the scoring pipeline that implements #37 and adds a Setup python module, in which we should centralize the validations related to the .toml file.

I also re-introduced the haddock-style `begin/` and `data/` directories.


src/haddock/setup.py
```diff
@@ -0,0 +1,71 @@
...
+        module_directory = Path(__file__).parent / 'modules'
+        if 'order' not in self.params['input']:
+            raise SetupError("Workflow does not specify the"
+                             " order of execution")
+
+        steps = copy.deepcopy(self.params['input']['order'])
+        for step in steps:
+            module_loc = module_directory / step / 'default.py'
+            if not module_loc.exists():
+                logger.warning(f'Module {step} is not valid, it will'
```
joaomcteixeira, line 32: I am adding some comments here while I study and work on your PR.
I think *excluding* a module from the execution when it does not exist is **NOT** a good approach because this can facilitate passing silent errors because of typos or mistakes. In my opinion if a module specified by the user is not present we should **abort** the execution, and report the error to the user.

src/haddock/setup.py
```diff
@@ -0,0 +1,71 @@
...
+        params = copy.deepcopy(self.params)
+        input_params = params['input']
+        project_dir = Path(input_params['project_dir'])
+        input_params['project_dir'] = project_dir.absolute()
+
+        data_dir = project_dir / 'data'
+        begin_dir = project_dir / 'begin'
+
+        if project_dir.exists():
+            logger.warning('This code is not production ready')
```
joaomcteixeira, line 54: This specific sentence should be part of a general warning to the user and triggered at start up, not as part of this class method.

src/haddock/setup.py
```diff
@@ -0,0 +1,71 @@
...
+import shutil
+import copy
+from pathlib import Path
+from haddock.error import SetupError
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class Setup:
```
joaomcteixeira, line 11: It might be okay to have a file named `setup.py`, though I dislike it a bit. But the `class Setup` is too general and does not specify what it does. But I understand your intentions. I would suggest a more pipeline based approach. Let me work on that.

src/haddock/setup.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def validate(self):
+        # ====
+        # Major validation function
+        # ===
+
+        # check if steps are valid
+        module_directory = Path(__file__).parent / 'modules'
+        if 'order' not in self.params['input']:
+            raise SetupError("Workflow does not specify the"
```
joaomcteixeira, line 25: If we want to be specific, here should be `ConfigError`.

src/haddock/setup.py
```diff
@@ -0,0 +1,71 @@
...
+        data_dir.mkdir()
+
+        for mol_identifier in input_params['molecules']:
+            input_mol = Path(input_params['molecules'][mol_identifier])
+            shutil.copy(input_mol, data_dir)
+
+            begin_mol = (begin_dir / f'{mol_identifier}.pdb').absolute()
+            shutil.copy(input_mol, begin_mol)
+
+            input_params['molecules'][mol_identifier] = begin_mol
```
joaomcteixeira, line 69: How do you keep the link between the new `begin_mol` and the original `mol_identifier` path that is now in `data_dir`?

joaomcteixeira:
Adding my suggestions at: https://github.com/rvhonorato/haddock3/pull/1

joaomcteixeira:
Okay. I think I finished reviewing your PR; my reviews and suggestions at: https://github.com/rvhonorato/haddock3/pull/1, if you accept them they will become part of this PR automatically.

It is a great idea @rvhonorato to firstly check the configuration file and prepare the files and folders for run. I like that idea. My suggestions in your branch are basically pythonic and architecture refactoring. I dislike the OOP-everywhere paradigm. It might be worthwhile in other languages but in python not really. It is useful where it is useful, but creates a lot of headaches otherwise. I haven't implemented any tests, but you'll see tests are easy to write with the changes I propose.

You'll notice the new folder `gear/`. We can call it differently if you wish. But to my experience, I am organizing my code in three main folders: `clis/`, `libs/` and `gear/`. `libs` have all the general and abstract implementations, those that can serve anywhere in the project. While `gear/` is dedicated to specific logic to serve specific implementations. To my experience, this gives very nice results. I rename `haddock.setup.py` because `setup.py` has a very specific meaning in the python world, and it may confuse other developers or us in the future.

Let me know,
cheers!

src/haddock/workflow.py
```diff
@@ -2,96 +2,95 @@
...
-            raise HaddockError(f"Recipe {recipe_path} does not exist or cannot"
-                               " be opened")
+    """The WorkflowManager reads the workflow and executes them."""
+    def __init__(self, workflow_params, start=0):
         self.start = start
         # Save current working path
-        self.workspace = path.parent.absolute()
-        # Create a recipe from a TOML file
-        self.recipe = Recipe(path)
+        self.cwd = workflow_params['input']['project_dir']
```
joaomcteixeira, line 17: I haven't find any place where this attribute is used.

src/haddock/setup.py
rvhonorato: Think so too, added this here just to develop things, will ammend it.

src/haddock/setup.py
rvhonorato: You don't, the data is there just for safekeeping the pre-processed inputs, it's how its done in production (2.4)

rvhonorato:
Thanks for the review and the edits @joaomcteixeira !

PR #41, opened 2021-07-26
Re-implement rigid body docking

rvhonorato:
This PR re-implements the rigid body docking feature as the `rigidbody` module in the new architecture. Its `cns/` folder contains only the protocols that are needed for it.


amjjbonvin:
Probably indeed CNS scripts in this pull - but we can clean later. E.g. build missing is only used when creating the topology / start structures. Should be part of a separate module (e.g. gentop)

rvhonorato:
> Probably indeed CNS scripts in this pull - but we can clean later. E.g. build missing is only used when creating the topology / start structures. Should be part of a separate module (e.g. gentop)

Yes the issue about it is #42 and the topology is already its own module.

joaomcteixeira:
Please give me some time to inspect this PR

joaomcteixeira:
Just looking very quickly now:
Why are you uploading again the `trans_vector_#` files if they are already on `data/toppar/initial_positions` ?

rvhonorato:
Quick screenshot of a superimposition of 100 models to show that the module is working and using restraints:

Screenshot 2021-07-27 at 13 57 46

rvhonorato:
> Why are you uploading again the `trans_vector_#` files if they are already on `data/toppar/initial_positions` ?

These are also CNS scripts, so I moved them to this module's `cns/`

src/haddock/modules/topology/default.py
```diff
@@ -47,11 +77,13 @@ def run(self, module_information):
 
             # Split models
             logger.info(f"Split models if needed for {step_molecule_path}")
-            models = sorted(PDBFactory.split_ensemble(step_molecule_path))
+            ens = PDBFactory.split_ensemble(step_molecule_path)
```
joaomcteixeira, line 80: this is a very minor comment. But in general I would avoid creating this temporary variables. It creates confusion to the reader. If `ens` is not going to be used anywhere else, so use `sorted()` directly without creating the variable. If the temporary variable is really needed, use:

```python
_ = do_something()
real_variable = do_other(_)
```

src/haddock/modules/topology/default.py
```diff
@@ -72,43 +104,42 @@ def run(self, module_information):
...
+        if not_found:
+            self.finish_with_error("Several files were not generated:"
+                                   f" {not_found}")
+
+        # Save module information
+        io = ModuleIO()
+        for model in models:
+            io.add(PDBFile(model))
+        io.add(expected, "o")
+        io.save(self.path)
```
joaomcteixeira, line 142: This block before was inside the `for` loop. Now it is outside. Is this intended?

src/haddock/modules/topology/default.py
```diff
@@ -5,16 +5,45 @@
...
 from haddock.error import StepError
 from haddock.defaults import TOPOLOGY_PATH
 
-
 logger = logging.getLogger(__name__)
 
 
+def generate_topology(input_pdb, step_path, recipe_str, defaults,
+                      protonation=None):
+    """Generate a HADDOCK topology file from input_pdb"""
```
joaomcteixeira, line 20: I would suggest we start working on nice docstrings from scratch. I know this function was already implemented somewhere else but since we are moving it to another file lets use the opportunity to write proper docstrings. What is `input_put`, `step_path`, `recipe_str`, `defaults`, and `protonation`?

joaomcteixeira:
Following today's conversation, `rigidbody/cns` has now 98 CNS scripts. Some script names there point to very general purposes, for example the `centroids_*.cns` family. My question is whether these are really SPECIFIC for rigid body docking, or we can already think of placing them in a module-independent `cns` dedicated folder.

amjjbonvin:
Probably best if I take a look - cannot think so many are required just for the rigid body docking

rvhonorato:
> Following today's conversation, `rigidbody/cns` has now 98 CNS scripts.

45 + translation vectors (approximately). Keep in mind there are nested dependencies, the first level account for ~32

src/haddock/modules/topology/default.py
rvhonorato: yes, before it was generating the topology for the models into an assembly, now it does both the models from the assembly and different inputs.

joaomcteixeira:
Can't the nested dependencies by pulled from another folder? Maybe via a bridge python file?

src/haddock/modules/topology/default.py
rvhonorato: We should probably decide on a format and implement it in the whole project, lets save it for another issue.

rvhonorato:
> Its cns/ folder contains only the protocols that are needed for it.

What I mean with the nested dependencies is that: cns scripts can call another cns scripts that can call another cns script that can call another cns script

rvhonorato:
> Probably best if I take a look - cannot think so many are required just for the rigid body docking

There are not 98, I replied to it in a comment above.

joaomcteixeira (commented):
@rvhonorato plase have a look at https://github.com/rvhonorato/haddock3/pull/2 before merging.

As we discussed today, we have a roadmap to change some things here. But if you prefer to merge as it is now and than work on what we discussed with @amjjbonvin, perfect for me.

src/haddock/modules/topology/default.py
joaomcteixeira: I agree. I see these functions are very similar if not equal.

src/haddock/modules/topology/default.py
joaomcteixeira: Is is able to continue from a half-run run? If you tested it and it works, perfect.

PR #43, opened 2021-07-27
Add tool to identify CNS script dependencies

### LINKED ISSUE #42: Re-implement CNS-dependency finder script
In version alpha we had part of the code that would scan the `.cns` script and return its dependency tree up to 5 levels of complexity, this was used to compose the `.inp` script that would be executed by CNS, adding the script to relevant tags RUN, @@, etc.

 The dependency tree can get quite complex since there are many nested calls:

```
mainscript.cns
  |_ script_A.cns
  |_ script_B.cns
    |_ script_B1.cns
  |_ script_C.cns
    |_ script_C1.cns
      |_ script_C1.1.cns
        |_ script_C1.1.1
      |_ script_C1.2.cns
    |_ script C2.cns 
```
      
This part of the code is missing in the `main` branch, we should re-implement it as a tool to help us remove unnecessary CNS scripts from the core modules.

rvhonorato:
This PR implements a tool to help us clean the CNS dependencies in the HADDOCK core modules and closes #42. It takes the main CNS script as input and the production `protocols` directory; Ex:

```
$  python tools/cns_dependency_tree.py src/haddock/modules/topology/cns/generate-topology.cns /Users/rodrigo/repos/haddock24/protocols 
[27/07/2021 11:27:21] L41 INFO - Protocols directory: /Users/rodrigo/repos/haddock24/protocols
[27/07/2021 11:27:21] L42 INFO - Retrieving dependencies of /Users/rodrigo/repos/haddock3/src/haddock/modules/topology/cns/generate-topology.cns
[27/07/2021 11:27:21] L47 INFO - First level dependencies: 9
[27/07/2021 11:27:21] L49 INFO - Looking for deeper dependencies
[27/07/2021 11:27:21] L76 INFO - Total dependencies found: 10
[27/07/2021 11:27:21] L78 INFO - generate-topology.cns depends on: patch-types-cg.cns patch-bb-cg.cns build-missing.cns prot_break.cns dna_break.cns covalheme.cns coval-ace-cys.cns auto-his.cns flex_segment_back.cns flexauto-neighbors.cns
```

This script is not very pretty but we can work on it later. :)

### DIFF
--- a/tools/cns_dependency_tree.py
+++ b/tools/cns_dependency_tree.py
@@ -1,0 +1,80 @@
+from pathlib import Path
+import re
+import sys
+import argparse
+import logging
+
+logging.basicConfig(level="DEBUG",
+                    format=("[%(asctime)s] L%(lineno)d %(levelname)s - %(message)s"),
+                    datefmt="%d/%m/%Y %H:%M:%S")
+
+
+def retrieve_dependency(cns_script):
+    """Read a CNS script and identify its dependencies."""
+    dep_list = []
+    marker_regex = r"@RUN:(.*\.cns|.*\.inp)"
+    with open(cns_script, 'r') as cns_fh:
+        for line in cns_fh.readlines():
+            match = re.findall(marker_regex, line)
+            if match:
+                dependency = match[0].split('/')[-1]
+                dep_list.append(dependency)
+    return dep_list
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='')
+    parser.add_argument('cns_script')
+    parser.add_argument('protocols_folder')
+    args = parser.parse_args()
+
+    protocol_dir = Path(args.protocols_folder)
+    if not protocol_dir.exists():
+        logging.error(f"{protocol_dir} does not exist.")
+        sys.exit()
+
+    main_cns_script = Path(args.cns_script)
+    if not main_cns_script.exists():
+        logging.error(f"{main_cns_script} does not exist")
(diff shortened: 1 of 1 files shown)


joaomcteixeira:
That looks a nice idea. But we should make it part of the haddock CLI tools, not a simple script. Let me work on that since I pulled #28 .

joaomcteixeira:
Question: why the script needs the `/Users/rodrigo/repos/haddock24/protocols` ? Is that because the `haddock24/protocols` is planned to be transferred to haddock3 but it is not yet here?

Will this dependency on HADDOCK24 disappear after #41 ? If that is so, maybe that second argument can be a fixed folder within haddock3 folder structure.


It would be nice to have the output also as you suggested on #42 :wink: , looks much prettier.

rvhonorato:
> Question: why the script needs the `/Users/rodrigo/repos/haddock24/protocols` ?

Because this is where the CNS protocols are, haddock3 is building on top of it.

> Will this dependency on HADDOCK24 disappear after #41 ? If that is so, maybe that second argument can be a fixed folder within haddock3 folder structure.

I'm afraid we will only keep relying on the production version (either 2.4 or 2.5) for a long time since it has many features that need to be implemented here; both ones that are already there as well as other that are being implement continuously.

amjjbonvin:
We should start a haddock3 CNS protocol directory

rvhonorato:
Closing this one since I already used this script to implement the core modules, we don't need to have it in the repository anymore.

PR #50, opened 2021-07-29
Implement the semi flexible refinement module (`flexref`)

rvhonorato:
This PR implements the semi flexible refinement module `flexref`. 

Screenshot with yellow = result from `rigidbody`, green = after `flexref`
Screenshot 2021-07-29 at 14 58 07


src/haddock/modules/topology/default.py
```diff
@@ -125,11 +125,11 @@ def run(self, module_information):
             if not processed_topology.is_file():
                 not_found.append(processed_topology.name)
             topology = TopologyFile(processed_topology,
-                                    path=(self.path / TOPOLOGY_PATH))
+                                    path=self.path)
             expected.append(topology)
             expected.append(PDBFile(processed_pdb,
                                     topology,
-                                    path=(self.path / TOPOLOGY_PATH)))
```
joaomcteixeira, line 132: Are you correcting a previous bug here?

src/haddock/ontology.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, file_name, file_type, path='.'):
         self.created = datetime.datetime.now().isoformat(' ', 'seconds')
         self.file_name = Path(file_name).name
         self.file_type = file_type
-        self.path = str(Path(path).parent.absolute())
```
joaomcteixeira, line 30: Is this a bug correction?

src/haddock/modules/__init__.py
```diff
@@ -63,8 +63,7 @@ def _load_previous_io(self):
 
     def previous_path(self):
         if self.order > 1:
-            return (self.path.resolve().parent.absolute() /
-                    f"{MODULE_PATH_NAME}{self.order-1}")
+            return (self.path.resolve().parent.absolute() / self.stream['input']['order'][self.order-1])
```
joaomcteixeira, line 66: Nano detail:

```python
return self.path.resolve().parent / self.stream['input']['order'][self.order - 1]
```

joaomcteixeira (commented):
There is nothing I can add or comment on the CNS part right now.

Other comments:

* I disagree with renaming `rigid_body_docking` to `simple_docking` in the examples. The former name is very specific and tells what is happening. The second is very generic.
* I think the example of "flexref" should the a NEW example folder and not an addition to the previous one. It is better to have 10 examples one adding a different thing to the next and having one single example doing 10 things. IMHO
* I left also some inline comments on parts that did not match this PR in my opinion. Were those bugs that needed to be corrected? Some changes are difficult to follow, why changed `rigidbody.defaults` to add flexref, was it something critical missing?

I haven't tested it yet, sorry, but if if the example is halted after "rigidbody", is it possible to resume the run directly to "flexref" ? In other words, is ModuleIO compatible between "rigidbody" and "flexref"? I hope I am not mistaking the concepts.

I don't see any critical errors apart from my comments. On you hands to act and merge. Good work!

src/haddock/modules/topology/default.py
rvhonorato: yep

src/haddock/ontology.py
rvhonorato: yep

rvhonorato:
>     * I disagree with renaming `rigid_body_docking` to `simple_docking` in the examples. The former name is very specific and tells what is happening. The second is very generic.
> 
>     * I think the example of "flexref" should the a NEW example folder and not an addition to the previous one. It is better to have 10 examples one adding a different thing to the next and having one single example doing 10 things. IMHO

So far none of the examples are valid, they are not real applications but simply demonstrations of the current capability of the code. What purpose will it serve to simply have a rigid-body sampling that does not even output the haddock-scores?

My goal is to have one complete example that shows both how-to-use and what can be done with haddock3, this small conceptual cases can go over to become integration tests or part of the developer documentation.

rvhonorato:
> I haven't tested it yet, sorry, but if if the example is halted after "rigidbody", is it possible to resume the run directly to "flexref" ? In other words, is ModuleIO compatible between "rigidbody" and "flexref"? I hope I am not mistaking the concepts.

The module is compatible but there's no resuming, not sure if I understand what you mean.

amjjbonvin:
Probably the question is: What happens if a run is for some reason stopped before completion?
Can we restart from the current state of the run and continue?

> =The module is compatible but there's no resuming, not sure if I understand what you mean.
>

rvhonorato:
> Probably the question is: What happens if a run is for some reason stopped before completion? Can we restart from the current state of the run and continue?
> […](#)
> =The module is compatible but there's no resuming, not sure if I understand what you mean.

Not implemented, yet!

PR #56, opened 2021-08-02
Add `emref` module

rvhonorato:
This PR adds the energy minimization refinement `energyminref`.


amjjbonvin (approved):
Will test once everything is in place. 

A few comments:

1) I would rather call it for simplicity: `EMref`

2) The order seems weird. EMref should come after flexref (and either waterref or EMref, but not both)

joaomcteixeira (commented):
I see no issues regarding the python modules.

Regarding the `boxtp20.pdb`, does it need to be accompanied by the respective CHARMM License or reference to the original repository or paper? If so, you can add a README file in the `cns/` folder.

Everything else looks fine for me, for as far as I can evaluate.

amjjbonvin:
That box comes out of ARIA if I am correct, and probably even Xplor/CNS originally.
I would not worry about it

rvhonorato:
>     1. I would rather call it for simplicity: `EMref`

Currently there's no capitalization support, so it just becomes `emref`, should we add support for capitalization? This would mean that `emref` can be a different module from `EMref`.

Should we also enforce a maximum number of character in the module names? What should be the maximum?

>     2. The order seems weird. EMref should come after flexref (and either waterref or EMref, but not both)

The example in itself does not make any sense yet, but I removed `emref` from it.

amjjbonvin:
> Currently there's no capitalization support, so it just becomes emref, should we add support for capitalization? This would mean that emref can be a different module from EMref.
> 

Do you mean Python does not distinguish capital letters?

Then we can keep emref (although EMref looks nicer).

And we should rename waterref to MDref (with two options for the solvent to use: water or DMSO

joaomcteixeira:
We can have CAPS-insensitive that's for sure. We can advocate for `EMref` instead of `emref`. but internally, all will be converted to `.lower()`.

PR #59, opened 2021-08-02
New modules structure and names

joaomcteixeira:
Addresses part of #53


joaomcteixeira:
@amjjbonvin @rvhonorato this PR is ready for review. There are a lot of file changes but they refer mostly to moving files between folders. Focus on `*.py` files for actual changes.

Ideas applied in this PR:

* reorganised `modules` according to [our discussion](https://github.com/haddocking/haddock3/discussions/53#discussioncomment-1094788).
  * HADDOCK3 modules are now inside category subfolders.
  * Instead of having a `run.py` inside the module's folder, I have moved the previous `default.py` to `__init__.py`. I think here it is safe to do this antipattern and have the construction logic inside `__init__.py`.
* modules can be specified in the configuration files without the need to specify their category.
  * For example, `emscoring` without the need to specify `scoring.emscoring`.
* `topology` separated into `topoaa` and `topocg` according to: https://github.com/haddocking/haddock3/discussions/53#discussioncomment-1130007
  * @rvhonorato , I also added the folders for `topocg` despite their are empty, but the folder is there
* execution folders are now prefixed by a number `00`, `01`, `02` , following their order of execution.
  * But, "[So could generate 00_cgtopo and 00_aatopo](https://github.com/haddocking/haddock3/discussions/53#discussioncomment-1130007)" is not yet fully implemented.

Initially, I thought to implement the whole #53 but that is too much for a single PR. I will continue that in another branch. The most relevant future change will be removing the `[stage]` layer from the configuration file and code.

rvhonorato:
> execution folders are now prefixed by a number `00`, `01`, `02` , following their order of execution.
> 
>     * But, "[So could generate 00_cgtopo and 00_aatopo](https://github.com/haddocking/haddock3/discussions/53#discussioncomment-1130007)" is not yet fully implemented.

Not sure if I understand, why have this?

amjjbonvin:
Do you mean both 00_cgtopo and 00_aatopo?

Both are needed in case CG is used since we need the aa topo to convert back to all atoms

rvhonorato:
After reading the discussions I think I understand it, if you are just numbering the modules by their order this is fine.

Topology must always be step0, but we can deal with the relationships once we come to it.

rvhonorato (approved):

PR #60, opened 2021-08-03
Example fix

rvhonorato:
The past PRs broken the example, this one fixes it.


joaomcteixeira:
This is actually a good opportunity to explain and enforce mandatory parameters. The problem was that sampling was not being provided, so the code crashed when trying to read that parameter. However a proper error message was not defined. We should make sure mandatory parameters for each haddock3 module are identified and reported to the user in case these are missing in the configuration file.

PR #62, opened 2021-08-03
Remove input layer from configuration file

### LINKED ISSUE #61: Warn the user about missing mandatory parameters
This is actually a good opportunity to explain and enforce mandatory parameters. The problem was that sampling was not being provided, so the code crashed when trying to read that parameter. However a proper error message was not defined. We should make sure mandatory parameters for each haddock3 module are identified and reported to the user in case these are missing in the configuration file.

_Originally posted by @joaomcteixeira in https://github.com/haddocking/haddock3/issues/60#issuecomment-891943689_

Also, if a mandatory parameter is missing from the configuration file, the default value should NOT be used. The user should be informed all the time.

joaomcteixeira:
Pull Request #38 introduced the concept of the `[input]` layer in the configuration file. However, we have seen that such a layer only adds complexity to the configuration `toml` file, thus adding burden to the user. In contrast, such separation is not necessary for the proper functioning of the python shell. It is true that removing the `input` layer forces general input parameters names to be unique and not collide with module names, but that is pretty feasible.

I run into this consideration while working on #53 and hence #59, which came in a discussion after #38.

Things that have been changed:
* config `toml` files do not require `[input]` parameter layer
* parameter `project_dir` was renamed to `run_dir`
* `[input.molecules]` are now a list of file paths. Because molecules were converted internally to `mol_i` where `i>0`, there is no need to force the user to write this. In the version proposed in this PR file paths are given in an ordered list which order can be directly rearranged without the need to update the `mol_i` identifier. 
  * if `molecules` is just one, a string can be given instead of a list of a single item.
* How can we track general mandatory parameters? A new `parameters.py` module was added to the `gear/` folder. A set of mandatory general parameters was defined. More definitions and related logic can be added in the future.
* Added a special context and decorator to raise `ConfigurationError` when a required key in the config file is not present. This should become less necessary as we improve the mandatory params definitions. Or, instead, it could become the only method to check for mandatory parameters instead of having them defined in a separate module. But for now, it also serves subkeys.
* lightdock module now works. Its functionality was broken after the latest improvements.

other changes:

* params related to `clustering.fcc` from `scoring.toml` were commented out because `fcc` is not yet implemented.
* minor pythonic changes here and there

It is a bad practice to address two questions in a pull request, but this also closes #61 

All examples run on my computer.


amjjbonvin:
Beware that the order might be important when passing files from one stage to another since not all models of one stage might be used in the next stage.

Or is this only for the initial setup and then internally between modules things are properly sorted / passed?

> [input.molecules] are now a list of file paths. Because molecules were converted internally to mol_i where i>0, there is no need to force the user to write this. In the version proposed in this PR file paths are given in an ordered list which order can be directly rearranged without the need to update the mol_i identifier.
> if molecules is just one, a string can be given instead of a list of a single item.

joaomcteixeira:
The current changes concern only the initial setup. The connection between modules is handled internally. After #38, settings such as `[stage.topology.molecules]` were moved to the `[input]` layer as `mol_1` , `mol_2`, etc. But when looking to the `input` layer definition in the `toml` file, one realises that the boilerplate can be reduced to a set of parameters living in the first layer of the config dictionary without the need to specifically define an `input` level.

So, this PR does **not** revert what was done in #38, but improves it by simplification.

With this PR it is also easier to report to the user which mandatory general parameters are missing in the configuration file. Thus addressing #61

rvhonorato (approved):
Looks right to me

joaomcteixeira:
Thanks @rvhonorato 
these will go straight to #59

PR #63, opened 2021-08-09
Removing `stage` key as a requirement for the configuration file

joaomcteixeira:
Continue implementations discussed in #53.
Here's the draft PR while I work on it. Comments and suggestions are very welcomed.
Cheers,

### DIFF
--- a/src/haddock/gear/prepare_run.py
+++ b/src/haddock/gear/prepare_run.py
@@ -62,7 +62,15 @@ def setup_run(workflow_path):
     remove_folder(params['run_dir'])
     create_begin_files(params)
 
-    return params
+    # get a dictionary without the general config keys
+    modules_params = remove_dict_keys(
+        params,
+        config_mandatory_general_parameters,
+        )
+
+    copy_ambig_files(modules_params, begin_dir)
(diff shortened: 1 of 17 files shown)


joaomcteixeira:
Done, ready to review.

The `stage` key is now not needed in the configuration file. Also, I decoupled some parts while opening space for some future implementations if needed. After doing this, I got the sensation that we can dispose even of the `order` parameter.

src/haddock/gear/prepare_run.py
```diff
@@ -75,6 +77,10 @@ def setup_run(workflow_path):
 
     # return the modules parameters and other parameters that may serve
     # the workflow
+
+
+    print(modules_params)
```
rvhonorato: these should be `logging`?

rvhonorato (changes requested):

rvhonorato (approved):

src/haddock/gear/prepare_run.py
joaomcteixeira: A bad print I left it behind,. sorry

PR #64, opened 2021-08-10
Handling defaults

joaomcteixeira:
Revisits how defaults parameters are stored and handled. Bellow the list of changes:

* defined the path to the default configuration file as a module global instead of being defined inside the `HaddockModule` `__init__.py` method.
  * the path to the default config is than defined as a keyword parameter
  * default toml files are now named `defaults.toml` in every method.
* In `BaseHaddockModule`, the concept `default_path` or `defaults` was renamed to `params` because what is being handled are parameters and *default* is just a temporal causality.
* `params` implemented as  a `@property` 
* Hence, the run-specific parameters are given at `.run()` and not at the class instantitation.
* all modules now update the params to the run-specific input parameters in `run()`
* Removed the `"params"` layer from the default configuration `toml` files.
* `BaseHaddockModule` is now an `ABC` to enforce implementation of the `run` method without the need of the `try:catch`.

After this PR, user-given parameters efficiently replace default values for all parameters and not only for those strictly required.


rvhonorato (approved):

joaomcteixeira:
Thx @rvhonorato . Yet it is best to merge this PR only after #67 is merged. I believe it is easier to modify this PR after #67 than the other way around. What do you think?

rvhonorato:
Think so too

joaomcteixeira:
@rvhonorato , I believe this is done. Can you have closed look at it and run the examples on your side to see if everything is working? Let me know if something is not clear.

rvhonorato:
All good, nice changes! Examples also working fine

PR #66, opened 2021-08-11
reviewing subprocesses

joaomcteixeira:
This PR reviews the multiprocessing implementations with the aim to simplify the code, make it cleaner and more modular. What was done:

* I found no reason for the `Engine` class to exist. The `Engine` class was only asserting the number of jobs against the number of cores and running the `Scheduler`. I removed the `Engine` class.
* created a separated and flexible function to parse the number of cores against the number of jobs (see `parse_ncores`) that can be used anywhere in the program and is not bound to a class. This function can be nicely parametrized according to the `ncores`, `njobs`, and `max_cpus`.
* The `__init__` of the `Scheduler` was refactored. Also, `num_processes` can now be used as a class attribute.
* The general class `Job` was slightly changed in the `run` method. @rvhonorato can you confirm everything is working properly? It was a simple change to remove the `if`
* Finally, the previous `paralell.py` module was directly into `haddock/`, I moved that logic to `libs.liparallel`. As we discussed in the meeting, to avoid having many files in the main folder.


joaomcteixeira:
I think this is also ready, after we merge #67 I will update this according to the `main` and done.

joaomcteixeira:
Open for review.

@rvhonorato test also the `fcc` example because I changed something in the `Job` class. If you can, test all examples please :-) `lightdock` will fail because of `path issues`, I will send a new PR correcting that, no worries.

Modifications:
* `parallel.py` now in `libs/`
* logic related to subprocessing now in `libss/
* removed `Engine` class
* added new general purpose `parse_ncores` function

rvhonorato:
@joaomcteixeira could you add a description to this PR?

joaomcteixeira:
> @joaomcteixeira could you add a description to this PR?

Completely, forgot about it. Done. See first comment in the thread.

rvhonorato (approved):
Sorry I took so long to have a look at this one!

Fine changes and everything running OK

joaomcteixeira:
Thanks @rvhonorato

Issue #32, opened 2021-07-22
Re-implement basic features


rvhonorato:
We dropped the `alpha` development branch in favor of a more robust code and development direction, however it currently does not have the basic features that were present in the `alpha`. 

This set us back quite a bit since we need to re-implement the features in the new code; to put it back in line I've split the basic modules between `implemented`, `tested` (covered by unit tests) and the workflows between `implemented` and `benchmarked`

* * *
## Core Modules Implemented

- [X] Topology generation
- [x] Rigid body sampling (`rigidbody`)
- [x] Semi-flexible refinement (`flexref`)
- [x] Water refinement (`mdref`)
- [x] Final EM (`emref`)
- [x] Analysis (`clustfcc`)
- [x] Analysis (`caprieval`)
- [X] Scoring

## Core Modules Tested

- [ ] Topology generation
- [ ] Rigid body sampling (`rigdock`)
- [ ] Semi-flexible refinement (`flexref`)
- [ ] Water refinement (`mdref`)
- [ ] Final EM (`emref`)
- [ ] Analysis (`clustfcc`)
- [ ] Analysis (`caprieval`)
- [ ] Scoring

***

## Core Workflows Implemented

- [x] Docking
- [X] Scoring

joaomcteixeira:
Very nice @rvhonorato Thanks for putting it in a task list form.
I believe @amjjbonvin had considered renaming it0, it1, and itw.

amjjbonvin:
> I believe @amjjbonvin <https://github.com/amjjbonvin> had considered renaming it0, it1, and itw.
> 
> 

Yes indeed! The old naming very much implies a sequential and fixed protocol.

I would rather name those:

- rigid-docking, eg.g rigdock 
- flexible refinement, e.g. flexref
- water refinement, e.g. water
- final EM, e.g. EMref

Then we can mix and match to create custom protocols

rvhonorato:
Change the names to match the new naming!

> - final EM, e.g. EMref

This name implies (at least for me) that this module could only be used in the last step of the workflow, is that it?

amjjbonvin:
No EMref is generic in principle - but it is currently the final stage in haddock2.4

rvhonorato:
Full docking possible again after #67  🥳

rvhonorato:
We wrap this one up with #177

PR #67, opened 2021-08-11
Implement `clustfcc` module

rvhonorato:
This PR implements the last pending core module after `beta1` migration (#32) by adding the `clustfcc` module.

For it to work in the workflow philosophy I also had to add a `seletopclust` "helper module" that handles the selection;

the `seletopclusts` takes two arguments, the ranking of the clusters to be selected and how many models of each. It uses the `pdb.score` taken from the `previous_io` to calculate the cluster and model rankings.

```toml
[seletopclust]
top_cluster = [1, 2, 3, 4] # will select the top4 clusters
top_models = 4 # will select the best 4 models inside each cluster
```

This effectively means we can now do things like:

`topology > rigid-body > fcc-clustering > select-top-cluster > flexible-refinement > md-refinement`

***

I also added `haddocking/fcc` as a submodule, pulling from the python3 branch to avoid code repetition.


joaomcteixeira (changes requested):
Apart from my little comments, if it works, it looks good. :+1:

src/haddock/engine.py
```diff
@@ -1,10 +1,38 @@
 """Running CNS scripts"""
 import subprocess
-from haddock.error import CNSRunningError
+from haddock.error import CNSRunningError, JobRunningError
```
joaomcteixeira, line 3: In #66 this file will be in `libs`, no problem. :+1:

src/haddock/engine.py
```diff
@@ -1,10 +1,38 @@
...
+
+        # capture output to stdout
+        cmd += f' {self.input}'
+        with open(self.output, 'w') as outf:
+            p = subprocess.Popen(shlex.split(cmd), 
+                                 stdout=outf,
+                                 close_fds=True)
+            out, error = p.communicate()
+        p.kill()
+        outf.close()
```
joaomcteixeira, line 30: you don't need `outf.close()` is already with `with`.

src/haddock/engine.py
```diff
@@ -1,10 +1,38 @@
...
+        self.executable = executable
+        self.args = args
+
+    def run(self):
+        if self.args:
+            cmd = f"{self.executable} {' '.join(map(str, self.args))}"
+            self.executable = shlex.split(cmd)
+
+        # capture output to stdout
+        cmd += f' {self.input}'
```
joaomcteixeira, line 23: I think here there's a bug if `args` are not given because `cmd` won't be defined.

rvhonorato: oh yeh, indeed!

joaomcteixeira (approved):
Look good :relaxed:

PR #72, opened 2021-08-13
Add third-party module `gdock`

rvhonorato:
This PR adds the unrequested third-party module [`gdock`](https://github.com/rvhonorato/gdock) that demonstrates the easiness of integration with the current haddock3 workflow;

`topology > gdock > cluster-fcc > select-top-cluster > flexible-refinement > md-refinement`

[`gdock`](https://github.com/rvhonorato/gdock) has an option to do a  "dry run" in which it just generates models and nothing else. By passing this option to the module we can generate a bunch of models with the genetic algorithm sampling routine and then pass them to the flexible-refinement and md-refinement. 🐙 

* * *

```
##############################################
#                                            #
#                 HADDOCK 3                  #
#                                            #
##############################################

Starting HADDOCK 3.0.beta-unreleased on 2021-08-13 10:14:00

Python 3.9.5 (default, May 18 2021, 12:31:01) 
[Clang 10.0.0 ]

[2021-08-13 10:14:59,771] haddock.libs.libutil:L53 WARNING - /Users/rodrigo/repos/haddock3/examples/thirdparty/gdock/run1 exists and it will be REMOVED!
[2021-08-13 10:14:59,812] haddock.workflow:L34 INFO - Reading instructions of [topoaa] step
[2021-08-13 10:14:59,812] haddock.workflow:L34 INFO - Reading instructions of [gdock] step
[2021-08-13 10:14:59,812] haddock.workflow:L34 INFO - Reading instructions of [clustfcc] step
[2021-08-13 10:14:59,812] haddock.workflow:L34 INFO - Reading instructions of [seletopclusts] step
[2021-08-13 10:14:59,812] haddock.workflow:L34 INFO - Reading instructions of [flexref] step
[2021-08-13 10:14:59,813] haddock.workflow:L34 INFO - Reading instructions of [mdref] step
[2021-08-13 10:14:59,860] haddock.modules.topology.topoaa:L56 INFO - Running [allatom] module
[2021-08-13 10:14:59,860] haddock.modules.topology.topoaa:L57 INFO - Generating topologies
[2021-08-13 10:14:59,861] haddock.modules.topology.topoaa:L66 INFO - 1 - /Users/rodrigo/repos/haddock3/examples/thirdparty/gdock/1a2k_l_u.pdb
[...]

PR #73, opened 2021-08-13
Move examples data to its own folder

rvhonorato:
From the suggestion of https://github.com/haddocking/haddock3/pull/72#pullrequestreview-729395995, to avoid data duplication in the examples.


joaomcteixeira (approved):
I haven't tested it. But if it works, it looks very good! :+1:

PR #74, opened 2021-08-13
Add `seletop` module

rvhonorato:
For haddock to work in the **classical** way we also need a model to select the top X models from a previous stage. In the production version the rigid-body generates 1000 models, then the top 200 are selected to the next stage.

This can be done in haddock3 with the `seletop` module; it gets all the models from the previous_io and ranks them according to their score and outputs a subset equal to `select`

I also changed the docking example to reflect the default haddock protocol.

### DIFF
--- a/examples/docking/docking.toml
+++ b/examples/docking/docking.toml
@@ -4,10 +4,10 @@
 # in which order the steps must be executed?
 order = ["topoaa",
          "rigidbody",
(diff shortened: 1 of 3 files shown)


joaomcteixeira:
I will look it it carefully after merging #64. This time this PR should be updated after #64 :relaxed:

PR #77, opened 2021-08-13
Add `seletop` module

rvhonorato:
For haddock to work in the **classical** way we also need a model to select the top X models from a previous stage. In the production version the rigid-body generates 1000 models, then the top 200 are selected to the next stage.

This can be done in haddock3 with the `seletop` module; it gets all the models from the previous_io and ranks them according to their score and outputs a subset equal to `select`


joaomcteixeira (changes requested):
`seletop.toml` must be renamed to `defaults.toml`

src/haddock/modules/analysis/seletop/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        logger.info("Running [seletop] module")
+
+        super().run(params)
+
+        # Get the models generated in previous step
+        if type(self.previous_io) == iter:
+            self.finish_with_error('This module cannot come after one'
+                                   ' that produced an iterable')
+
+        models_to_select = [(p, p.score) for p in self.previous_io.output if p.file_type == Format.PDB]
```
joaomcteixeira, line 34: No need to make the tuple. I believe this is cleaner
```python
models_to_select = [p for p in self.previous_io.output if p.file_type == Format.PDB]
models_to_select.sort(key=lambda x: x.score)

if len(models_to_select) < params['select']:
        logger.warning('Number of models to be selected is larger'
        ' than generated models, selecting ALL')

selected_models = models_to_select[:params['select']]
```

joaomcteixeira (changes requested):
Looks a very clean addition. Can you please change the lines I suggested and that is it, it will read much cleaner. Great job! Good team!

PR #78, opened 2021-08-13
Implement benchmarking suite

rvhonorato:
:construction:


rvhonorato:
the script to benchmark haddock3 success rate will be implemented elsewhere.

Issue #69, opened 2021-08-12
Make weights as a parameter to the `rigidbody` `flexref` and `mdref` 


joaomcteixeira:
This eventually should be a parameter. But it is best to make this an issue and addressed it after this PR and #64 are merged.

_Originally posted by @joaomcteixeira in https://github.com/haddocking/haddock3/pull/67#discussion_r687605504_

PR #81, opened 2021-08-20
Implements dedicated configuration reader

joaomcteixeira:
We have discussed about drop using `toml` as a configuration reader and implement our own configuration reader. Why?

* We decided HADDOCK3 should allow running the same module several times. Hence, it becomes awkward for the user to manually add integers to modules' name in case of repetition
* we drop the use of order in the configuration files, no longer needed

It is complex to implement a parser with the same capabilities as `toml` so here only the needed functionalities were added. Still, the parser is flexible. See docstring in `gear.config_reader`.

Please help me see if there is something I am missing. Previous simplifications of the configuration file where PR in #62 and #63

**Note** I added `emscoring` twice in the examples for proof-of-concept.

I am keeping this PR as draft so we can give it some thoughts before merging, but everyone is welcomed to comment, the code is working.

Here is an example of a possible configuration file:

```toml
# ====================================================================
# Rigid-body docking example

run_dir = "run1"

# starting molecules
# a multi-line list cannot have comments inside
# an empty line is mandatory after a multi line list
molecules =  [
    "../data/1a2k_l_u.pdb",
    "../data/1a2k_r_u.pdb"
    ]

[topoaa]
autohis = true # comments are allowed

[rigidbody] # also in headers
ambig = '../data/1a2k_ti.tbl'
sampling = 20

[seletopclusts]
top_cluster = [1, 2, 3, 4]
top_models = 4

[flexref]
ambig = '../data/1a2k_ti.tbl'
cool1_steps = 500

[seletopclusts]
top_cluster = [1, 2, 3, 4]
top_models = 4

[mdref]
ambig = '../data/1a2k_ti.tbl'
cool1_steps = 10

[clustfcc]
contact_distance_cutoff = 5.0
fraction_cutoff = 0.6
threshold = 4
```


joaomcteixeira:
Ready. Let's wait until @amjjbonvin returns to decide on merging. But meanwhile is ready for reviews / comments / suggestions.

rvhonorato:
Nothing to add on the parser, I think its a great implementation!

joaomcteixeira:
> However would there be a way in which the user could manipulate the order in which the module would be executed without cut-and-copy? Could we have order as an optional?

Sorry @rvhonorato i had missed your comment. Yes, that is possible to implement. However it would break the "repeated names" functionality. Also, I think it would just add too much possibilities. I think if we want to have repeated modules running without having to worry about numeric suffixes we must not use the `order`. What do you think? I am happy that you found positive the other possibilities.

joaomcteixeira:
I also added now an option for dictionary keys with the typical `[rigidbody.weights]`, as I am working now on #69

rvhonorato:
> > However would there be a way in which the user could manipulate the order in which the module would be executed without cut-and-copy? Could we have order as an optional?
> 
> Sorry @rvhonorato i had missed your comment. Yes, that is possible to implement. However it would break the "repeated names" functionality. Also, I think it would just add too much possibilities. I think if we want to have repeated modules running without having to worry about numeric suffixes we must not use the `order`. What do you think? I am happy that you found positive the other possibilities.

I also don't know what is best. I'm more inclined to solve this via a web-interface 😉

rvhonorato (approved):

amjjbonvin:
> However would there be a way in which the user could manipulate the order in which the module would be executed without cut-and-copy? Could we have order as an optional?
> 
You might end up with inconsistencies between the modules order and the order in the toml file.

Cut and paste is easy :-)

joaomcteixeira:
Since we are moving away from `toml` and implementing our own config reader, we should change configuration files extension also. Here are some suggestions:

* `.hd3`
* `.hdd`
* `.hcfg`
* `.h3c`

What are your suggestions or votes for the above options?

joaomcteixeira:
In Slack, people voted and we discussed `.cfg` is the best option.

PR #88, opened 2021-09-06
Reactivates `--restart` option

### LINKED ISSUE #75: parameter `--restart` likely not functioning properly - investigate

joaomcteixeira:
The `--restart` option was not working. Now it works.

* if `--restart` is given, the previous run folder is **not** deleted in case it existed.
* the "previous" module folder is obtained by the order parameter and not considering the last created folder in the series `00_`.

Closes #75

### DIFF
--- a/src/haddock/clis/cli.py
+++ b/src/haddock/clis/cli.py
@@ -6,7 +6,8 @@ from argparse import ArgumentTypeError
 from functools import partial
 
 from haddock.version import CURRENT_VERSION
-from haddock.libs.libutil import file_exists, non_negative_int
+from haddock.libs.libutil import file_exists
+from haddock.gear.restart_run import add_restart_arg
 
 
 # Command line interface parser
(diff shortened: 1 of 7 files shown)


rvhonorato (commented):
What if the .toml is edited and the order is different?

joaomcteixeira:
good point. let me have a look at it.

joaomcteixeira:
Yes, folder than get repeated. What I can do is to delete the folders from the `--restart` number.

joaomcteixeira:
Okay @rvhonorato , it now addresses what you said.

If `--restart` is given, the folder after (and included) the restart number will be deleted. For example if `--restart 2` the folder with `02_*`, `03_*`, etc, will be deleted from the `run_dir`. Nothing else will be touched in the run dir.

If `--restart` is not given, the whole `run dir` will be deleted as already implemented.

If `--restart 0` is given, folders from `00_` will be deleted but all the other initial files in the `run dir` will not.

Functionalities related to the `restart` implementation are in `gear.restart_run.py` module. The main code just calls its functionalities.

Give it a try and let me know. It was working on my side when swapping `mdref` and `flexref`.

rvhonorato:
> Give it a try and let me know. It was working on my side when swapping `mdref` and `flexref`.

What if you add/remove?

joaomcteixeira:
The same. If you define `--restart 2` the run will delete all folders after (and included) `02_*`, and keep the previous ones. And, the calculation will start from step 2. So you can add, remove, swap, steps after the `--restart` without problem.

Obviously, crazy things can happen,  if you change the steps **before** the `restart` option. The user has always to ensure that the modules can be piped. I think for now it is premature to raise any special error message if we detect that the previous folders were altered. Maybe it is best to see what happens when using the software more extensively before adding any safes for now?

rvhonorato (approved):

PR #89, opened 2021-09-07
ReOrganizes some files, folder, and code

joaomcteixeira:
This is a large PR but it's mostly concerns file moves and consequent update of the import statements.

* modules in `src.haddock` where moved to its corresponding place under `libs`, as discussed.
* the `cns` related python code was moved to `libcns`.
* a new `cns/` folder now has CNS-related files and **does not** contains any python. Hence, previous `data.toppar/` was moved to `cns.toppar`. In the future, modular CNS scripts can go here too.
* definitions were moved to `core/`: `exceptions.py`, `defaults.py`, `cns_paths.py`
* HADDOCK3 paths are now independent from modules (decoupled) and defined lib-wide under `__init__`
* path definitions pointing to CNS files were improved (see `core.cns_paths.py`)
* `PDBFactory` is now `libpdb`


rvhonorato (approved):

joaomcteixeira:
You can better see the result of this PR if you navigate the branch -> https://github.com/joaomcteixeira/haddock3/tree/libpdb/src/haddock

PR #91, opened 2021-09-14
Make `weights` configurable parameters

### LINKED ISSUE #69: Make weights as a parameter to the `rigidbody` `flexref` and `mdref`

joaomcteixeira:
Closes #69

### DIFF
--- a/examples/docking/docking.toml
+++ b/examples/docking/docking.toml
@@ -31,22 +31,23 @@ autohis = true
 ambig = '../data/1a2k_ti.tbl'
 sampling = 20
 
-[clustfcc]
-contact_distance_cutoff = 5.0
-fraction_cutoff = 0.6
-threshold = 4
+#[clustfcc]
+#contact_distance_cutoff = 5.0
+#fraction_cutoff = 0.6
+#threshold = 4
+#
+#[seletopclusts]
+#top_cluster = [1, 2, 3, 4] # will select the best one
+#top_models = 4 # will select the top4 models inside the cluster
 
-[seletopclusts]
-top_cluster = [1, 2, 3, 4] # will select the best one
-top_models = 4 # will select the top4 models inside the cluster
-
-[flexref]
-ambig = '../data/1a2k_ti.tbl'
-cool1_steps = 500
 
 [mdref]
 ambig = '../data/1a2k_ti.tbl'
-cool1_steps = 10
+cool1_steps = 2
+
+[flexref]
+ambig = '../data/1a2k_ti.tbl'
+cool1_steps = 50
 
 # ====================================================================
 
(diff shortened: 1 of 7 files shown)


examples/docking/docking.toml
```diff
@@ -31,22 +31,43 @@ autohis = true
...
+[flexref]
+ambig = '../data/1a2k_ti.tbl'
+cool1_steps = 50
+
+[flxref.weights]
+vdw = 1.0
+elec = 1.0
+desol = 1
+air = 0.1
+bsa = -0.01
```
rvhonorato, line 70: imo I think having this as a separate category adds a bit of complexity, these are parameters just like `cool1_steps` so why having them in `[flexref.weights]` and not in `[flexref]`?

joaomcteixeira: good question. I went for the simplest strategy to have `self.params['weights']`. But you raised a good point. Question: is it really clear that the parameter `air` refers to a *weight* ? Should we add a `w_` prefix for weight related parameters?

amjjbonvin:
That’s what we have in run.cns for haddock2.X

> good question. I went for the simplest strategy to have self.params['weights']. But you raised a good point. Question: is it really clear that the parameter air refers to a weight ? Should we add a w_ prefix for weight related parameters?

examples/docking/docking.toml
rvhonorato: Sorry this confusion is probably because of my messy implementation of putting the weights straight into the module. 

This values are not used by the CNS code to do the calculation BUT they go with the [defaults.toml ](https://github.com/haddocking/haddock3/blob/cb2b2d3324cb329c42a963d42c2ad4896861c25b/src/haddock/modules/refinement/mdref/defaults.toml#L1729)

rvhonorato: Until we refactor the CNS scripts I'd say its best to group the weights together with the other parameters.

amjjbonvin:
But in might make sense to use rather similar names - makes it easy to connect with the CNS code.


> This values are not used by the CNS code to do the calculation BUT they go with the defaults.toml  <https://github.com/haddocking/haddock3/blob/cb2b2d3324cb329c42a963d42c2ad4896861c25b/src/haddock/modules/refinement/mdref/defaults.toml#L1729>

rvhonorato:
@joaomcteixeira I will refactor that so that each module takes the correct weight from its default, is that ok?

rvhonorato (approved):

Issue #80, opened 2021-08-13
Reduce the number of files generated


rvhonorato:
Something for us to keep in mind is the amount of files generated by each module, we should aim to minimize and clean them as much as possible, keeping relevant information for debugging purposes.

PR #97, opened 2021-09-20
Structures the log system

joaomcteixeira:
Enhances the previous logging structure.

Added features:

* two files are now saved: `haddock3.log` and `haddock3.debug`, where the latter is only created if `--log-level debug` is given.
* log only outputs for `sys.stdout` if there is one. For example, some times clusters have no terminal output, for these cases the log prints nothing.
* log files are written **only** at the end of the execution and **not** on every log call. Log is kept in memory and only flushed to files at the end.
* log files are created even if the execution breaks
* in case the an error happens before reading the configuration file, the log files are saved in the `CWD`. If the error happens after reading the config, the log files are saved in the `run dir` folder.

### DIFF
--- a/examples/scoring/scoring.toml
+++ b/examples/scoring/scoring.toml
@@ -18,7 +18,7 @@ molecules = "../data/T161-rescoring-5.pdb"
 autohis = true
 
 ########################
-[emscoring]
+#[emscoring]
 
 ########################
 #[clustering]
--- a/src/haddock/__init__.py
+++ b/src/haddock/__init__.py
@@ -1,5 +1,23 @@
 """HADDOCK3 library."""
+import logging
+from os import get_terminal_size
 from pathlib import Path
+
+from haddock.libs.liblog import add_streamhandler
+
+
+log = logging.getLogger(__name__)
+log.handlers.clear()
+log.setLevel(logging.DEBUG)
+
+try:
+    # runs in clusters likely won't have terminal output
+    get_terminal_size()
+except OSError:
+    has_terminal = False
+else:
+    add_streamhandler(log)
+    has_terminal = True
 
 
 haddock3_source_path = Path(__file__).resolve().parent
--- a/src/haddock/__init__.py
+++ b/src/haddock/__init__.py
@@ -3,7 +3,7 @@ import logging
 from os import get_terminal_size
 from pathlib import Path
 
-from haddock.libs.liblog import add_streamhandler
+from haddock.libs.liblog import add_sysout_handler
 
 
 log = logging.getLogger(__name__)
@@ -16,7 +16,7 @@ try:
 except OSError:
     has_terminal = False
 else:
-    add_streamhandler(log)
+    add_sysout_handler(log)
     has_terminal = True
 
 
--- a/src/haddock/clis/cli.py
+++ b/src/haddock/clis/cli.py
@@ -1,13 +1,13 @@
 #!/usr/bin/env python3
 import argparse
-import logging
 import sys
 from argparse import ArgumentTypeError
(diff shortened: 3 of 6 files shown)


amjjbonvin:
> 
> log files are written only at the end of the execution and not on every log call. Log is kept in memory and only flushed to files at the end.

Does the log file also reports on the progress of the run (used to be the case if I remember correctly). In that case only writing it at the end is not very helpful.

Also if for some reason a run hangs, you won’t know at which stage it is…

joaomcteixeira:
hmm that's a good point. 

I can create another handler that logs only the title of the current stage that is running and not all the info and debug messages. So, if the run halts we can know at which stage it is.

But, I believe saving to the log file at every log call (even debug) might be very demanding for the disks maybe. So, I will keep those two handlers in memory and saved only when the run ends. I need to handle the `kill` statement I believe :wink:

amjjbonvin:
Rodrigo should probably comment, I remember we had in the previous version a live percentage of completion info while running.

rvhonorato:
As @amjjbonvin mentioned, in haddock2.x the `.log` doubles as monitoring and is very useful to catch lost jobs in the queue. We did not encounter this issue here yet because the batch submission system is not yet implemented, so it comes to: how will we monitor the jobs in regards to batch submission? 

I think here we can improve on how its done for haddock2.x (with the [`Queue*py`](https://github.com/haddocking/haddock/tree/haddock2.4/Haddock/Main) python modules) and handle the monitoring when we come to it, but as far as logging goes


   

>  two files are now saved: haddock3.log and haddock3.debug, where the latter is only created if --log-level debug is given.

remember we are trying to reduce the number of files (every one counts!) https://github.com/haddocking/haddock3/issues/80, so why not have it all in the .log?

> log only outputs for sys.stdout if there is one. For example, some times clusters have no terminal output, for these cases the log prints nothing.

in this case would it not be better to always save instead of not save?

> log files are written only at the end of the execution and not on every log call. Log is kept in memory and only flushed to files at the end.

again, for logging i think this is completely fine, we will need to think about it when we come to the monitoring.

rvhonorato:
> 
> Rodrigo should probably comment, I remember we had in the previous version a live percentage of completion info while running.

Indeed! I used https://github.com/tqdm/tqdm, but if I remember correctly we decided not to use it because it adds the carrier return `\r` which will complicate parsing the .log file.

amjjbonvin:
We could think of a separate status command that can be run on an existing run and returns the current status.

Something like that must be part already the restart option I would think

amjjbonvin:
Should check what haddock2.5 outputs as log - quite nice I thought

joaomcteixeira:
Many points given, I think we can separate them in different tasks, some to be addressed here, other I think should go to other issues/PRs.

1. The question of the two files. I thought about it to facilitate the reading when wanting only debug or only log. But @rvhonorato is right and we should reduce the number of files to the most. So, I will make it all log in a single file with a nice `LOG` and `DEBUG` prefix in the lines so the lines can be easily parsed with `grep`.
2. i thought about saving the log file only at the end to avoid be writing to the disk on every since log call. @amjjbonvin it is possible to know the current state of the run by going to the run dir and seeing what is the last folder created and files inside. Nonetheless, we can perfectly have a `state.log` that reports only the current stage that is running and when it started. Or I can make somehow an interface to ask for the progress of the run. This should go to another PR.
3. we can have a progress bar to monitor the progression of the workflow and of each stage. that should go to another PR.

I will be mindful about all these consideration and come back with additions modifications.

joaomcteixeira:
Very nice comments, closing and opening in a new one.

PR #104, opened 2021-10-29
Removes the need for `activate_haddock` global variables setup

### LINKED ISSUE #94: Remove the need for `activate_haddock`
Some of the HADDOCK3 parameters are currently defined as `bash` global variables. These forces the user to `source` the `activate_haddock` file, as well as, to handle `bash` scripting, for example to update the desired number of cores. We need to remove the need for the [`activate_haddock` file](https://github.com/haddocking/haddock3/blob/main/bin/activate_haddock) and have those parameter defined in the HADDOCK3 configuration file or in other parts of the software.

`activate_haddock` configures the parameters bellow. Ticks inform on those re-implemented.

- [x] `export HADDOCK3_NUM_CORES=4`
- [x] `export HADDOCKPATH=pwd `
- [x] `export PYTHONPATH=$PYTHONPATH:$HADDOCKPATH`
- [x] `export PATH=$PATH:$HADDOCKPATH/bin`
- [x] `export HADDOCK3_CNS_EXE="${HADDOCKPATH}/bin/cns"`

joaomcteixeira:
Closes #94 by addressing pending tasks: `CNS_EXE` and `ncores`.

## 1. `CNS_EXE` path

There is now one single instruction on how to install the `cns` executable script. Likewise, there is a single code route to retrieve that path. HADDOCK3 now expects `cns` to be properly linked inside the `bin` folder. Removed the need to pass through the global `bash` variable `HADDOCK3_CNS_EXE`. **However**, the user can *alternatively* give the path for the `cns` executable in the run configuration file. If a specific path is not specified in the configuration file, HADDOCK3 tries to retrieve the path in `bin/cns` (usual behavior). I made `cns_exec` to be definable as a parameter to facilitate future release of HADDOCK3 in PyPI for example.

## 2. `ncores` parameter

The `ncores` parameter defines how many CPU threads will be used during the calculations. Before, this value was configured by the global `bash` variable:

```bash
export HADDOCK3_NUM_CORES=4
```

With this Pull Request, `ncores` can be specified in the **configuration file**. There are two different ways to specify the number of cores:

1. A global format. That is, specifying the `ncores` parameter in the general parameter section of the configuration file; that is, before defining the modules parameters
2. Module-specific parameter. A part from the general settings, `ncores` can be defined specifically for each module. Module's defined `ncores` overwrite the global value.

I implemented `2.` because it was easy to do and gives us extra versatility. By default all threads (minus 1) are used, according to https://github.com/haddocking/haddock3/blob/3387ac04fb2c0d0069f7b86bb9bf3ef142282d16/src/haddock/libs/libutil.py#L84 with value `n=None`.


amjjbonvin (approved):

examples/docking/docking.toml
```diff
@@ -11,6 +11,7 @@ order = ["topoaa",
 
 # directory in which the scoring will be done
 run_dir = "run1"
+ncores = 8
```
rvhonorato, line 14: If we have this parameter here does it mean it always needs to be passed in the configuration file? This won't make sense once we reach the actual execution methods of HPC/GRID

joaomcteixeira: No, it doesn't mean that. It means that you can add it to control the number of CPUs. But if you don't add it, HADDOCK3 will use all CPUs available minus 1.

Note, it is also possible to have the `ncores` specified for each module separately. If that is not specified, the program will use the global value.

amjjbonvin:
What will happen if there is only one CPU … N-1=0. Not a very likely situation these days anyway

joaomcteixeira:
You are so right @amjjbonvin ! see #107 - this one was :godmode: (hope you know this meme :wink:)

PR #106, opened 2021-11-04
Updates the logging system

joaomcteixeira:
Redoing in a different way after comments in #97 

Updates and cleans the logging system.

* added complete control to where the logging messages are saved
* standard output is saved to `stdout`. Before it was saved to `stderr` because of the default value of [StreamHandler](https://docs.python.org/3/library/logging.handlers.html#logging.StreamHandler). This is noted mostly when running SBATCH
* `error` messages as saved also to `stderr`
* logging is saved in file `log` inside the run folder.

I needed to do a quick trick to keep the logging in stdout and in the `log` file to accommodate the first steps during the configuration reading. Basically, messages are reported to `stdout` and to a `StringIO` until the `log` file is defined. After that, the buffer in StringIO is saved to the log file, and the logging continues normally after that. See `clis/cli.py`.

This PR **does not** clean the logging messages *per se*, nor does it adds progression bars as discussed in #97. Instead, this PR just updates the logging system itself. These other functionalities will be built on top of this and will be submitted in other PRs.


amjjbonvin (approved):
Seems fine to me - the logic makes sense.
But git report conflicts that need to be resolved manually

joaomcteixeira:
Yes, now gives the expected output (not file size), running with `sbatch` on `tintin`:

```
 1 -rw-r--r--. 1 joaomctx users     0 Nov  5 14:03 hd3_scoring_err
 9 -rw-r--r--. 1 joaomctx users  4091 Nov  5 14:03 hd3_scoring_out
```

PR #116, opened 2021-11-12
Benchmarking system and configuration.

joaomcteixeira:
Okay,

PR development for benchmarking scripts and setup.

I made this branch in the home repository and not in a fork because we are going to work on this :relaxed:

**How to test and generate benchmark jobs:**

1. `git pull`
2. `git checkout benchmark`
3. Install the new CLI: `python setup.py develop --no-deps`
4. `haddock3-bm -h`, and follow the instructions.

`haddock3-bm` creates a folder with subfolder for each of the source `PDB_ID` folder cases. Inside each, there is a folder for each *benchmark scenario*.


joaomcteixeira:
strange this codecov report... it should decrease the coverage %...

joaomcteixeira:
**Todo:**

- organize results models folder by the size of input model folders.

joaomcteixeira (commented):
Should we add a prefix to the output folder names so they are sorted by size also when sorting by name? Otherwise, we need to be playing with `ls -ltr` to resort them again.

rvhonorato:
Changed a few things about the folder structure to make it a bit more "HPC-compatible" by setting everything relative to a root folder, ex: `haddock3-benchmark-test/200B`, receptor/ligand molecules are copied to `input` together with the `scenario.cfg`, job files go into `jobs` and the errors are logged in `logs`.

Since haddock is taking care of the stdout, changed the slurm script to `--ouput=/dev/null` so we don't have duplicated logs.

Also changed the scenarios (workflows) to reflect better the 2.4 pipeline, note that we are still all the analysis modules. Once we wrap this up we should finish those before doing the benchmark.

rvhonorato:
Also rolled back the sort by size, if this script is not the one that will be running the jobs makes no sense for it to do any sorting.

src/haddock/clis/cli_bm.py
```diff
@@ -217,29 +216,23 @@ def job_setup(header, path, conf_f):
...
 
-    path : Path or str
+    run_path : Path or str
         Path where results will be stored. Must be synchronized with the
         `run_dir` path in the configuration file.
 
     conf_f : Path or str
         Path to the configuration file.
     """
+    conda_sh = Path(Path(sys.executable).parent,
```
joaomcteixeira, line 226: I would suggest we clean this line. what was the problem with `$HOME/miniconda3`?

rvhonorato: my miniconda3 is not in the $HOME :P

joaomcteixeira:
nice the `dev/null` somehow addresses #117 for this. I will address  #117 properly in another PR but for now is perfect and also necessary to avoid creating another file.

src/haddock/clis/cli_bm.py
joaomcteixeira: Solved

joaomcteixeira:
and suddenly we have no `Queue` class in the `daemon` :smiling_imp: Why we need the class we if have no state? :feelsgood:

joaomcteixeira:
@amjjbonvin @rvhonorato this benchmark setup is ready for review. Read through the `docs/benchmark-tut.md` file and use the tools to see if you like the behaviour and feeling.

rvhonorato (approved):
Looks ok to me

rvhonorato:
What is missing here?

joaomcteixeira:
I think looking to `alpha` one benchmark to make sure I didn't miss anything. I think the documentation is also complete.

joaomcteixeira:
I haven't implemented here neither the [`analyze_bm`](https://github.com/haddocking/haddock3/blob/alpha1/benchmark/analyze_bm.py) nor the [`plots`](https://github.com/haddocking/haddock3/blob/alpha1/benchmark/plots_bm.py). Should I do it, or its this analysis better suited for the integration tests (#178) ?

amjjbonvin:
It’s not integration tests. But useful info also when you are changing the code.
So not only testing.

joaomcteixeira:
Okay, so that will go in another PR. This PR will be just for setting up the benchmark runs and the deamon. Whatever we do with that later is another issue. @rvhonorato @amjjbonvin you can go and merge this PR then. Remember to do `python setup.py develop --no-deps` to get the new CLIs.

amjjbonvin (approved):

PR #122, opened 2021-11-15
Implement `caprieval`

rvhonorato:
This PR implements the `caprieval` module. Its purpose is to calculate the CAPRI metrics, `I-RMSD`, `L-RMSD`, `I-L-RMSD` and `fnat`

The cutoffs, atoms and receptor/ligand chains are passed as parameters.

If no reference structure is provided, it takes the best ranking model as reference instead.

It uses `fccpy` instead of the "old" `contacts.ccp` script. 

The geometric operations and rmsd calculations were implemented from scratch in pure python using Kabsch algorithm.

It outputs a `capri.tsv` file that looks like this:
```
model	fnat	irmsd	lrmsd	ilrmsd	
rigidbody_0.pdb	0.175	6.10	14.53	5.70	
rigidbody_1.pdb	0.250	8.84	24.13	4.72	
rigidbody_2.pdb	0.225	3.70	11.71	3.82	
rigidbody_3.pdb	0.225	3.70	13.70	3.82	
rigidbody_4.pdb	0.225	3.70	11.71	3.82	
```

Note: I've quadrupled checked the values, and this module has shown that there's a bug with `rigidbody` in which it is producing identical models. This will be treated in a separate issue/PR #123, this PR here is only about the `caprieval` module.


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "src/fcc"]
 	path = src/fcc
 	url = https://github.com/haddocking/fcc.git
+	ignore = dirty
```
joaomcteixeira, line 4: once `fccpy` gets developed this should become a dependency and not a submodule.

requirements.yml
```diff
@@ -3,10 +3,11 @@ channels:
     - conda-forge
     - defaults
 dependencies:
-    - python==3.8
+    - python==3.9
```
joaomcteixeira, line 6: i am working with @JoaoRodrigues to understand what in FCCpy is incompatible with python 3.8. If there is no special reason we should keep it `3.8` compatible.

requirements.yml
```diff
@@ -3,10 +3,11 @@ channels:
...
     - pip
     - numpy
     - jsonpickle
     - tox
     - pip:
         - pdb-tools
+        - git+https://github.com/joaorodrigues/fcc@fcc2
```
joaomcteixeira, line 13: This is okay, so we don't need to submodule anymore.

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+            *ignore,
+            init_params=DEFAULT_CONFIG,
+            **everything):
+        super().__init__(order, path, init_params)
+
+    @classmethod
+    def confirm_installation(cls):
+        """Confirm if contact executable is compiled."""
+        dcfg = read_config(DEFAULT_CONFIG)
+        exec_path = Path(bin_path, dcfg['contact_exec'])
```
joaomcteixeira, line 395: We should make that `contact_exec` some module.

joaomcteixeira (commented):
just comments

rvhonorato:
This is finalized, still failing because of py38 issues, @joaomcteixeira up to you to either bump or refactor the dependency

joaomcteixeira:
You can merge, bump it to python 3.9. Don't worry about the tests I will correct it in another PR. Don't squash merge next time :wink:

joaomcteixeira:
But, have you worked out the new fccpy dependency? that fccpy works only in a branch of @JoaoRodrigues . Not so safe for now, though

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -0,0 +1,452 @@
...
+            self,
+            order,
+            path,
+            *ignore,
+            init_params=DEFAULT_CONFIG,
+            **everything):
+        super().__init__(order, path, init_params)
+
+    @classmethod
+    def confirm_installation(cls):
```
joaomcteixeira, line 388: This must be filled

joaomcteixeira (changes requested):

src/haddock/modules/analysis/caprieval/__init__.py
joaomcteixeira: Sorry, no need because it is a dependency.

tox.ini
```diff
@@ -5,54 +5,24 @@ ignore_basepython_conflict = true
...
 
 # configures which environments run with each python version
 [testenv]
 basepython =
-    {py38}: {env:TOXPYTHON:python3.8}
     {py39}: {env:TOXPYTHON:python3.9}
     {build,lint,radon,safety}: {env:TOXPYTHON:python3}
 passenv = *
 
-[testenv:py38]
```
joaomcteixeira: This block needs to exist. change to `[testenv:py39]`

tox.ini
```diff
@@ -5,54 +5,24 @@ ignore_basepython_conflict = true
...
-# execute pytest
-commands =
-    pytest --cov --cov-report=term-missing --cov-append --cov-config=.coveragerc -vv --hypothesis-show-statistics {posargs}
-# after executing the pytest assembles the coverage reports
-commands_post = 
-    coverage report
-    coverage html
-    coverage xml
-
 [testenv:py39]
```
joaomcteixeira: this block you can remove.

PR #129, opened 2021-11-22
Models start 1-index

joaomcteixeira:
Before models were starting 0-index line in python. Now they start 1-index.

@amjjbonvin do you think it is useful to have also models named `01`, `02`, ... if there are up to 99 models and `001`, `002`, etc if there are up 999 models, etc, to facilitate `ls` command? Or is it going to break any CNS part?


rvhonorato (approved):

amjjbonvin (approved):
A simple fix :-)

PR #134, opened 2021-11-23
Update docking example to Classic HADDOCK docking

rvhonorato:
Updates the docking example to reflect the classic HADDOCK docking protocol. 

🚧 🚧 🚧 🚧


amjjbonvin:
I suggest to have a `data` directory in each example and not as a common one. As we add more examples (like the ones for 2.4) the current structure would become messy with too many files in one location and possible conflicts.

amjjbonvin:
And to really test things we should run the same examples as 2.4 (and not only BM5).

rvhonorato:
> I suggest to have a `data` directory in each example and not as a common one. As we add more examples (like the ones for 2.4) the current structure would become messy with too many files in one location and possible conflicts.

Yes this is how it is in this pull request

rvhonorato:
Surpassed by #145

Issue #132, opened 2021-11-23
Wrong syntax for 1-index


rvhonorato:
Introduced in #129;

```python
>>> [(i, j) for i, j in enumerate(['A','B','C'])]
[(0, 'A'), (1, 'B'), (2, 'C')]
>>> [(i, j) for i, j in enumerate(1, ['A','B','C'])]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'list' object cannot be interpreted as an integer
>>> [(i, j) for i, j in enumerate(['A','B','C'], start=1)]
[(1, 'A'), (2, 'B'), (3, 'C')]
```

joaomcteixeira:
:facepalm:

PR #136, opened 2021-11-23
Adds support for global variables for CNS

joaomcteixeira:
Adds global variables for CNS runs according to what we discussed.

@amjjbonvin can you read over the documentation of `CNSJob` and see if they agree with expectations? Also, can you test out the examples? With this changes the `rigid-body` example broke. Likely because the `RUN` variable does not point to the same place as before. Should we update the var or the CNS scripts?

**EDIT:** likely the example broke because of #132

### DIFF
--- a/src/haddock/libs/libsubprocess.py
+++ b/src/haddock/libs/libsubprocess.py
@@ -3,6 +3,7 @@ import os
 import shlex
 import subprocess
 
+from haddock import toppar_path
 from haddock.core.defaults import cns_exec
 from haddock.core.exceptions import CNSRunningError, JobRunningError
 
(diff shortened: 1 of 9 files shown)


rvhonorato (approved):

src/haddock/libs/libsubprocess.py
```diff
@@ -45,20 +46,45 @@ def __init__(
...
             cns_exec=None,
             ):
         """
         CNS subprocess.
 
-        :param input_file: input CNS script
-        :param output_file: CNS output
-        :cns_folder: absolute execution path
-        :cns_exec: CNS binary including absolute path
+        Parameters
```
amjjbonvin, line 57: Shouldn't this also document the toppa_path?

joaomcteixeira: Okay, will commit on that.

src/haddock/libs/libsubprocess.py
```diff
@@ -45,20 +46,45 @@ def __init__(
...
+
+        cns_folder : str of pathlib.Path
+            The path where the CNS scripts needed for the module reside.
+            For example, `modules/rigidibody/cns`.
+
+        mod_path : str of pathlib.Path
+            Path where the results of the haddock3 module executing this
+            CNS job will be saved.
+
+        config_path : str of pathlib.Path
```
amjjbonvin, line 74: Here you mean the user configuration file, right?

joaomcteixeira: Yes, the path pointing to the user configuration file. But we need to folder, not the file itself. Let me check that.

joaomcteixeira:
Have a look to everything now. But the `rigid-body` example is not working on my side.

joaomcteixeira:
@amjjbonvin question: the `toppar` is the one in `cns/toppar` or the one in `modules/sampling/rigidbody/cns/toppar` ?

amjjbonvin:
That’s normal - need to correct the scripts

amjjbonvin:
> 
> the one in cns/toppar 

I will delete the other dirs once I update the scripts

joaomcteixeira:
Okay, now it is the `cns/topppar` as you can see `toppar` is imported from : https://github.com/haddocking/haddock3/blob/2eeab502c56e95fb543606b7bc6289ce3abf1ea9/src/haddock/__init__.py#L16

Feel free to merge this PR if you feel it does the job. So you can continue the CNS progress.

amjjbonvin (approved):

Issue #138, opened 2021-11-23
Weird behaviour when PDB input files have no chainID


amjjbonvin:
In the `topoaa` module, when an input PDB file has no chainID/segID defined the module puts `$PRO` as segid which causes problems downstream. No segIDs were defined in the config files since those are in principle defined in the `default.cfg` file of the module.

This does not happen when a PDB input has the chainID present. 

In `default.cfg` there are default segIDs defined for each molecule. Those should be used.

joaomcteixeira:
Is it correct to say that each input PDB file must have a single chain?

joaomcteixeira:
I am working on this right now.
If we have two input structures. One with chainID `A` and the other without chainID/segID (bugged). Which letter should we assign to the second structure? `A` or `B`? In other words, should the chainID/segID added to the structures with this information missing follow the IDs of the ones which have, or can they be independent?

rvhonorato:
Since I was assigned first, I already fixed this :)

rvhonorato:
> I am working on this right now. If we have two input structures. One with chainID `A` and the other without chainID/segID (bugged). Which letter should we assign to the second structure? `A` or `B`? In other words, should the chainID/segID added to the structures with this information missing follow the IDs of the ones which have, or can they be independent?

Basically needs to be the same as for 2.x

rvhonorato:
I won't have any more time this week to clean and PR this one, so probably better if you go ahead @joaomcteixeira

amjjbonvin:
In 2.4 the chains are assigned in the order of the input molecules (if not specified).

For 3.0 I would do the same, first molecule gets A, second B and so on, unless overwritten by parameters in the config file

joaomcteixeira:
Will address this as part of a sanitization process for all PDBs entering the haddock3 pipeline.

rvhonorato:
Would be nice to have this PDBHandler part as an independent library so we could also use it in the webserver;[ check what it does now]( https://github.com/haddocking/haddock-webserver-flask/blob/master/tools/PDBHandler.py).

rvhonorato:
Is this one still relevant?

joaomcteixeira:
Yes, it's relevant. It's solved in #144. But I need to finish that PR.

rvhonorato:
Apparently it was fixed in #144

PR #140, opened 2021-11-23
Cleaned emref module and adapted other to new env variables

amjjbonvin:
Docking example now works including the emref module.
Adapted all previous modules to use the new environment variables (related to issue 
Next step is mdref

PS: The environment variables are printed in the output for each model - must be an old debug statement remaining somewhere in the python code


rvhonorato (commented):
194 files, impossible to check everything, will trust @amjjbonvin also on this one!

src/haddock/modules/refinement/emref/__init__.py
```diff
@@ -108,7 +110,7 @@ def run(self, **params):
                 self.path,
                 self.recipe_str,
                 self.params,
-                ambig_f=self.params.get('ambig', None),
+                ambig=self.params.get('ambig', None),
```
joaomcteixeira, line 113: This was a good bug!

joaomcteixeira:
> PS: The environment variables are printed in the output for each model - must be an old debug statement remaining somewhere in the python code

taking note of this

joaomcteixeira:
I agree with Rodrigo. Got ahead and merge this PR on your own :rocket: :guitar: :fire:

src/haddock/modules/refinement/emref/__init__.py
joaomcteixeira, line 113: There is a bug here with `ambig` not being `ambig_fname`. I will correct that and inspect also other things. Please wait before merging.

joaomcteixeira (changes requested):
Give me some time to clear some small bugs relating to var names.

joaomcteixeira:
The `ambig_fname` was not being loaded due to an error in the variable name being pulled from the param dictionary `ambig` instead of `ambig_fname`. I am correcting that and working on a way to block future var typing errors at this stage. Var typing errors are already corrected when writing the config.

joaomcteixeira:
@amjjbonvin I corrected the capturing of `ambig_fname` parameter. Before there was an error and the ambig file was **NOT** being loaded into the CNS `.inp` file because `self.params.get('ambig', None)` was being used and masked the bug. Now that `ambig_fname = ""` is part of the `defaults.cfg` this can be rewritten to a normal `self.params['ambig_fname']`.


You may wish to review the calculations again after [this change](https://github.com/haddocking/haddock3/pull/140/commits/0b7678445cf520311a261fea95faa035e1abeb80).

amjjbonvin:
Ok- thanks - is this part of the PR I should merge?

Did you check the other modules, they all seem to use the same construction, i.e. ambig and not ambig_fname…

And it was working.

joaomcteixeira:
Yes, I checked the other modules. All corrected now. The new changes are already part of the of this branch. When the PR it merged the new changes will go along. I was working on #138 from this branch. If you want we can solve that before merging this one. Would be best I think.

amjjbonvin:
Ok - then try to solve it - I won’t update further the modules until this is solved.

joaomcteixeira:
Perfect. Converted to `draft` until #142 is solved.

joaomcteixeira:
Because we closed #142, I think this is safe to merge. Whatever comes after #142 is big enough to hold by itself.

joaomcteixeira (approved):

PR #142, opened 2021-11-24
Solves chainID and segID missing in input molecule

### LINKED ISSUE #138: Weird behaviour when PDB input files have no chainID

joaomcteixeira:
I am pull requesting this to the `haddock3-emref` branch from #140 .

Closes #138 

* makes nice use of `pdbtools` lib mod from https://github.com/haddocking/pdb-tools/pull/107
* Chain/SegIDs are added by order to the molecules. For example, if there are 3 molecules (`A`, `missing`, `C`). The `missing` will get `B`.

But, if there are 3 molecules (`A`, `missing`, `B`). The missing will get `C`. Is this expected? Should we correct also this and have all molecules `ABC...` in order, or is it something the user must correct?

### DIFF
--- a/src/haddock/libs/libpdb.py
+++ b/src/haddock/libs/libpdb.py
@@ -12,6 +12,24 @@ from haddock.libs.libutil import get_result_or_same_in_list
 from haddock.modules import working_directory
 
 
+slc_record = slice(0, 6)
+slc_serial = slice(6, 11)
+slc_name = slice(12, 16)
+slc_altloc = slice(16, 17)
+slc_resname = slice(17, 20)
+slc_chainid = slice(21, 22)
+slc_resseq = slice(22, 26)
+slc_icode = slice(26, 27)
+slc_x = slice(30, 38)
+slc_y = slice(38, 46)
+slc_z = slice(46, 54)
+slc_occ = slice(54, 60)
+slc_temp = slice(60, 66)
+slc_segid = slice(72, 76)
+slc_element = slice(76, 78)
+slc_model = slice(78, 80)
+
+
 def get_supported_residues(haddock_topology):
     """Read the topology file and identify which data is supported."""
     supported = []
--- a/src/haddock/core/exceptions.py
+++ b/src/haddock/core/exceptions.py
@@ -19,6 +19,12 @@ class ModuleError(HaddockError):
     pass
 
 
+class MoleculeError(HaddockError):
+    """Error in a HADDOCK3 module."""
+
+    pass
+
+
 class StepError(HaddockError):
     """Error in a HADDOCK3 workflow step."""
 
--- a/src/haddock/libs/libstructure.py
+++ b/src/haddock/libs/libstructure.py
@@ -3,7 +3,7 @@ import os
 import string
 from pathlib import Path
 
-from pdbtools import pdb_chainxseg, pdb_segxchain, pdb_chain
+from pdbtools import pdb_chainxseg, pdb_segxchain, pdb_chain, pdb_tidy
 
 
 slc_record = slice(0, 6)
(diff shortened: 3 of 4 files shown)


amjjbonvin:
> But, if there are 3 molecules (A, missing, B). The missing will get C. Is this expected? Should we correct also this and have all molecules ABC... in order, or is it something the user must correct?
> 

If the user does not specify in the config file specific prot_segid entries,  then the order of the molecules should be used and the existing chainID/segID overwritten.

May-be good to add a message in the log file about the chainID assigned to each molecule

joaomcteixeira:
How should the user define the `prot_segid` in the configuration file?

```toml
run_dir = "run1"
ncores = 20
# cns_exec = "path/to/bin/cns" # optional

# ensemble to be scored
molecules =  [
    "./data/1a2k_l_u.pdb",
    "./data/1a2k_r_u.pdb"
    ]

[topoaa]
autohis = true
[topoaa.mol1]
prot_segid = 'A'
```

Like this? And in this case the other molecules should follow the alphabetical order. Is that it?

amjjbonvin:
Sounds like a good solution indeed.

But in this particular examples - scoring - this should not apply.
Since molecules will contain multiple chains already.

I did not check yet the scoring module, but it should not depend on the topoaa one (if implemented as we have in 2.4).

joaomcteixeira:
So far, the changes proposed in this PR only affect the `topoaa` module. Basically the `chaindIDs` and `segIDs` are modified (if needed) before writing the files to the `00_topoaa` folder. After that, haddock3 goes as per usual.

src/haddock/libs/libpdb.py
```diff
@@ -12,6 +12,24 @@
 from haddock.modules import working_directory
```
amjjbonvin, line 14: What will happen if a PDB file is truncated (e.g. the lines do not have the correct official length, missing for example the segid columns?

amjjbonvin: This is sometimes happening depending on the software used... or are you using some PDB-tools tool to first "clean" the file?

joaomcteixeira: Currently in the `main` branch, there is "sanitize" step that takes place in `topoaa` after the input molecule is split into models:

https://github.com/haddocking/haddock3/blob/2dea812673cee9fb759a5bba52f9576d4504208f/src/haddock/modules/topology/topoaa/__init__.py#L94

This sanitize step removes unneeded lines containing the [following strings](https://github.com/haddocking/haddock3/blob/2dea812673cee9fb759a5bba52f9576d4504208f/src/haddock/libs/libpdb.py#L26).

And [removes lines](https://github.com/haddocking/haddock3/blob/2dea812673cee9fb759a5bba52f9576d4504208f/src/haddock/libs/libpdb.py#L20) from residues not supported by the [topology format](https://github.com/haddocking/haddock3/blob/main/src/haddock/cns/toppar/haddock.top). Residues are removed either if `RESIdue` or `Residue` in the topology format. Have a look if that is desired.

This part of the process was inherited from `beta1` branch, so now is the time to revisit it and correct it.

Question: Should all PDBs be sanitized before entering the HADDOCK3 workflow?

amjjbonvin:
> This sanitize step removes unneeded lines containing the following strings <https://github.com/haddocking/haddock3/blob/2dea812673cee9fb759a5bba52f9576d4504208f/src/haddock/libs/libpdb.py#L26>.
> 
> 

We can probably extend a bit this. E.g. the ANISOU lines can be discarded.

> And removes lines <https://github.com/haddocking/haddock3/blob/2dea812673cee9fb759a5bba52f9576d4504208f/src/haddock/libs/libpdb.py#L20> from residues not supported by the topology format <https://github.com/haddocking/haddock3/blob/main/src/haddock/cns/toppar/haddock.top>. Residues are removed either if RESIdue or Residue in the topology format. Have a look if that is desired.
> 
> 
Actually residues are removed if not present as RESI in the haddock.top <http://haddock.top/> file
But this is somewhere a limitation, for example for small molecule docking for which the ligand is never known.

In 2.4 we can provide ligand.top <http://ligand.top/> and ligand.param files so that it is taking into account.
CNS will simply discard any unknown residues.

I was already thinking of adding in the default.cfg the following parameters:

	ligand_top_fname
	ligand_param_fname

And modify the scripts in the various modules to read those if defined.
For this to work the pre-processing should thus not discard those.
Several options here:

1) We output a warning with a list of unknown residues
2) The python code checks in a ligand.top <http://ligand.top/> file is given in the config file that matches the unknown ligand name


Also another preprocessing that should be done by default is to rename the MSE residues (selenomethionine) to MET (methionine) and change the entries from HETATM to ATOM.

And while at it, we might also check for double occupancies, and insertions, and raise a warning about those.

joaomcteixeira:
Okay. So, what was done in this PR is correct but after your comment having it in the `topoaa` module might not be the best place. We should have a PDB "sanitize"-like step before sending the files to the haddock3 workflow. In that way it would be easier to add/remove things. In other words, every PDB entering the `topoaa` module should be clean. Having this step separated will facilitate its incorporation into other libraries (if needed).

Closing this PR but I will use it as a reference for the next one.

PR #147, opened 2021-11-25
Bugfix `caprieval` 

### LINKED ISSUE #146: `caprieval` fails if interface arrays have different dimenions
If the reference and the model have a different number of atoms, the following is raised:

```text
Exception has occurred: ValueError
shapes (3,400) and (399,3) not aligned: 400 (dim 1) != 399 (dim 0)
  File "/Users/rodrigo/repos/haddock3/src/haddock/modules/analysis/caprieval/__init__.py", line 43, in kabsch
    C = np.dot(np.transpose(P), Q)
  File "/Users/rodrigo/repos/haddock3/src/haddock/modules/analysis/caprieval/__init__.py", line 212, in irmsd
    U = kabsch(P, Q)
  File "/Users/rodrigo/repos/haddock3/src/haddock/modules/analysis/caprieval/__init__.py", line 443, in run
    capri.irmsd(cutoff=self.params['irmsd_cutoff'])
  File "/Users/rodrigo/repos/haddock3/src/haddock/libs/libworkflow.py", line 104, in execute
    module.run(**self.config)
  File "/Users/rodrigo/repos/haddock3/src/haddock/libs/libworkflow.py", line 25, in run
    step.execute()
  File "/Users/rodrigo/repos/haddock3/src/haddock/clis/cli.py", line 143, in main
    workflow.run()
  File "/Users/rodrigo/repos/haddock3/src/haddock/clis/cli.py", line 62, in cli
    main(**vars(cmd))
  File "/Users/rodrigo/repos/haddock3/src/haddock/clis/cli.py", line 67, in maincli
    cli(ap, main)
  File "/Users/rodrigo/repos/haddock3/src/haddock/clis/cli.py", line 154, in <module>
    sys.exit(maincli())
```

We should properly handle this error and also add an option to `ignore_missing`; this option should ignore atoms in the `model` that are not present in the `reference`.

rvhonorato:
This PR fixes #146 by adding an option to `ignore_missing` in which only atoms present in the reference interface are used for the calculations.


src/haddock/modules/analysis/caprieval/defaults.cfg
```diff
@@ -3,9 +3,10 @@ irmsd = true
 fnat = true
 lrmsd = true
 ilrmsd = true
-irmsd_cutoff = 10.0
+irmsd_cutoff = 5.0
```
amjjbonvin, line 6: Hum - isn't CAPRI using all interface residue making contacts within 10A?  At least for protein-protein

src/haddock/modules/analysis/caprieval/defaults.cfg
```diff
@@ -3,9 +3,10 @@ irmsd = true
...
 fnat_cutoff = 5.0
 lrmsd_cutoff = 10.0
 receptor_chain = 'A'
```
amjjbonvin, line 9: Now we are working on taking the segIDs directly from the input PDBs. What if they don't use A and B?  Does it mean the user would have to define those for things to work?

rvhonorato: Something we need to refactor across all modules once we implement it.

amjjbonvin: But currently the code takes the segID from the input PDB and does not overwrite it, correct.
So we have to assume (for the time being) that the molecules have alphabetical segIDs

rvhonorato: Yep this is why for now the receptor and ligand chains are passed as a parameter, once we implement it we need to refactor it to get from the internal structure instead.

joaomcteixeira (approved):
Found no critical things.

PR #156, opened 2021-11-27
Add support for ligands and DNA and refactor `mdref` module

amjjbonvin:
- Added option to provide custom ligand param/top files
- Added protein-ligand example
- Added protein-DNA example and adapted CNS modules for it

Note that the `caprieval` module does not work for DNA/RNA (yet)


examples/docking-protein-DNA/docking-protein-DNA.cfg
```diff
@@ -0,0 +1,55 @@
...
+#reference = './data/target.pdb'
+
+[seletop]
+select = 5
+
+[flexref]
+ambig_fname = './data/cro_air.tbl'
+epsilon = 78
+noecv = false
+dnarest_on = true
```
rvhonorato, line 40: Can this be true and not take `dna_rest_fname`?

examples/docking-protein-ligand/data/run.cns.patch
```diff
@@ -0,0 +1,63 @@
+--- run.cns	2019-03-12 11:23:14.508739493 +0100
```
rvhonorato, line 1: Does the example need this file?

examples/docking-protein-ligand/docking-protein-ligand.cfg
```diff
@@ -0,0 +1,52 @@
...
+
+# ====================================================================
+# Parameters for each stage are defined below
+#####################################################################
+# WARNING: THE PARAMETERS HERE ARE ILLUSTRATIVE
+#  THE WORKFLOW IS WORK-IN-PROGRESS
+#####################################################################
+[topoaa]
+autohis = true
+ligand_param_fname = "./data/ligand-prodrg.param"
```
rvhonorato, line 23: the `./` is not needed, but makes no difference to have it

amjjbonvin:
No - it does not - but I had it to know which parameters to change for ligand settings.
Can be removed at a later stage once the ligand example will work

rvhonorato (approved):
Looks ok to me.

amjjbonvin:
I made `dna-rna_restraints` an automatic script to figure out the base pairing - no need to generate custom ones anymore.
Which is why `dna_rest_fname` is no longer needed (and we don’t need anymore X3DNA as the server uses to generate it)

rvhonorato:
nice!

rvhonorato:
> Note that the `caprieval` module does not work for DNA/RNA (yet)

Indeed it does not work because its considering only protein backbone atoms.

amjjbonvin:
A simple solution would be to add to the atom selection `P,C1,C9`

amjjbonvin:
@rvhonorato you might want to check the mdref module which I just pushed - working fine in my hands
We should then merge this one into the main.

rvhonorato (commented):
The `mdref` works fine for both protein and DNA examples but the ligand example is not running.

`libpdb.py:L80` only keeps residues that are present in the topology file

```python
# cns_paths.py:L9
topology_file = Path(toppar_path, "haddock.top")

# libpdy.py:L30
_to_keep = get_supported_residues(topology_file)

# libpdy.py:L91
                res = line[17:20].strip()
                if res and res in _to_keep:
                    good_lines.append(line)
```

This results in `00_topoaa/oseltamivir_zwitterion.pdb` being empty.

I see that the ligand toppar are passed correctly to the `.inp`

```fortran
eval ($ligand_param_fname="./data/ligand-prodrg.param")
eval ($ligand_top_fname="./data/ligand-prodrg.top")
```

But this ligand topology should be used by the sanitizer to decide which residues to keep. 

Its something that should be refactored in `libpdb.sanitize()` and in the `topoaa` module

for example:
```python
# topoaa/__init__.py

if self.params['ligand_top_fname']:
    libpdb.sanitize(model, overwrite=True, custom_topology=self.params['ligand_top_fname'])
else:
    libpdb.sanitize(model, overwrite=True)

# libpdb.py:L80

def sanitize(pdb_file_path, overwrite=True, custom_topology=False):
    if custom_topology:
        #> load the custom topology and add it to _to_keep list

```

I will do a patch in the python part and add to this PR if that's ok with you! 🛠️

rvhonorato:
Ligand example now also working ✅ 

Refactored the relevant python parts, added `ligand_top_fname` that was missing in the other CNS-based moduled,  updated `docking-protein-ligand.cfg` and brought it in line with `main`.

Ready to merge 🟢

examples/docking-protein-ligand/docking-protein-ligand.cfg
```diff
@@ -26,27 +26,30 @@ delenph = false
...
+# [caprieval]
+# reference = './data/target.pdb'
 
 [seletop]
 select = 5
 
 [flexref]
 ambig_fname = './data/ambig-passive.tbl'
+ligand_param_fname = "./data/ligand-prodrg.param"
+ligand_top_fname = "./data/ligand-prodrg.top"
```
joaomcteixeira, line 45: There's no need to add the `./` before the path.

examples/docking-protein-DNA/data/run.cns.patch
```diff
@@ -0,0 +1,74 @@
+--- run.cns	2019-03-11 12:07:52.402707184 +0100
```
joaomcteixeira, line 1: Why there is a CNS file in the examples folder?

rvhonorato: oh, forgot to remove this file

examples/docking-protein-ligand/docking-protein-ligand.cfg
rvhonorato: old habits die hard 😅

joaomcteixeira:
Apart from the `run.cns.patch` there are several other CNS-related files in the examples/ folders. If they are necessary, we should think about moving them to the main code as implemented functions; instead of having them in the examples. Otherwise, if they are left-overs committed by mistake, just remove them.

rvhonorato:
> Apart from the `run.cns.patch` there are several other CNS-related files in the examples/ folders.

`.tbl` are restraints `.top` are topologies and `.param` are parameters, so these are all information, not scripts

amjjbonvin:
Note that the ligand issue is not only related to not having its RESI in the haddock.top file - I tried that and still empty coordinate file after the sanitise step...

PR #161, opened 2021-11-29
Add homotrimer docking example with C3 symmetry

amjjbonvin:
Added homotrimer docking example with C3 symmetry

@rvhonorato - `caprieval` will have to be adapted to deal with such a case. 

Probably best to calculate the l-RMSD by fitting on chain A and calculating the RMSD on all other chains.

And there is the issue of the possible permutations of chains, e.g. ABC vs ACB - which in principle are the same complex but for RMSD calculations will give different values.


examples/docking-protein-homotrimer/docking-protein-homotrimer.cfg
```diff
@@ -0,0 +1,127 @@
...
+[topoaa]
+autohis = true
+
+[rigidbody]
+ambig_fname = 'data/1qu9_whiscy_air.tbl'
+sampling = 20
+noecv = true
+ncs_on = true
+numncs = 2
+ncs_sta1_1="2"
```
joaomcteixeira, line 31: you can write `ncs_stat1_1 = "2"` if you wish, spaces are allowed, not mandatory though. But if those are numbers, they should be written without `"`. Do they need to be written to CNS with `"` despite the fact they are numbers?

amjjbonvin:
> you can write ncs_stat1_1 = "2" if you wish, spaces are allowed, not mandatory though. But if those are numbers, they should be written without ". Do they need to be written to CNS with " despite the fact they are numbers?
> 
If we do it without double quotes, we need to change the CNS scripts - which is fine in principle. 
CNS has a function to encode and decode strings.

But what do you then define in the default.cfg file?

examples/docking-protein-homotrimer/docking-protein-homotrimer.cfg
```diff
@@ -0,0 +1,127 @@
...
+numc3sym = 1
+c3sym_sta1_1="2"
+c3sym_end1_1="128"
+c3sym_seg1_1="A"
+c3sym_sta2_1="2"
+c3sym_end2_1="128"
+c3sym_seg2_1="B"
+c3sym_sta3_1="2"
+c3sym_end3_1="128"
+c3sym_seg3_1="C"
```
joaomcteixeira, line 121: I see we have duplicated parameters here among modules. We can add a `take_from_previous = true` kind-of parameter to capture parameters from the previous modules. For example, the hierarchy could be:

`defaults` -> `previous module` -> `current module`.

**Only** if `take_from_previous = true`

joaomcteixeira:
variables defined in the configuration files are loaded to python variables by the `config_reader` module. human-like strings are converted to strings. numbers to numbers, lists to lists, etc. Once these objects exist in the python-shell, they need to be written to the CNS module. These are two independent steps. It is possible to inject the variables in the modules without the config files (for example if you are using Jupyter Notebooks to run HADDOCK)

amjjbonvin:
But how would that work with a seletop module in between which has very different parameters.
And not all modules have the same parameters, which might lead to validation errors.

joaomcteixeira:
That's a good point. But if you like the idea I will find a clean way to make it work :wink:

joaomcteixeira:
On another note, I am still worried about these numbers with "quotes". Are they supposed to be strings in CNS or is that all numbers in CNS require quotes? The way variables are written to CNS is defined here:

https://github.com/haddocking/haddock3/blob/09505863e8747bc379d33ea49059201cd1be87e9/src/haddock/libs/libcns.py#L34-L45

amjjbonvin:
Could simplify things possibly - but not a big deal to copy paste params - but also more sensitive to errors

amjjbonvin:
It all depends on how the CNS scripts are handling those.

We can decide to put numbers without quotes, but need to change/check the CNS scripts.

And again, how do you define an empty variable in the default.cfg file?

None?

amjjbonvin:
For all the variables I defined in the example as numbers between quotes. 
If we decide to remove the quotes what do you define then in the default.cfg file when that variable is undefined?

rvhonorato:
If you define a number with " it will result in `$eval (variable="2")`

amjjbonvin:
But now if a variable (even number) is not set in the default.cfg file of the module it is defined as 

ncs_sta1_1 = ""

So if we want to get rid of quotes for numbers, what should you then define in that file?

ncs_sta1_1 = None ?

Something must be defined.

rvhonorato:
To get rid of the quotes for number just remove them from the .cfg
ncs_sta1_1 = 2
`eval ($ncs_sta1_2=2)`

But indeed, trying to set them as empty will make CNS recognize them as a string `eval ($ncs_sta1_2="")`

Is there a CNS statement for NONE variables? If not then we need to refactor the .inp creator to not write empty `""` or `None`

rvhonorato:
Looked into the manual and it does not seem so.

So another issue, we need to refactor the config gears to not write empty parameters to the CNS input files.

joaomcteixeira:
The only thing you need to NOT write variables to the CNS file is to filter out the `None`s in this loop:

https://github.com/haddocking/haddock3/blob/09505863e8747bc379d33ea49059201cd1be87e9/src/haddock/libs/libcns.py#L32

I am not understanding if you want to AVOID writing empty variables to CNS or if you are looking for a CNS-syntax to define empty variables. I think is the former.

amjjbonvin:
Empty parameters are not an issue for cns in principle - but again what would you define for those in the default config file.

Currently the machinery is working fine as is

examples/docking-protein-homotrimer/docking-protein-homotrimer.cfg
amjjbonvin: But you might only want part of the parameters to be taken over - so adds complexity...

examples/docking-protein-homotrimer/docking-protein-homotrimer.cfg
amjjbonvin: Again - if we change it we need to change as well the CNS code since currently it is decoding those strings into numbers. Actions on the Python and CNS levels must be coordinated if we change things

examples/docking-protein-homotrimer/docking-protein-homotrimer.cfg
joaomcteixeira: If that is the case, so better to keep copy+pasting until the need actually arises. I thought the situation was more strict.

joaomcteixeira:
Run the examples locally and they are working also on my side. Merging. Enough of adding more stuff to this PR :stuck_out_tongue_closed_eyes:

PR #163, opened 2021-11-29
`None` variables are NOT written to CNS

### LINKED ISSUE #162: Do not write empty parameter to CNS input
We need to refactor the config gears to not write empty parameters to the CNS input files.

_Originally posted by @rvhonorato in https://github.com/haddocking/haddock3/issues/161#issuecomment-981648329_

and we need of course to define what is the standard way of defining empty parameters in the `defaults.cfg`

I vote for `variable = None`, which should be used for booleans, string, integers and floats.

joaomcteixeira:
Closes #162 

Parameters defined with `None` will NOT be written to the CNS header files. For example:

`bad_param = None`
`bad_param = none`

BUT, empty strings WILL be written to CNS files.

I prefer that empty variables are defined only by `None`.

Please test it, and if does not work let me know. Because it SHOULD work just with these changes.

**Important:** I don't know the behaviour of CNS if it expects a variable somewhere down the line that was not defined in the header.

See also #161


rvhonorato (approved):

joaomcteixeira:
@amjjbonvin @rvhonorato  I need your help and close attention in this last commit, I might have found a bug for a silenced functionality.

1. I implemented the *empty vars* as we discussed, those are the changes in the `libcns`. (nice FP-piping there :wink:)
2. Question 1: When adding paths to the CNS files, do you want them to be absolute or relative to the CWD or where the config file is?
3. Most importantly (the :bug:), when injecting the parameters in the CNS file [there is this code](https://github.com/haddocking/haddock3/blob/81e5ae0df003566389e73ee9cbf23081ab1872cd/src/haddock/libs/libcns.py#L56-L62), which does some special treatment of nested dictionaries to the `chain` keyword. However, where is that `chain` coming from? When I was refactoring the `libcns`, I came to the conclusion that `chain` **might be** what is defined in the `topoaa/defaults.cfg` as `input`, since it is the only parameter data that has that nested dictionary structure. **Is that it?** If yes, the previous commit corrects it, and now those parameters ARE saved in the CNS file, when before they were not. If not, please help me understand it, I might need to revert the last commit.

Try running some examples in this branch generating ONLY the topologies to see if you have an expected behaviour. There are some prints, use them or never mind them.

amjjbonvin:
Another related issue:

I am testing now the manual definition of histidines, but this now requires adding all possible parameters to the default.cfg file of the `topoaa` module. But here is the problem: some protein might have a lot, some a few. Now for things to work we have to define many entries such as: 
```
hisd_1 = None
hisd_2 = None
...
```

And if we need more variables because a protein contain more HIS, then the default file has to be edited. Could we for such nested variables allow for some flexibility, i.e. you only need to define one in the defaults.cfg file and any increased index would be accepted.

joaomcteixeira:
We can have a unique parameter for His and define it as a list, for example:

`his = [0, 1, 0, 2, 3, 0]`

each number is a histidine in the sequence and the number reflects a protonation state. That is how BioExcel Building blocks is doing.

Otherwise we can have a dictionary:

```
[topoaa]
[topoaa.hisd]
hisd_1=none
...
```
this would avoid the need to edit the `defaults`, since the latter just needs to have an empty `[hisd]` entry.

amjjbonvin:
> 2. Question 1: When adding paths to the CNS files, do you want them to be absolute or relative to the CWD or where the config file is?

It could be defined with respect the MODDIR variable (this of course assumes that all data are within the user directory where haddock3 is called from

>  However, where is that `chain` coming from? 

We don't have it in the `default.cfg` file. Instead we have per molecule the `prot_segid` entry

Chain might be then coming from what is defined in the PDB file itself.

amjjbonvin:
> We can have a unique parameter for His and define it as a list, for example:
> 
> `his = [0, 1, 0, 2, 3, 0]`
> 
> each number is a histidine in the sequence and the number reflects a protonation state. That is how BioExcel Building blocks is doing.

By default all are charges and I only want to specific - per molecule - the other states when needed (either hisd or hise).
The CNS module expect those if `autohis = false` and `numhisd` or `numhise` > 0

```
> [topoaa]
> [topoaa.hisd]
> hisd_1=none
```
This would not work since you need to define the HIS per molecule. And each molecule can have multiple entries.
I would think that some construction as in the `default.cfg` file of the `topoaa` module should be the way to go...  A

But this parameter is not showing anywhere in the `defaults.cfg` file.... How do I know how to define the histidine states?

amjjbonvin (approved):
I will trust you on this one and test later :-)

joaomcteixeira:
Done.
@amjjbonvin Now the `topoaa/defaults.cfg` `[input]` params are written to the `CNS` file. Before they were not.

Also, to add parameters to the `[input]` use:

```toml
[topoaa]
autohis=false
[topoaa.input.mol1]
whatever_new_parameter_here=false
# these parameters are not confirmed if they exist in the defaults config (for now)
```

src/haddock/libs/libcns.py
```diff
@@ -25,45 +27,94 @@ def generate_default_header(protonation=None):
...
+        return f'eval (${param}={value}){linesep}'
+
+    elif isinstance(value, str):
+        return f'eval (${param}="{value}"){linesep}'
+
+    elif isinstance(value, Path):
+        return f'eval (${param}="{str(value)}"){linesep}'
+
+    elif isinstance(value, (int, float)):
+        return f'eval (${param}={value}){linesep}'
```
rvhonorato, line 110: This will have all decimals of the float, should we do something about it to make it CNS-safe?

rvhonorato (approved):
👍🏽

PR #176, opened 2021-12-06
Refactor code to allow for refinement

### LINKED ISSUE #173: Refactor CNS refinement modules to be executed after `topoaa`
With https://github.com/haddocking/haddock3/pull/152 we changed the data structure that `topoaa` exports which made it incompatible with the refinement modules.

These need to be refactored to be able to be executed after `topoaa`.

Additionally, the CNS code is expecting a single `input_pdb_filename_1`, meaning that the multiple molecules need to be merged as a single structure with multiple chains.

rvhonorato:
This PR refactors multiple modules to allow for refinement also adds it as an example.

The preferred way to submit things to refinement should be as an ensemble:

```python
# Ensemble to be refined
molecules = [
     "data/e2a-hpr_1GGR_ens.pdb",
    ]
```

If molecules are given separately, you need to use `merge`
```python
molecules =  [
    "data/e2a-hpr_1GGR_A.pdb",
    "data/e2a-hpr_1GGR_B.pdb"
    ]
# ...
[merge]
```

By using `merge` its possible to refine one ensemble against another, for example:
```python
molecules = [
    "receptor_ensemble.pdb",
    "smallmolecule_ensemble.pdb"
    ]
# ...
[merge]
```

`merge` will create all combinations (cartesian product), but we could also add a parameter to only do it pairwise.

I did touch the scoring modules yet, will do that in another PR since they are probably outdated.

* * *
Closes #173


src/haddock/gear/read_io.py
```diff
@@ -0,0 +1,22 @@
+"""Gear to handle the reading of the IO."""
```
joaomcteixeira, line 1: If this file is so heavily dependent on libs.libontology and is something we forsee a core part of the workflow, it should go to `libontology` directly.

src/haddock/libs/libpdb.py
```diff
@@ -57,6 +58,18 @@ def split_by_chain(pdb_file_path):
     return get_new_models(pdb_file_path)
 
 
+def merge(pdb_file_list, new_pdb_file_path):
```
joaomcteixeira, line 61: Honestly, I would use `pdbtools` here:

```python
from pdbtools import pdb_merge, pdb_tidy

def merge(pdb_file_list, new_pdb_file_path):
    input_fhs = (open(p) for p in pdb_file_list)
    _ = pdb_merge.run(input_fhs)
    merged = pdb_tidy.run(_)
    Path(new_pdb_file_path).write_text(os.linesep.join(merged))
```

I think there is no typo bug, the function should work as is. Also, your function is not closing the input files for you. While `pdb_merge` does it :+1:

src/haddock/gear/read_io.py
```diff
@@ -0,0 +1,22 @@
...
+from haddock.libs.libontology import Format, PDBFile
+
+
+def load_from_previous(io_output):
+    """Load previous IO."""
+    models_to_refine = []
+    if not all([isinstance(e, PDBFile) for e in io_output]):
```
joaomcteixeira, line 8: you don't need to create a list:

```python
if not all(isinstance(e, PDBFile) for e in io_output):
```

src/haddock/modules/refinement/emref/__init__.py
```diff
@@ -93,11 +94,12 @@ def run(self, **params):
         jobs = []
 
         # Get the models generated in previous step
-        models_to_refine = [
-            p
-            for p in self.previous_io.output
-            if p.file_type == Format.PDB
-            ]
+        models_to_refine = load_from_previous(self.previous_io.output,
```
joaomcteixeira, line 97: Does this work? `check_balance` is not defined in the function definition.

src/haddock/modules/refinement/merge/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+            input_dic[i] = []
+            for key in model_dic:
+                input_dic[i].append(model_dic[key])
+
+        # Generate all combinations between the ensembles
+        models_to_be_merged = [
+            values for values in product(*input_dic.values())
+            ]
+        merged_models = []
+        for i, model_list in enumerate(models_to_be_merged):
```
joaomcteixeira, line 46: here use `enumerate(models_to_be_merged, start=1)` to avoid saving 0-indexed files to the system.

src/haddock/modules/refinement/merge/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+            pdb_fname = Path(self.path, f"merged_{i}.{Format.PDB}")
+            pdb_list = [e.full_name for e in model_list]
+
+            libpdb.merge(pdb_list, pdb_fname)
+
+            topology = list(chain(*[e.topology for e in model_list]))
```
joaomcteixeira, line 52: Is `model_list` a list? Why not `topology = [e.topology for e in model_list]` . Maybe I am missing something.

src/haddock/modules/refinement/merge/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+
+            pdb = PDBFile(pdb_fname, topology, path=self.path)
```
joaomcteixeira, line 54: Is `topology` parameter receiving a list?

src/haddock/modules/sampling/rigidbody/__init__.py
```diff
@@ -148,7 +153,9 @@ def run(self, **params):
 
                 # Create a model for the expected output
                 model = PDBFile(output_pdb_fname, path=self.path)
-                model.topology = [e.topology for e in combination]
+                # FIXME: Investigate this topology structure and simplify
+                topology_list = [e.topology for e in combination]
+                model.topology = list(chain(*topology_list))
```
joaomcteixeira, line 158: This is redundant. `model.topology = topology_list` is enough.

src/haddock/modules/topology/topoaa/__init__.py
```diff
@@ -97,51 +97,57 @@ def run(self, molecules, **params):
             log.info(f"Split models if needed for {step_molecule_path}")
             splited_models = libpdb.split_ensemble(step_molecule_path)
 
-            # nice variable name, isn't it? :-)
```
joaomcteixeira, line 100: I see you didn't like my comments :-P

joaomcteixeira (changes requested):
some changes for cleaning and improving

amjjbonvin (approved):
I will test once merged :-)

I see you are also including my CNS module changes into this PR.
Is this coming from my haddock3-example?

rvhonorato:
> I will test once merged :-)
> 
> I see you are also including my CNS module changes into this PR. Is this coming from my haddock3-example?

yes! used it to test the refinement

amjjbonvin:
So it will be part of your merge, right?

rvhonorato:
@amjjbonvin caught a bug in which the capri metrics would be wrong, this was because `libpdb.get_pdb_file_suffix_variations` returned `list(abs_path.glob(f"{basename.stem}{sep}*{basename.suffix}"))` and [glob returns an arbitrary order list](https://stackoverflow.com/a/6773661), which resulted in the metric being evaluated over swapped chains. 

I added a sort 😉

src/haddock/gear/read_io.py
rvhonorato: Your call

src/haddock/libs/libpdb.py
rvhonorato: this is `pdbtools`, check on the top `from pdbtools.pdb_merge import concatenate_files`

src/haddock/modules/refinement/emref/__init__.py
rvhonorato: ops forgot to delete it

src/haddock/modules/refinement/merge/__init__.py
rvhonorato: it can be a nested list if the model has multiple chains

src/haddock/modules/refinement/merge/__init__.py
rvhonorato: same as my comment above

src/haddock/modules/sampling/rigidbody/__init__.py
rvhonorato: `topology_list` might be a nested list

rvhonorato:
> So it will be part of your merge, right?

depends on who does it first! if we merge this pr you'll need to `update from master` but if you do it, then I need to update

amjjbonvin:
One more thing about this refinement refactoring: In principle we want to refine a complex and generate multiple models (e.g. 20 or 50 per starting conformation), using a different random seed every time.

This means that when going from `topoaa` to e.g. `mdref` one must be able to specify the `sampling` parameter.

amjjbonvin:
Also another strange behaviour which for some reason does not happen in the docking-protein-protein example is a duplication of PDB files when giving the models as separate molecules:

```
(haddock3) bash-3.2$ ls -1
total 9528
 176 e2a-hpr_1GGR_A.pdb
  72 e2a-hpr_1GGR_A_A.inp
5192 e2a-hpr_1GGR_A_A.out
 224 e2a-hpr_1GGR_A_A_haddock.pdb
 344 e2a-hpr_1GGR_A_A_haddock.psf
 224 e2a-hpr_1GGR_A_haddock.pdb
 104 e2a-hpr_1GGR_B.pdb
  72 e2a-hpr_1GGR_B_B.inp
2664 e2a-hpr_1GGR_B_B.out
 128 e2a-hpr_1GGR_B_B_haddock.pdb
 192 e2a-hpr_1GGR_B_B_haddock.psf
 128 e2a-hpr_1GGR_B_haddock.pdb
```

I am giving as input `e2a-hpr_1GGR_A.pdb` and `e2a-hpr_1GGR_B.pdb`, but as you can see four haddock pdb files are created.

rvhonorato:
> This means that when going from `topoaa` to e.g. `mdref` one must be able to specify the `sampling` parameter.

Sure, but I could not find the name of this parameter in the CNS, which one should it be?

amjjbonvin:
It is not a CNS parameter, but the python code must generate separate input files for each model request, even if only one starting model is provided.

I.e. it would be the sampling parameter of the other modules…

rvhonorato:
> Also another strange behaviour which for some reason does not happen in the docking-protein-protein example is a duplication of PDB files when giving the models as separate molecules:

Ops, forgot to remove those as well, fixed in the last commit;

```bash
run1/00_topoaa [173●] » ls -1
e2a-hpr_1GGR_ens.pdb
e2a-hpr_1GGR_ens_1.pdb
e2a-hpr_1GGR_ens_1_A.inp
e2a-hpr_1GGR_ens_1_A.out
e2a-hpr_1GGR_ens_1_A_haddock.psf
e2a-hpr_1GGR_ens_1_B.inp
e2a-hpr_1GGR_ens_1_B.out
e2a-hpr_1GGR_ens_1_B_haddock.psf
e2a-hpr_1GGR_ens_1_haddock.pdb
e2a-hpr_1GGR_ens_2.pdb
e2a-hpr_1GGR_ens_2_A.inp
e2a-hpr_1GGR_ens_2_A.out
e2a-hpr_1GGR_ens_2_A_haddock.psf
e2a-hpr_1GGR_ens_2_B.inp
e2a-hpr_1GGR_ens_2_B.out
e2a-hpr_1GGR_ens_2_B_haddock.psf
e2a-hpr_1GGR_ens_2_haddock.pdb
io.json
```

PS: using Bash? 😄

amjjbonvin:
Still - why split the files when they contain multiple chains?

And then merge then automatically, while if we provide separated chains you then need to add a merge step in the config file...

rvhonorato:
> It is not a CNS parameter, but the python code must generate separate input files for each model request, even if only one starting model is provided. I.e. it would be the sampling parameter of the other modules…

Oh allright, we should name it in a way that it's clear that it is not a substitute for `seletop`, shall we do it per-model? then we can name it `conformation_sampling` ?

rvhonorato:
Or expect people to properly split their multi-chained ensembles before inputting for scoring. 

I think that the way it is is more convenient since it supports both, but up to rule # 1!

src/haddock/libs/libpdb.py
joaomcteixeira: Yes, but your code repeats the `concatenate_files`. You are sending one file at the time, and getting one line at the time. Also `pdbmerge` is quite simple, if you don't want to use it, it fine. But the present code is using and desusing it :wink:

joaomcteixeira:
Wait, don't merge yet. Rodrigo `sort` is nice, but we have already another function to help the `sort` further for `_1` like files. to avoid `_1`, `_10`, etc, and have `_1`, `_2`... We can add this here also to avoid bugs. Because I implemented that in another PR as well.

amjjbonvin:
Sounds like a good choice.

But where would you input it? In the `mdref` module params?

> then we can name it conformation_sampling ?

amjjbonvin:
But do you also split the files for scoring???

And then merge them again…

Why not simply give the complex directly to the module?  Why this complexity?

Again the CNS `topoaa` modules can handle models with multiple chains in them.

rvhonorato:
Ok, my implementation relied heavily on the idea of supporting both multi-chain ensembles and single-chain ensembles and of course single-chain molecules and I changed the python engine around this.

So I am going to delete code this and do it again and taking into account your comments, thanks!

rvhonorato:
> Why not simply give the complex directly to the module? Why this complexity?

To generically handle all possible input methods without cross-module dependency and support all input types: Single-chain Model, Single-chain ensemble, Multi-chain Model and Multi-chain ensemble.

But as I said, I already scratched this code and will try to think of a different implementation.

joaomcteixeira:
Try to find a common path for all of them. For example, converting all to an ensemble and then moving from there. Or if not; these are just 4 cases. Implement 4 functions for each case. It is a well-defined problem. It won't grow beyond this.

PR #178, opened 2021-12-06
Add integration tests

rvhonorato:
🚧 🚧 🚧 🚧 🚧


tests/modules/rigidbody/test_rigidbody.py
```diff
@@ -0,0 +1,98 @@
+"""Test rigidbody module."""
+import json
+from pathlib import Path
+
+from haddock.modules.sampling.rigidbody import HaddockModule as RigidBody
+
+
+RECIPE_PATH = Path(
```
joaomcteixeira, line 8: you can import these directly
`from haddock.modules.sampling.rigidbody import DEFAULT_CONFIG`

tests/modules/rigidbody/test_rigidbody.py
```diff
@@ -0,0 +1,98 @@
...
+
+    rigidbody.run()
+
+    # Test things ================================ #
+    # Test if files were generated
+    inp_file = Path(DATA_PATH, 'rigidbody_1.inp')
+    out_file = Path(DATA_PATH, 'rigidbody_1.out')
+    pdb_file = Path(DATA_PATH, 'rigidbody_1.pdb')
+
+    assert inp_file.exists() is True
```
joaomcteixeira, line 79: don't need to `is True`. `assert inp_file.exists()` suffices

tests/modules/rigidbody/test_rigidbody.py
```diff
@@ -0,0 +1,98 @@
...
+    with open(io, 'w') as io_fh:
+        json.dump(io_data, io_fh)
+
+
+def cleanup():
+    """Clean the files generated by the test."""
+    files_to_be_removed = [
+        'rigidbody_1.inp', 'rigidbody_1.out', 'rigidbody_1.pdb',
+        'WARNING', 'io.json']
+    for element in files_to_be_removed:
```
joaomcteixeira, line 54:
```python
for element in files_to_be_removed:
    f = Path(DATA_PATH, element).resolve()
    f.unlink(missing_ok=True)
```

tests/modules/rigidbody/test_rigidbody.py
```diff
@@ -0,0 +1,98 @@
...
+
+def test_rigidbody():
+    """Test the rigidbody module."""
+    initialize()
+
+    rigidbody = RigidBody(
+        order=1, path=DATA_PATH, initial_params=DEFAULT_CONFIG
+        )
+
+    rigidbody.params.update(CONF_PARAMS)
```
joaomcteixeira, line 68: don't do this. send the params to the `run` instead. During tests we should try to use the APIs, instead of injecting things under the hood.
```python
rigidbody.run(sampling=1, **CONF_PARAMS)
```

tests/modules/rigidbody/test_rigidbody.py
```diff
@@ -0,0 +1,98 @@
...
+DEFAULT_CONFIG = Path(RECIPE_PATH, "defaults.cfg")
+DATA_PATH = Path(Path(__file__).resolve().parent, 'data')
+
+CONF_PARAMS = {
+    'config_path': DATA_PATH,
+    'mode': 'local'
+    }
+
+
+def initialize():
```
joaomcteixeira, line 21: Why don't you create an initialized `json` directly instead of having  a function to create it? what are you testing? I would suggest creating the function and loading the json in the test.

joaomcteixeira (commented):
It's getting its shape :wink: some comments for improving, imho

tests/modules/rigidbody/test_rigidbody.py
rvhonorato: Because its my first time with pytest and I don't know how its best to do things 😄

joaomcteixeira: pytests just runs functions and see if they raise any exception. normally with pytest, you receive some parameters (optionally) and assert a result against an expectation. Or, you run something and see if some errors come out. Go for it :+1:

rvhonorato:
Nuking this once since it's gone stale and we should anyway re-think on how to do this tests.

PR #182, opened 2021-12-07
Add option to "sortby" in `caprieval`

### LINKED ISSUE #153: Add option to "sortby" in `caprieval`
We could sort the output of `caprieval` by `score` or some of the calculated metrics.

rvhonorato:
I re-implemented the output function [because life is better with pandas](https://images-na.ssl-images-amazon.com/images/I/61a0D1YzjNL.jpg) 🐼 

***
This closes #153.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,5 @@ pdb-tools==2.1.0
 jsonpickle==2.0.0
 numpy==1.20.2
 tox==3
+pandas==1.3.4
 git+git://github.com/joaorodrigues/fcc@fcc2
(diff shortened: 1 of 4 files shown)


amjjbonvin (approved):
Nice option!

joaomcteixeira:
I actually strongly disagree we add `pandas` to the requirements. It's a very heavy dependency that we need to manage on top of everything and breaks one of the features we are working hard to have in haddock3, which is having a lightweight python-shell that people can install to use haddock3's features in servers or locally without the need to install the simulation machinery while keeping everything lightweight. I would do an internal sorting function.

rvhonorato:
Reverted on #184 and reopened #153

Issue #186, opened 2021-12-08
Add sampling parameter to refinement modules


rvhonorato:
Should be similar to what is done for `rigidbody`

rvhonorato:
Should we add this parameter per-model or in total @amjjbonvin? 

To me it makes more sense to do it per-model:
`samping_factor = 20` would sample each conformation 20 times etc

Can quickly explode but we can also add a limiter specially since it might not make too much sense to sample 1 conformation 1000 times?

amjjbonvin:
Per model makes indeed more sense.

But would that possibly also affect the regular docking workflow?
You don’t want to have the number of `rigidbody` models increased by that factor.
Or the default should be 1

rvhonorato:
yes we can set the default to 1 so it would not affect anything, but then we can get creative with it and try some sort of "supersampling" protocol 😄 

`rigidbody > select top 10 > refine with factor 10`

amjjbonvin:
We have now much more flexibility indeed! Nice

Or cluster after rigid body and then refine

PR #187, opened 2021-12-08
Refactor code to allow for Refinement

### LINKED ISSUE #173: Refactor CNS refinement modules to be executed after `topoaa`

rvhonorato:
This is a re-do of #176.

I took a different, more complex approach making more derivative changes instead of changing it in the base (`topoaa`).

This PR is specifically for the changes to the code to allow for refinement, I'll add the sampling in a different issue/pr - https://github.com/haddocking/haddock3/issues/186

***
closes #173


src/haddock/libs/libcns.py
```diff
@@ -22,10 +24,19 @@ def generate_default_header(protonation=None):
...
-    water_box = load_waterbox(cns_paths.water_box['boxtyp20'])
-
-    return (param, top, link, topology_protonation, trans_vec, tensor, scatter,
-            axis, water_box)
+    water_box = load_waterbox(cns_paths.water_box["boxtyp20"])
+
+    return (
+        param,
+        top,
+        link,
```
amjjbonvin, line 32: Is there a water box generation step at the python level? Isn't that handled by the CNS module?

src/haddock/libs/libcns.py
```diff
@@ -117,42 +126,42 @@ def write_eval_line(param, value, eval_line='eval (${}={})'):
...
 
 
 def load_ff_parameters(forcefield_parameters):
     """Add force-field specific parameters to its appropriate places."""
-    ff_param_header = f'{linesep}! FF parameters{linesep}'
-    ff_param_header += f'parameter{linesep}'
-    ff_param_header += f'  @@{forcefield_parameters}{linesep}'
-    ff_param_header += f'end{linesep}'
+    ff_param_header = f"{linesep}! FF parameters{linesep}"
+    ff_param_header += f"parameter{linesep}"
```
amjjbonvin, line 137: Note that there is no need anymore to write the reading of `haddock.top` and `haddock.param` to the CNS input files. This is now handled by the `read_param.cns` script in each module

src/haddock/libs/libcns.py
```diff
@@ -117,42 +126,42 @@ def write_eval_line(param, value, eval_line='eval (${}={})'):
...
 
 
 def load_ff_topology(forcefield_topology):
     """Add force-field specific topology to its appropriate places."""
-    ff_top_header = f'{linesep}! Toplogy{linesep}'
-    ff_top_header += f'topology{linesep}'
-    ff_top_header += f'  @@{forcefield_topology}{linesep}'
-    ff_top_header += f'end{linesep}'
+    ff_top_header = f"{linesep}! Toplogy{linesep}"
+    ff_top_header += f"topology{linesep}"
```
amjjbonvin, line 147: Same comment as before - the only top file required to be passed to the CNS input files is for ligands if present.

amjjbonvin (approved):
Just test and working fine.

The `conformational_sampling` is not yet implemented, correct? At least I don't find it in the `default.cfg` file of the refinement module

src/haddock/libs/libcns.py
rvhonorato: It comes from this one, but I think this has always been there, is it needed?
```python
def load_waterbox(waterbox_param):
    """Add waterbox information."""
    water_header = f'{linesep}! Water box{linesep}'
    water_header += f'eval ($boxtyp20 = "{waterbox_param}" ){linesep}'

    return water_header
```

rvhonorato:
> Just test and working fine.
> 
> The `conformational_sampling` is not yet implemented, correct? At least I don't find it in the `default.cfg` file of the refinement module

not yet! I'll do it in another PR after this one

src/haddock/libs/libcns.py
rvhonorato: removed!

src/haddock/libs/libcns.py
rvhonorato: removed!

joaomcteixeira (approved):

PR #159, opened 2021-11-28
Change log statements

rvhonorato:
I changed the log statements so it display information that is more relevant to the way we usually look at the HADDOCK execution (also more in line with what we are using in 2.x).

I also made it display the % of completion, this is how it looks now with this PR.

```text
[2021-11-28 15:50:03,299 cli INFO] 
##############################################
#                                            #
#                 HADDOCK 3                  #
#                                            #
##############################################

Starting HADDOCK 3.0.0 on 2021-11-28 15:50:00

Python 3.9.0 | packaged by conda-forge | (default, Nov 26 2020, 07:54:06) 
[Clang 11.0.0 ]

[2021-11-28 15:50:05,623 libutil WARNING] /Users/rodrigo/repos/haddock3/examples/docking-protein-protein/run1 exists and it will be REMOVED!
[2021-11-28 15:50:05,663 libworkflow INFO] Reading instructions of [topoaa] step
[2021-11-28 15:50:05,663 libworkflow INFO] Reading instructions of [rigidbody] step
[2021-11-28 15:50:05,663 libworkflow INFO] Reading instructions of [caprieval] step
[2021-11-28 15:50:05,663 libworkflow INFO] Reading instructions of [seletop] step
[2021-11-28 15:50:05,664 libworkflow INFO] Reading instructions of [flexref] step
[2021-11-28 15:50:05,664 libworkflow INFO] Reading instructions of [caprieval.1] step
[2021-11-28 15:50:05,664 libworkflow INFO] Reading instructions of [emref] step
[2021-11-28 15:50:05,665 libworkflow INFO] Reading instructions of [caprieval.2] step
[2021-11-28 15:50:05,670 __init__ INFO] Running [topoaa] module
[2021-11-28 15:50:05,671 __init__ INFO] [topoaa] Generating topologies
[2021-11-28 15:50:05,671 __init__ INFO] [topoaa] Molecule 1: e2aP_1F3G.pdb
[2021-11-28 15:50:05,678 __init__ INFO] [topoaa] Sanitizing molecule e2aP_1F3G.pdb
[2021-11-28 15:50:05,696 __init__ INFO] [topoaa] CNS input created at 00_topoaa/e2aP_1F3G.inp
[2021-11-28 15:50:05,696 __init__ INFO] [topoaa] Molecule 2: hpr_ensemble.pdb
[2021-11-28 15:50:05,725 __init__ INFO] [topoaa] Sanitizing molecule hpr_ensemble_1.pdb
[...]

PR #188, opened 2021-12-08
Tweak logs

rvhonorato:
Small tweak to the logs, think we missed some in #159


joaomcteixeira (approved):
I imagine many more of these will come. That's good :relaxed:

Did you notice the `self.log` stuff added in #159 ? :relaxed: Now you get the module name in the log and the parent `run` works as a decorator for child `_run` of each module :wink: very handy and cool.

rvhonorato:
> I imagine many more of these will come. That's good ☺️
> 
> Did you notice the `self.log` stuff added in #159 ? ☺️ Now you get the module name in the log and the parent `run` works as a decorator for child `_run` of each module 😉 very handy and cool.

I did yes, very nice! How do you add the warnings and errors?

PR #191, opened 2021-12-08
Do not delete the run dir.

joaomcteixeira:
I know it is very boring to be deleting folders when we are running tests, but it is just too dangerous to have haddock3 deleting folders without asking.

### DIFF
--- a/src/haddock/gear/greetings.py
+++ b/src/haddock/gear/greetings.py
@@ -5,7 +5,7 @@ import sys
 from datetime import datetime
 from functools import partial
 
-from haddock import version
+from haddock import contact_us, version
 
 
 international_good_byes = [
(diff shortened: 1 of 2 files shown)


amjjbonvin (approved):

rvhonorato:
Any opposition to have a `--force` option?

joaomcteixeira:
I would say we may forget to `--force` option turned on, or go by inertia giving `--force` all the time, as it happens for other software. You can always do `rm -r rundir1 && haddock ...cfg`. I really vote *no* for the `--force`.

rvhonorato:
you are the boss!

rvhonorato (approved):

rvhonorato:
What is missing in this one?

joaomcteixeira:
Nothing. I was just waiting because we are in very active development of the modules themselves so I thought merging it now would hinder more than help. But if that is not the case, go ahead and merge it. :+1:

rvhonorato:
> You can always do rm -r rundir1 && haddock

😉

joaomcteixeira:
Answering your question @rvhonorato about the fork, you can actually use `--restart 0`. It will do the same as the `--force`.

PR #193, opened 2021-12-13
Fix HPC completed percentage

rvhonorato:
Ops made a small error when calculating the % of completed jobs.


joaomcteixeira (approved):
:+1:

Issue #194, opened 2021-12-13
Create a flexible benchmark scenario loader


rvhonorato:
Now the benchmarking scenarios are hardcoded in `cli_bm`, we should have a more flexible way of defining such scenarios.

joaomcteixeira:
> https://github.com/haddocking/haddock3/pull/195#pullrequestreview-831363521

I thought a lot about where to put the configuration templates when I designed the CLI. You are right to say that haven the templates inside the CLI might not be the best. But I am not sure if it is best/feasible having them outside functions (for example plain text files). However, we could have the template scenario configs in a module outside the CLI so that the client can import them. But I trust they should be functions that return a formatted string. That's pretty flexible and easy to document with the docstring.

rvhonorato:
Addressed with https://github.com/haddocking/haddock-runner

PR #195, opened 2021-12-13
Small tweaks to the benchmark scripts

rvhonorato:
Made some very minor changes to the benchmark scripts, mostly changing the name of some variables and removing one of the `create_cfg_scn_2` which did not make much sense. 

But doing so I just noticed that the benchmark scenarios are hardcoded in the script, this is not very practical and we should have a more flexible way of defining it, let's work on it on https://github.com/haddocking/haddock3/issues/194

### DIFF
--- a/src/haddock/clis/cli_bm.py
+++ b/src/haddock/clis/cli_bm.py
@@ -30,9 +30,10 @@ daemon without directly using the `haddock3-dmn` client.
 USAGE:
     haddock3-bm -h
     haddock3-bm <BM dir> <output dir> [OPTIONS]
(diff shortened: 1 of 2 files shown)


joaomcteixeira (approved):
I agree with the changes here. :+1:

I thought a lot about where to put the configuration templates when I designed the CLI. You are right to say that haven the templates inside the CLI might not be the best. But I am not sure if it is best/feasible having them outside functions (for example plain text files). However, we could have the template scenario configs in a module outside the CLI so that the client can import them. But I trust they should be functions that return a formatted string. That's pretty flexible and easy to document with the docstring. #194

PR #203, opened 2021-12-19
Updates clarifies install instructions

joaomcteixeira:
Updates and clarifies install instructions.


amjjbonvin (approved):
Approved.

We will also have to add possible some instructions about CNS compilations. But this could be into a separate CNS-code directory. In haddock2.4 we are providing a set of routines with which CNS must be recompiled to have all the proper functionalities.

rvhonorato (approved):

PR #209, opened 2021-12-20
Write `ambig_fname` as full path

### LINKED ISSUE #208: Make `libcns` use all supported restraints
Currently `libcns` supports various restraints but only `ambig` is written via the `prepare_cns_input` function, we should thus refactor this function to take all possible restraints into account when preparing the `.inp`.

Here they should use full paths.

rvhonorato:
This closes #208, note that for now `libcns` only writes `ambig.tbl`, for it to use the other supported restraints we need to refactor the `prepare_cns_input` function.

### DIFF
--- a/src/haddock/libs/libcns.py
+++ b/src/haddock/libs/libcns.py
@@ -387,6 +387,7 @@ def prepare_cns_input(
     input_str = prepare_multiple_input(pdb_list, psf_list)
 
     if ambig_fname:
+        ambig_fname = Path(ambig_fname).resolve()
         ambig_str = load_ambig(ambig_fname)
     else:
         ambig_str = ""


amjjbonvin:
> This closes #208 <https://github.com/haddocking/haddock3/issues/208>, note that for now libcns only writes ambig.tbl, for it to use the other supported restraints we need to refactor the prepare_cns_input function.
> 

Doesn’t it write also `unambig` and `hbonds` and `dihedrals` ???
Those are defined in the default cfg files and should thus be supported.

amjjbonvin (approved):
Seems straightforward to also add unambig_fname and hbond_fname

rvhonorato:
> This closes #208 <#208>, note that for now libcns only writes ambig.tbl, for it to use the other supported restraints we need to refactor the prepare_cns_input function.
> Doesn’t it write also `unambig` and `hbonds` and `dihedrals` ??? Those are defined in the default cfg files and should thus be supported.

Yes they are supported but they are not written to `.inp`, will close this one and refactor `libcns` to also write them.

amjjbonvin:
Isn’t anything defined in default.cfg written to the inp file?

rvhonorato:
> Isn’t anything defined in default.cfg written to the inp file?

Yes, whatever values are in the defaults are written to the `.inp`, but as it is now it will only copy the default value, meaning whatever you enter via the configuration file will not be passed.

PR #211, opened 2021-12-20
Fix `libparallel` to correctly use `ncores`

### LINKED ISSUE #210: `libparallel` not running jobs with multiple `ncores`
> @joaomcteixeira  11:08
> the ncores parameter are passed correctly to the Scheduler the error comes after the @rvhonorato's #202 .

rvhonorato:
This closes #210

### DIFF
--- a/src/haddock/libs/libparallel.py
+++ b/src/haddock/libs/libparallel.py
@@ -6,6 +6,14 @@ from haddock import log
 from haddock.libs.libutil import parse_ncores
 
 
+def split_tasks(lst, n):
+    """Split tasks into N-sized chunks."""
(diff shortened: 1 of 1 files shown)


joaomcteixeira (commented):
I would suggest `split_tasks` becomes an independent function instead of a `@staticmethod`. Naive question, why do we need to fill the last list with `None` ?

rvhonorato:
Sure, where should it go?

joaomcteixeira:
in the same `libparallel` i would say

joaomcteixeira (approved):

amjjbonvin (approved):

PR #214, opened 2021-12-21
Export environment variables to a file per module.

joaomcteixeira:
To facilitate running the `.inp` CNS scripts independently from the HADDOCK3 shell we need a file with the environment variables that can be sourced independently. FYI, when the HADDOCK3's python shell runs the environment variables are sent directly to the `subprocess.Popen`.

Saving the `envvars` to a file is now part of the `BaseHaddockModule` class and saving triggers automatically for all Modules that use CNS. The new file is called `envvars` and is inside each modules folder `00_`, `01_`, etc; as long as that module uses CNS.

The `envvars` file has a `COMMON_PATH_HD3` variable that can be changed easily in case we transport the haddock installation and the examples folder to another computer.

`source envvars` and you are good to go.

Example:

```bash
#!/bin/bash
export COMMON_PATH_FOR_HD3=/root/user/haddock3
export MODULE=${COMMON_PATH_FOR_HD3}/src/haddock/modules/topology/topoaa/cns
export MODIR=${COMMON_PATH_FOR_HD3}/examples/refine-complex/run1/00_topoaa
export RUN=${COMMON_PATH_FOR_HD3}/examples/refine-complex/refine-complex.cfg
export TOPPAR=${COMMON_PATH_FOR_HD3}/src/haddock/cns/toppar
```

### DIFF
--- a/src/haddock/modules/__init__.py
+++ b/src/haddock/modules/__init__.py
@@ -168,19 +168,20 @@ class BaseHaddockModule(ABC):
         getattr(log, level)(f'[{self.name}] {msg}')
 
     def default_envvars(self, **envvars):
-        """Save envvars to disk."""
+        """Return default env vars updated to `envvars` (if given)."""
         default_envvars = {
             'MODULE': self.cns_folder_path,
             'MODIR': self.path,
             'RUN': self.params["config_path"],
             'TOPPAR': global_toppar,
             }
+
         default_envvars.update(envvars)
 
         return default_envvars
 
     def save_envvars(self, filename='envvars', **envvars):
-        """Save envvars to file."""
+        """Save envvars needed for CNS to a file in the module's folder."""
         common_path = os.path.commonpath(list(envvars.values()))
 
         envvars = {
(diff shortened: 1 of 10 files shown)


joaomcteixeira:
Tested all the examples, and they all run. Tested doing `cns < .inp > out` after sourcing `envvars` and it works.

src/haddock/modules/__init__.py
```diff
@@ -163,6 +167,40 @@ def log(self, msg, level='info'):
         """
         getattr(log, level)(f'[{self.name}] {msg}')
 
+    def default_envvars(self, **envvars):
+        """Return default env vars updated to `envvars` (if given)."""
+        default_envvars = {
+            "MODULE": self.cns_folder_path,
+            "MODIR": self.path,
```
amjjbonvin, line 174: Isn't is called rather MODDIR?

amjjbonvin (changes requested):
Please correct MODIR to MODDIR - this is what is currently be used (also written to the job files in HPC mode

rvhonorato (approved):

amjjbonvin (approved):
These relative path might also come handy when setting up runs to run under `/tmp` on a node

joaomcteixeira:
Added option to write relative paths  to `envvars`. Paths will be relative to `MODDIR` where `MODDIR` is the modules folder (`run1/00_topoaa`) for example. This expects the `source envvars` to be done in the module folder.

Also, in the `config files` you can give the general parameter `relative_envvars = true/false` to create relative or absolute paths in the `ennvvars`.

PR #223, opened 2021-12-22
Solves issue 220

amjjbonvin:
Fixed the flexible refinement issue
Updated protein-peptide example
Some cosmetics on CNS scripts

Note that the flexem is now working we are running into the problem that some models fail and the run then stops.


amjjbonvin:
Also added the `dihe_fname` parameter to all relevant `default.cfg` files

joaomcteixeira (approved):
> Also added the dihe_fname parameter to all relevant default.cfg files

Yes, it's important that all parameters related to file paths end with `_fname`, so that the #219 works.

PR #226, opened 2021-12-22
Edit `libhpc` with new variable scheme

### LINKED ISSUE #225: `libhpc` broken after exporting environment variables to a file per module
Introduced with #214

rvhonorato:
Closes #225


joaomcteixeira (approved):
I have not tested it. I think it is enough as is.

PR #238, opened 2021-12-30
add run examples scripts

joaomcteixeira:
Add script to run examples in a row. Why not have this here for now? :relaxed:

Usage (from within the `examples/` folder):

```bash
./run_examples.sh 0  # -> runs all examples regardless of errors
./run_examples.sh 1  # -> run all examples but will stop if an error happens
```


rvhonorato (approved):

amjjbonvin (approved):
You could take 0 as default in case of no argument given...

examples/run_examples.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+# Script used to run all examples at once.
+# Does not parse examples automatically. The exact examples that are run
+# are defined here in the script. Run this script from inside the `examples`
+# folder.
+#
+# USAGE:
```
amjjbonvin, line 8: I would implement a `-h` argument that returns the usage :-)

amjjbonvin (changes requested):

joaomcteixeira:
@amjjbonvin made it in python at the end.

`python run_examples.py -h`

you need to have the `haddock3` environment activated. But that also happened with the previous `sh` version. Any new examples must be added to the pipe manually by editing the script. I did not make an official CLI out of this because it only makes sense to run it inside `examples/` and because of the hardcoded examples (it should be this way).

amjjbonvin (approved):

examples/run_examples.sh
joaomcteixeira: :dark_sunglasses:

rvhonorato:
Python 1 x Bash 0! 😉

amjjbonvin:
What about the carbon footprint of the two? Which version uses more characters / lines to do the same?

joaomcteixeira:
:rofl: i needed to add more chars because of the `-h` option :stuck_out_tongue_closed_eyes: definitively I wrote the python version better.

PR #239, opened 2021-12-30
Implement unit tests for `caprieval`

rvhonorato:
This adds tests that should cover most of the capri (python) module of `caprieval`.


joaomcteixeira (approved):
There are three file that are generated during the tests:

```
#       tests/golden_data/blosum62.izone
#       tests/golden_data/blosum62_A.aln
#       tests/golden_data/blosum62_B.aln
```

Best would be to have these files removed. If that is not trivial, nevermind.

Apart from that, very nice implementation! Happy you're liking `pytest` :smile:

amjjbonvin (approved):
Did you check that the values we are calculating are matching (to some extend) what DockQ is reporting (for iRMSD, lRMSD and Fnat) - just as a control.

rvhonorato:
> Did you check that the values we are calculating are matching (to some extend) what DockQ is reporting (for iRMSD, lRMSD and Fnat) - just as a control.

I did a quick check with ProFit, values are a bit different since it does least-squares fitting and here we do simple kabsch rotation. But indeed we should do a larger-scale test to see the correlation.

PR #241, opened 2021-12-31
Update `clustfcc`

### LINKED ISSUE #240: Bring `clustfcc` up to date with the latest changes
We have plans to replace the engine with a new FCC version #133, but it might be faster to just update the current implementation.

rvhonorato:
This closes #240 by updating the `clustfcc` module with the latest changes, I also added a stub example `docking-protein-protein-clustfcc` with a low `fraction_cutoff` to ensure clusters are generated.

Another important note is that this module is **not compatible** with HPC since `HPCWorker` only handles `CNSJobs`.


amjjbonvin (changes requested):
Rather than decreasing the cutoff I would suggest to use a proper cutoff (not to confuse users), and change the behaviour to mimic what is happening in 2.4, namely reduce the required minimum number of members to define a cluster, from the default of 4 to 3,2 and eventually 1 (1 will always give clusters).

Also rather than creating a new example, I would simply adapt the current example, after all clustering should be the default behaviour.  (But for the tests you could distinguish eventually)

src/haddock/libs/libparallel.py
```diff
@@ -57,7 +57,12 @@ def __init__(self, tasks, ncores=None):
         #  so we know that 2 comes before 10
         task_name_dic = {}
         for i, t in enumerate(tasks):
-            task_name_dic[i] = t.input_file, len(str(t.input_file))
+            try:
+                task_name_dic[i] = t.input_file, len(str(t.input_file))
```
joaomcteixeira, line 61: I don't like this nomenclature. If that is a tuple, explicitly add a tuple

```python
(t.input_file, len(str(t.input_file)))
```

joaomcteixeira (changes requested):
The part in the `caprieval/__init__.py` is okay. But I don't like the other parts in the `libsubprocess` and `Job`. We should have a different class for that. What the `scheduler` requests to the `Job` should always be the same attribute. These kinds of switches should be handled by the specific subclass. 

Also, `Job` should not have the `arg_first` argument. We should have a subclass for that. Since we are using OOP, let's make use of it :relaxed:

rvhonorato:
> The part in the `caprieval/__init__.py` is okay. But I don't like the other parts in the `libsubprocess` and `Job`. We should have a different class for that. What the `scheduler` requests to the `Job` should always be the same attribute. These kinds of switches should be handled by the specific subclass.
> 
> Also, `Job` should not have the `arg_first` argument. We should have a subclass for that. Since we are using OOP, let's make use of it ☺️

Indeed! Opened a new one about it: https://github.com/haddocking/haddock3/issues/245

examples/docking-protein-protein/docking-protein-protein.cfg
```diff
@@ -55,5 +55,8 @@ noecv = true
 [caprieval]
 reference_fname = 'data/e2a-hpr_1GGR.pdb'
 
+[clustfcc]
+threshold = 1
```
amjjbonvin, line 59: Why threshold = 1? To make sure you are getting clusters? Why not implement the automatic reduction of the threshold until you do find clusters (as done in 2.4)?

amjjbonvin (commented):
I would really  implement the option to reduce the threshold automatically until some clusters do appear...

examples/docking-protein-protein/docking-protein-protein.cfg
rvhonorato: Good point! We'll do that in another issue: https://github.com/haddocking/haddock3/issues/253

PR #252, opened 2022-01-03
`gdock` update

rvhonorato:
This PR just updates the third-party module `gdock` to the latest code.


src/haddock/modules/sampling/gdock/__init__.py
```diff
@@ -114,19 +111,32 @@ def _run(self):
...
 
         # retrieve the structures
         output_structures = []
-        structure_folder = Path(self.path, 'gdock-integration/structures')
+        structure_folder = Path('gdock-integration/structures')
         for model in structure_folder.glob('*pdb'):
-            pdb = PDBFile(model, path=model.parent)
+            # Make sure the output is tidy, this should be handled
+            #  by gdock, but check it here to be sure
+            tmp_f = tempfile.NamedTemporaryFile(delete=False)
```
joaomcteixeira, line 134: I think is best if you write this with the following:

```python
from pdbtools import pdb_tidy 
with open(model, 'r') as fin:
    lines = list(pdb_tidy.run(fin, strick=False))  # be explicit in the `strict`

with open(model, 'w') as fout:
    fout.write(os.linesep.join(lines))
```
Don't know if the `os.linesep` is needed. With this, you avoid creating a temporary file. I think after #144 this part is no longer needed. I will have that in mind.

Aside from this, :rocket:

src/haddock/modules/sampling/gdock/__init__.py
```diff
@@ -114,19 +111,32 @@ def _run(self):
...
-            cmd = f'{sys.executable} {gdock_exec} --dry input.toml'
+        cmd = f'{sys.executable} {gdock_exec} --dry input.toml'
 
-            subprocess.call(cmd, shell=True)
+        subprocess.call(cmd, shell=True)
 
         # retrieve the structures
         output_structures = []
-        structure_folder = Path(self.path, 'gdock-integration/structures')
+        structure_folder = Path('gdock-integration/structures')
```
joaomcteixeira, line 130: Currently, `structure_folder` will be searched inside the `0*_gdock` folder in the `run_dir`. Make sure this is what you want, or if you need some `../data` first.

src/haddock/modules/sampling/gdock/__init__.py
rvhonorato: thanks! I keep trying to find reasons to use tempfile but there is never a good one 😢

src/haddock/modules/sampling/gdock/__init__.py
rvhonorato: yep took me a while to figure this out, as its now the next module can find the structures fine.

src/haddock/modules/sampling/gdock/__init__.py
joaomcteixeira: Yeah, we are always trying to reduce the disk usage, so... :stuck_out_tongue_closed_eyes:

src/haddock/modules/sampling/gdock/__init__.py
joaomcteixeira: After #219 each module runs on its own folder. We need to keep this in mind now when defining paths.

PR #255, opened 2022-01-04
Implement automatic threshold in `clustfcc`

### LINKED ISSUE #253: Implement the automatic threshold reduction in `clustfcc`
Why threshold = 1? To make sure you are getting clusters? Why not implement the automatic reduction of the threshold until you do find clusters (as done in 2.4)?

_Originally posted by @amjjbonvin in https://github.com/haddocking/haddock3/pull/241#discussion_r777895960_

rvhonorato:
This PR closes #253 by adding a logic to **decrease** the threshold in case no clusters were found. This changed threshold is then passed back to the parameter dictionary so it can be written in the detailed `clustfcc.txt`.

### DIFF
--- a/src/haddock/modules/analysis/clustfcc/__init__.py
+++ b/src/haddock/modules/analysis/clustfcc/__init__.py
@@ -104,10 +104,27 @@ class HaddockModule(BaseHaddockModule):
             self.params['strictness'],
             )
 
-        _, clusters = cluster_fcc.cluster_elements(
-            pool,
-            threshold=self.params['threshold'],
-            )
+        cluster_check = False
+        while not cluster_check:
+            for threshold in range(self.params['threshold'], 0, -1):
+                log.info(f'Clustering with threshold={threshold}')
+                _, clusters = cluster_fcc.cluster_elements(
+                    pool,
+                    threshold=threshold,
+                    )
+                if not clusters:
+                    log.info(
+                        "[WARNING] No cluster was found, increasing threshold!"
+                        )
+                else:
+                    cluster_check = True
+                    # pass the actual threshold back to the param dict
+                    #  because it will be use in the detailed output
+                    self.params['threshold'] = threshold
+                    break
+            if not cluster_check:
+                # No cluster was obtained in any threshold
+                cluster_check = True
 
         # Prepare output and read the elements
         clt_dic = {}
--- a/src/haddock/modules/analysis/clustfcc/__init__.py
+++ b/src/haddock/modules/analysis/clustfcc/__init__.py
@@ -114,7 +114,7 @@ class HaddockModule(BaseHaddockModule):
                     )
                 if not clusters:
                     log.info(
-                        "[WARNING] No cluster was found, increasing threshold!"
+                        "[WARNING] No cluster was found, decreasing threshold!"
                         )
                 else:
                     cluster_check = True
--- a/examples/docking-protein-protein/docking-protein-protein.cfg
+++ b/examples/docking-protein-protein/docking-protein-protein.cfg
@@ -56,7 +56,6 @@ noecv = true
 reference_fname = 'data/e2a-hpr_1GGR.pdb'
 
 [clustfcc]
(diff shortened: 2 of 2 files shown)


src/haddock/modules/analysis/clustfcc/__init__.py
```diff
@@ -104,10 +104,27 @@ def _run(self):
             self.params['strictness'],
             )
 
-        _, clusters = cluster_fcc.cluster_elements(
-            pool,
-            threshold=self.params['threshold'],
-            )
+        cluster_check = False
```
joaomcteixeira, line 107: I strongly suggest writing the conditionals always with positive statements.

```python
cluster_missing = True
while cluster_missing:
    for ....
          if clusters:
               cluster_missing = False
               ...
          else:
              log(...)
```

i think with this, the `if not cluster_check` in line 125 is not even needed.

src/haddock/modules/analysis/clustfcc/__init__.py
```diff
@@ -104,10 +104,27 @@ def _run(self):
...
+        while not cluster_check:
+            for threshold in range(self.params['threshold'], 0, -1):
+                log.info(f'Clustering with threshold={threshold}')
+                _, clusters = cluster_fcc.cluster_elements(
+                    pool,
+                    threshold=threshold,
+                    )
+                if not clusters:
+                    log.info(
+                        "[WARNING] No cluster was found, increasing threshold!"
```
joaomcteixeira, line 117: `increasing` or `decreasing` ?

joaomcteixeira (changes requested):
Minor changes if you agree with me.

src/haddock/modules/analysis/clustfcc/__init__.py
amjjbonvin: Should be `decreasing` since we go from the default of 4 to 3, 2, and 1 (should not go lower than 1!)

joaomcteixeira: Yes, Rodrigo implemented it already like that. Just need to update the logging message.

src/haddock/modules/analysis/clustfcc/__init__.py
```diff
@@ -104,10 +104,27 @@ def _run(self):
             self.params['strictness'],
             )
 
-        _, clusters = cluster_fcc.cluster_elements(
-            pool,
-            threshold=self.params['threshold'],
-            )
+        cluster_check = False
+        while not cluster_check:
+            for threshold in range(self.params['threshold'], 0, -1):
```
joaomcteixeira, line 109: Here @amjjbonvin , it goes one by one down to 1.

amjjbonvin:
Ok - if you say so - I though it would go to 0…

src/haddock/modules/analysis/clustfcc/__init__.py
rvhonorato: ops!

rvhonorato:
> 



> Ok - if you say so - I though it would go to 0…

nope, it goes to 1:

```python
>>> list(range(4, 0, -1))
[4, 3, 2, 1]
```

src/haddock/modules/analysis/clustfcc/__init__.py
rvhonorato: In this logic it would get stuck if no cluster is generated, this only works if `threshold=1` always generates a cluster, right?

src/haddock/modules/analysis/clustfcc/__init__.py
rvhonorato: > Should be `decreasing` since we go from the default of 4 to 3, 2, and 1 (should not go lower than 1!)

Yes this is how its implemented, just made a typo in the log message.

joaomcteixeira (approved):

amjjbonvin:
Threshold 1 will always generate clusters since each model can be its own cluster

PR #260, opened 2022-01-05
Implement fault tolerance

### LINKED ISSUE #227: Allow runs to continue in case of a few failures
Currently the workflow will stop if one model fails. This is quite commonly happening (e.g. exploding models in the flexible refinement). HADDOCK2.4 has a mechanism to deal with that, restarting the job and only doing an EM step.

A simple alternative for the time being would be to allow a run to continue if less than a user-defined percentage fails (e.g. 5% as default). At least the workflow will complete in most cases, unless there is something seriously wrong and most models fail.

rvhonorato:
This PR closes #227 by adding a fault tolerance logic to the modules.

The new parameter `tolerance` is set in `defaults.cfg` as % and the check is made directly into `ModuleIO`.

### DIFF
--- a/src/haddock/libs/libontology.py
+++ b/src/haddock/libs/libontology.py
@@ -143,7 +143,7 @@ class ModuleIO:
         """Check how many of the output exists."""
         total = 0.0
         present = 0.0
-        for i, element in enumerate(self.output):
+        for element in self.output:
             if isinstance(element, dict):
                 total += len(element.values())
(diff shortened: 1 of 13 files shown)


amjjbonvin (approved):
A useful addition!  We should also adapt the current examples to make them complete. Probably means setting the tolerance in the example to 20 - need to check how many models fail

joaomcteixeira (approved):

PR #261, opened 2022-01-05
No limit of 20 molecules for `topoaa`

joaomcteixeira:
Useful when the next module is `emscoring`.

Activate it with the new parameter

```toml
[topoaa]
limit = false
```

@amjjbonvin


amjjbonvin (approved):
Ok - so setting `limit=false` should do the trick for scoring purposes (when giving a list instead of a single ensemble file)

joaomcteixeira:
It should work both ways. If you give an ensemble `topoaa` will split the ensemble into its models and work for each model separately. but you can also give a list of one molecule per model if you have the models in individual files.

PR #263, opened 2022-01-05
Update `libhpc`

rvhonorato:
With #219 we forgot to update this lib, we really need the tests. 😞

### DIFF
--- a/examples/docking-protein-protein/docking-protein-protein-hpc.cfg
+++ b/examples/docking-protein-protein/docking-protein-protein-hpc.cfg
@@ -2,7 +2,7 @@
 # Rigid-body docking example
 
 # directory in which the scoring will be done
-run_dir = "run1"
+run_dir = "run1-hpc"
 
 # ###
 mode = 'hpc'
--- a/src/haddock/libs/libhpc.py
+++ b/src/haddock/libs/libhpc.py
@@ -193,7 +193,7 @@ class HPCScheduler:
 
                     per = float(total_completed) / len(self.worker_list) * 100
 
-                    log.info(f">> {per:.0f}% done")
+                    log.info(f">> {per:.2f}% done")
                     if completed_count + failed_count == len(job_list):
                         completed = True
                         end = time.time()
--- a/src/haddock/libs/libhpc.py
+++ b/src/haddock/libs/libhpc.py
@@ -53,11 +53,11 @@ class HPCWorker:
         self.job_status = "unknown"
 
         self.moddir = Path(tasks[0].envvars['MODDIR'])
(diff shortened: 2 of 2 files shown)


joaomcteixeira:
Isn't for the `HPC` that we need the new parameter `self_contained = true` ? @rvhonorato try using that global parameter in the config file and try to solve the path problem without using the `resolve`. `MODDIR ` should point already to the correct place (or very close to).

rvhonorato:
Don't think so @joaomcteixeira, its the same error below with or without `self_contained = true`

```python
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/libs/libutil.py", line 307, in log_error_and_exit
    yield
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/clis/cli.py", line 148, in main
    workflow.run()
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/libs/libworkflow.py", line 30, in run
    step.execute()
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/libs/libworkflow.py", line 111, in execute
    module.run(**self.config)
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/modules/base_cns_module.py", line 47, in run
    self._run()
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/modules/topology/topoaa/__init__.py", line 171, in _run
    engine = Engine(jobs)
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/libs/libhpc.py", line 147, in __init__
    self.worker_list = [
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/libs/libhpc.py", line 148, in <listcomp>
    HPCWorker(t, j) for j, t in enumerate(job_list, start=1)
  File "/trinity/login/rodrigo/repos/haddock3/src/haddock/libs/libhpc.py", line 57, in __init__
    self.config_path = tasks[0].envvars['RUN']
KeyError: 'RUN'
```

This PR here removes exactly this `RUN` variable.

joaomcteixeira:
`RUN` has been depecrated. We don't use it anymore.

rvhonorato:
Yes, its why I am removing from `libhpc` 😄

joaomcteixeira:
with the `self_contained = true` you can run `cns < in > out` locally. So you should be able to do this for the HPC. Also, I made it such that also the `cns` executable is copied over the run dir.
Can you try o use always `self_contained = true` for the `HPC` mode and instead of manually writing the export variables to the job file, have the job file source the `envvars` created for each module? And use the `cns` script in the `run_dir`, that  is `../cns`.

rvhonorato:
Merging this based on our discussion on slack 👏🏽

PR #264, opened 2022-01-05
`tolerancy` should be `tolerance`

rvhonorato:
I could swear the correct spelling was `tolerancy`, turns out its `tolerance` 😅


joaomcteixeira (approved):
:coffee:

PR #265, opened 2022-01-05
Add missing parameters to `emscoring`

rvhonorato:
While running the CAPRI scoring we noticed some errors in `emscoring`, this was because of some missing parameters, with the changes proposed here it is working as expected.


src/haddock/modules/scoring/emscoring/cns/scoring.cns
```diff
@@ -55,6 +55,12 @@ evaluate ($Data.flags.vean = false)
 
 evaluate ($data.ncomponents=$ncomponents)
 
+evaluate ($nmol=1)
```
amjjbonvin, line 58: I don't see why this would be needed? Were there errors related to that? We read by default all param files - see `read_param.cns`

rvhonorato: At some point CNS is also looking for this:
```
     do (store6 = 1) (tag and segid $Toppar.prot_segid_$nchain1 
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

rvhonorato:
superseded by #270

PR #267, opened 2022-01-07
Implemented `JobInputFirst`

### LINKED ISSUE #245: Implement new class that can handle non `CNSJobs`
The part in the `caprieval/__init__.py` is okay. But I don't like the other parts in the `libsubprocess` and `Job`. We should have a different class for that. What the `scheduler` requests to the `Job` should always be the same attribute. These kinds of switches should be handled by the specific subclass. 

Also, `Job` should not have the `arg_first` argument. We should have a subclass for that. Since we are using OOP, let's make use of it :relaxed:

_Originally posted by @joaomcteixeira in https://github.com/haddocking/haddock3/pull/241#pullrequestreview-842639448_

joaomcteixeira:
Closes #245 

Creates a proper subclass. I decided to use `make_cmd` instead of using the `super()` function because `make_cmd` can serve a prestep for testing or any other functionality.

### DIFF
--- a/src/haddock/libs/libsubprocess.py
+++ b/src/haddock/libs/libsubprocess.py
@@ -8,36 +8,25 @@ from haddock.core.defaults import cns_exec as global_cns_exec
 from haddock.core.exceptions import CNSRunningError, JobRunningError
 
 
-class Job:
-    """A job to be executed by the engine."""
+class BaseJob:
+    """Base class for a subprocess job."""
 
-    def __init__(self, input, output, executable, *args, arg_first=False):
-        self.input = input
+    def __init__(self, input_, output, executable, *args):
+        self.input = input_
         self.output = output
         self.executable = executable
         self.args = args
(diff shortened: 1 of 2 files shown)


src/haddock/libs/libsubprocess.py
```diff
@@ -8,36 +8,25 @@
...
 
-class Job:
-    """A job to be executed by the engine."""
+class BaseJob:
+    """Base class for a subprocess job."""
 
-    def __init__(self, input, output, executable, *args, arg_first=False):
-        self.input = input
+    def __init__(self, input_, output, executable, *args):
+        self.input = input_
```
joaomcteixeira, line 15: here I changed to `input_` because `input` is a reserved python function, better not use it.

joaomcteixeira:
Clean, you can review now.

rvhonorato (approved):

PR #272, opened 2022-01-08
Implement `alascan` module

rvhonorato:
🚧  🚧  🚧  🚧  🚧


joaomcteixeira:
Very cool :dark_sunglasses: 

 I haven't seen in detail the PR yet, but think if it is really needed that `alascan` be a CNS module. Maybe it is best to have a module to create PDBs with `Ala` (that is a module that does the ala mutations) and then the `rigidbody` takes over. Think always in "one module one task" paradigm.

:rocket:

amjjbonvin:
For ala scanning it would be more a refinement scenario (e.g. flexref, mdref, emref and combinations there of)

rvhonorato:
> Very cool 🕶️
> 
> I haven't seen in detail the PR yet, but think if it is really needed that `alascan` be a CNS module. Maybe it is best to have a module to create PDBs with `Ala` (that is a module that does the ala mutations) and then the `rigidbody` takes over. Think always in "one module one task" paradigm.
> 
> 🚀

one module = `alascan` = one task = `alanine scanning` 😉

rvhonorato:
Closing this since it was just an experiment of possibilities, not really code to be used anywhere.

amjjbonvin:
I would implement it as an analysis module - it is nice and useful!

joaomcteixeira:
I thought was a good prototype also. Definitively the idea is very good. I restored the branch. There's no need to delete, if we want to recover that code in the future it is there in the branch.

PR #274, opened 2022-01-14
Convert `default.cfg` to YAML files

joaomcteixeira:
Convert modules' `default.cfg` to YAML files for better integration with other platforms (such as https://github.com/i-VRESSE/workflow-builder/pull/1) and as a documentation source for documentation generating scripts. This PR is the result of our bi-weekly meetings.

**These changes do NOT affect the HADDOCK3 toml-like user configuration file.**

The proposed `yaml` structure for modules' default config is as follows (**OUTDATED**):

```yaml
# this was the first version and was DISCARDED after several considerations - read further
basic:
  parameter_name:
     default: 1
     min: 0
     max: 9999
     type: integer
     hoover: mini doc sentence
     doc: extended information about the parameter
  parameter_name_2:
     default: some string
     max: 9999
     type: string
     hoover: mini doc sentence
     doc: extended information about the parameter
  # more parameters here
intermediate:
  # ... intermediate parameters here
guru:
  # ... guru parameters here
```

~~I think it is easier to maintain if the expert levels are subdictionaries in the yaml file (this is how it is implemented in this PR, but can be changed).~~ Nested dictionaries are also supported as is the case for the `topoaa` module.

Personally, I like the yaml structure and is giving very nice feeling and results while developing this PR.

The original `default.cfg` files were converted to YAML with [this script](https://gist.github.com/joaomcteixeira/d5ff5160de632565280e17a1508d36ef).

@sverhoeven what other keys per parameter do we need for IVRESSE, any additional requirements?

YAML pros against JSON:
1. accepts comments
2. accepts multiline strings easily

YAML cons against the current format:
1. it's harder to quickly inspect the parameters by eye because there are much more lines
  2. (edit) We solved this by creating the `haddock-cfg` CLI that translates `default.yml` to user readable haddock3 config files
2. I am not sure how sensitive is YAML to indentation.

**Implemented a `CLI` to retrieve the default configuration in the nice human-readable haddock3 format, for example:**

`haddock3-cfg -h`
`haddock3-cfg -m topoaa`

It will generate a haddock3 config with helper messages like the example below:

```
[module]
# basic
par1 = value # $min 0 $max 10 $hoover This is one sentence help text $doc This is a long doc to explain the variable correctly
par2 = value # $min 10 $max 100 $hoover This is one sentence help text $doc This is a long doc to explain the variable correctly

# intermediate
par3 = value
```

This generated config can be copy-pasted directly to the user's haddock3 config file.

**TODO:**
- [x] Once we remove the unnecessary parameters from the `defaults.cfg` (in another Pull Request) the new `defaults.yml` can be regenerated before merging this PR.
- [x] remove the old `defaults.cfg`

After this PR is accepted you need to perform the following commands from inside the `haddock3` folder and conda environment:
`conda env update --file requirements.yml --prune`
`python setup.py develop --no-deps`


joaomcteixeira:
Note, these **only** affect the defaults inside the core code. The user config will remain as is, simple and beautiful :stuck_out_tongue_winking_eye:

amjjbonvin:
But the user will not know how to write things since the default uses a different syntax…..

Asking for support troubles…

joaomcteixeira:
Thinking deeply about it, considering issues related to the architecture, maintainability, and user perspective, I think what is best is to prepare a `CLI` where the user can retrieve a default HADDOCK3 configuration with all the parameters and explanations. Or (maybe even better) we can have all the parameters nicely explained in the documentation web page by retrieving automatically all the information from the YAML files. Updating the YAML would automatically update the configuration documentation page. I think this is best.

amjjbonvin:
I suggest to first check carefully the current `cfg` files and remove anything not needed making sure that all examples still work. And only then convert to this yaml format. Will save us effort in editing the yaml... We need to process a lot of data and this requires manual editing to define the correct min/max values and all the relevant descriptions.

joaomcteixeira:
Sounds good. I have the script to regenerate the yamls, so we can leave this as a draft. I will prepare what's left, and when the parameters have been parsed, we recreate the yamls and merge this one. Also, I thought we can have the `basic`, `intermediate` and `guru` as the first layers of the YAML. Might be easier that way to move parameters up and down the layers and for us to quickly know which ones belong to which layer.

amjjbonvin:
Ok - I then need to find the time to clean and test the `defaults.cfg` files.

joaomcteixeira:
Added a `CLI` to extract the yaml configuration defaults to a haddock3 config file with the corresponding help message and ready to be copy+pasted to any haddock3 config file. If a proof of concept that the YAML is infact a good way to go because any API can extract all the info from there and there is a single place where we define everything related to the parameters.

Fetch this branch and run:

`python src/haddock/clis/cli_cfg.py -h` or try `python src/haddock/clis/cli_cfg.py -m topoaa`

You will get a file `haddock3_topoaa.cfg` in the `CWD`.

Once this PR is accepted, the `python setup.py develop --no-deps` will render the 'CLI` system-wide available. It is a good way for users to obtain info also.

joaomcteixeira:
Done. I have updated the first message of this PR, please read it.
This PR is ready to review and merge after the `defaults.cfg` files are clean from the unnecessary parameters.
I like the result of the default YAML idea.

src/haddock/modules/analysis/caprieval/defaults.yml
```diff
@@ -0,0 +1,101 @@
+basic:
+  reference_fname:
+    default: ''
+    type: path
+    length: 9999
+    hoover: No help yet
```
rvhonorato, line 6: do you mean `hover`?

src/haddock/modules/analysis/caprieval/defaults.yml
```diff
@@ -0,0 +1,101 @@
+basic:
```
rvhonorato, line 1: doesn't it make more sense if each parameter have a `level` instead of categorizing by it?

src/haddock/modules/analysis/caprieval/defaults.yml
joaomcteixeira: yes :relaxed:

src/haddock/modules/analysis/caprieval/defaults.yml
joaomcteixeira: Is it now easier to see all `basic` in a row? If you open the config file and want to see "which are all `intermediate` parameters", how do you do if params have a `level` key instead?

From the python parsing perspective is the same, there's no problem changing it.

rvhonorato: just a matter of preference I guess, for me would be better to have all parameters listed and each one would have a level definition, maybe less error-prone?

When someone is adding a new parameter i have a feeling they'll just add it in the same level as you define the levels now.

joaomcteixeira: It is possible also. We can always add `# basic` kind of comment to separate them visually. I am agnostic, to be honest. Either way is fine for me. I will wait for @amjjbonvin final verdict before changing the implementations. Alexandre will be the one changing these files the most.

src/haddock/modules/analysis/caprieval/defaults.yml
```diff
@@ -3,99 +3,99 @@ basic:
     default: ''
     type: path
     length: 9999
-    hoover: No help yet
+    hover: No help yet
```
rvhonorato, line 6: I think `tooltip` is better 😉

src/haddock/modules/topology/topoaa/defaults.yml
```diff
@@ -0,0 +1,3776 @@
...
+  tolerance:
+    default: 0
+    type: integer
+    min: -9999
+    max: 9999
+    hover: No help yet
+    doc: No help yet
+  input:
+    mol1:
+      prot_segid:
```
sverhoeven: input > mol1 or mol2 now looks like a dict of dicts while it is being parsed as an list of dicts. It would be nice to have a hint that it should be converted to a list.

Maybe something like
```suggestion
  input:
    type: list
    items:
      mol1:
        prot_segid:
```
or as mol1 and mol2 values are the same you could only define the type of items in the list once. This borrows from https://json-schema.org/understanding-json-schema/reference/array.html#items

```suggestion
  input:
    type: array
    index_prefix: mol
    items:
      prot_segid:
```

src/haddock/modules/analysis/caprieval/defaults.yml
sverhoeven: I see three ways
1. parameters nested within level
```yaml
basic:
 parameter1: ...
```

Each parameter belongs to 1 level and each greater level should include parameters of the smaller level.

2. levels per parameter
````
parameter1:
  ...
  levels: 
    - basic
````

It is harder to get overview which parameters are required for a level, but CLI, generated docs can help alleviate this.

3. levels outside parameters
```
parameters:
  parameter1: ...
levels:
  basic:
  - parameter1
```

The parameter keys need to be repeated

> A side note YAML can use [anchors and aliases](https://yaml.org/spec/1.2.2/#71-alias-nodes) to reduce duplication.

**Review by sverhoeven — COMMENTED** [review]
Regarding keys of a parameter
* The key `hoover` or `tooltip` sounds like it can only be used during a mouse over in a GUI. I would make it a bit more generic like `title` or `label` or `name`. So it can be used in generated docs or other places as well.
*  To perform validation when `type` is `path` it would be nice to add `accept` key for allowed [mimetypes of file extensions](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept)
* See inline comments in regarding nested dicts and levels
* A parameter can depend on others, for example `topoaa.input` and the global `molecules` parameter should have same number of items. It would be nice to capture that somehow. See [JSON schemas if-then-else](https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else) for inspiration.

**Comment by joaomcteixeira** [comment]
You are right about the `hover`. As you said, what about `title`, `short` (description), `long` (description)? Sound better. 

`input` is a dictionary of dictionaries, but indeed it can be made a dictionary of lists. Internally, at the python level, we use it as such. I will make that way in the YAMLs.

Regarding the expert level, each parameter can only be at one level: `basic`, `intermediate`, or `guru`. When we show the user the `intermediate` level, actually we will show both `basic` and `intermediate`. It adding up layers scheme, not selecting. 

Okay, we can add the `accept` key in the `Path` files.

**Comment by joaomcteixeira** [comment]
I've updated the `defaults.yml` files according to today's IVRESSE meeting (@sverhoeven). Also, I have updated tests and the new `haddock3-cfg` CLI. The users (you) can use this CLI to extract the default values of a module (YML) to a nice haddock3 config reader. For example:

```
# move to this branch
python setup.py develop --no-deps  # you need to install the new CLI
haddock3-cfg -m topoaa
haddock3-cfg -m rigidbody
```

Regarding the `yml` files themselves, this is an example of a structure (see below). All parameters have mandatory keys: `default`, `type`, `explevel`, and `title`. Other keys are optional depending on the parameter type. The `group` key addresses @amjjbonvin request that parameters should belong to groups that are displayed all together, for example. The script I used to convert from `cfg` to `yml` is [here](https://gist.github.com/joaomcteixeira/d5ff5160de632565280e17a1508d36ef), if you wish to try it, run it from within the `src/haddock/` folder; you need to have haddock3 installed.

These last commits already integrated the changes proposed in #300. In case the latter is not accepted I will revert here as well.

Let me know your thoughts. Have a great weekend!

```yml
tolerance:
  default: 0
  type: integer
  min: -9999
  max: 9999
  precision: 3
  title: No title yet
  short: No short description yet
  long: No long description yet
  group: ''
  explevel: basic
log_level:
  default: verbose
  type: string
  maxchars: 100
  minchars: 0
  title: No title yet
  short: No short description yet
  long: No long description yet
  group: ''
  explevel: basic
ambig_fname:
  default: ''
  type: file
  title: No title yet
  short: No short description yet
  long: No long description yet
  group: ''
  explevel: basic
crossdock:
  default: true
  type: boolean
  title: No title yet
  short: No short description yet
  long: No long description yet
  group: ''
  explevel: basic
```

rvhonorato (approved):

amjjbonvin (approved):
So now we should start adding info to those files.
I guess you used default values. For example for the tolerance you have now as min/max -9999/+9999 while this is a percentage (i.e. between 0 and 100).

joaomcteixeira:
Yes. For now, I just copied the `default` values. The other fields are not defined, I simply added some numbers to fill them out.

joaomcteixeira:
I see you accepted these changes, so I will update the rest of the code accordingly before merging this PR. If you wish to add any details to the configs do it meanwhile in this branch as well.

amjjbonvin:
For the `haddock3-cfg` cli, I would suggest to write the output to standard output instead of directly to a file. A user can always redirect the output. You might only want to check a parameter and not write a file every time

joaomcteixeira:
Made some updates: `git pull origin yamlcfg`, they are as follows:

* @amjjbonvin that's a good suggestion for the CLI. I have implemented it.
* Updated code also after #300 
* @amjjbonvin YAML accepts multiline strings. I have updated your `topoaa/defaults.yml` config file. [See the differences here](https://github.com/haddocking/haddock3/pull/274/commits/16061d91c4a35ae304435ecbd1bb71c17e72fa69). If you use colon `:` you need to have the strings in quotes also.
* If you run `tox -e py39 -- tests/test_modules_general.py` unittest will try to read all the `yml` configs and spot any errors. It might be tricky at first the `tox` output but at least you'll see if there are some red lines appearing.

joaomcteixeira:
@amjjbonvin this PR is ready to be merged. You can continue adding documentation to the `defaults.yml` files here or we can merge and continue adding documentation later in other PRs. I think it's best we merge. Let me know.

src/haddock/modules/analysis/caprieval/defaults.yml
joaomcteixeira: We have solved thid using the `explevel` key for each parameter.

src/haddock/modules/analysis/caprieval/defaults.yml
joaomcteixeira: Solved with `title`, `short` and `long`.

src/haddock/modules/analysis/caprieval/defaults.yml
joaomcteixeira: We have solved this adding an `explevel` to each parameter.

amjjbonvin:
> -    hoover: No help yet
> +    hover: No help yet
Since you expect the text to be between quotes should this also be the case when there is “No help yet” ?

amjjbonvin (approved):
Best to merge indeed and then keep adding description to the yams files.

joaomcteixeira:
> Since you expect the text to be between quotes should this also be the case when there is “No help yet” ?

No. You only need quotes when your strings have *special*  characters like colon `:`. We have `unittest` to ensure all `yaml` configs are parseable. When you are editing a YAML config you can then run `tox -e py39 -- tests/test_modules_general.py` to see if any errors arise. You need to be in the `haddock3` conda environment for this to work and in the main `haddock3` repository folder.

@sverhoeven we are merging :wink:

PR #277, opened 2022-01-16
Restrict characters accepted in strings and paths

### LINKED ISSUE #273: Check for valid characters in parameters that are strings.
```
/ not empty, check
        if (!runname.value.match(/^[0-9a-z_-]+$/)) {
            runnameError.textContent = "Only characters (A-Z), numbers (0-1) and \"_\" or \"-\" allowed.";
```
Add character check to haddock3 python shell level.

Thanks @rvhonorato for pointing this out.

joaomcteixeira:
Closes #273 and implements additional improvements:

1. The config reader now restricts the accepted characters in parameter values to word (`\w`) and some more depending on the cases. Meaning, strange (potentially malicious) characters will be automatically rejected when reading the haddock3 user configuration file.
2. Parameters that point to paths are also read as paths. These are parameters with suffix `_fname`, or named `executable`, `molecules`, and `cns_exec`. When implementing other parameters that are also paths, those should be manually added to the config reader otherwise they will be considered strings and strings don't accept path-like chars.
3. Paths can be defined without the preceding `./`.

Overall, this PR ensures no strange characters are passed to the haddock3 config which also prevents typo errors by the users.

Many new tests were added.


src/haddock/gear/config_reader.py
```diff
@@ -40,17 +40,34 @@
...
 
-# https://regex101.com/r/q2fuFl/1
-_string_re = re.compile(r'''^ *(\w+) *= *("(.*?)"|'(.*?)')''')
+# string regular expression here matches the expectations of modules' defaults
+# there are no modules defaults that expect any other character. So, if we
+# block it's read from here, avoid also downstream errors.
+# https://regex101.com/r/TRMx0B/1
+_string_re = re.compile(
+    r'''^ *(\w+) *= *'''
+    r'''("([a-zA-Z0-9\.]*?)"|'([a-zA-Z0-9\.]*?)')'''
```
amjjbonvin, line 49: Why not allowing for example `-` in a string?   Could be in a filename for example.

joaomcteixeira: So, that's the discussion I want to have. `-` is accepted on Paths but not strings because there are not the `-` character in string module's parameters. If we implement that only used chars can be input we correct for both rough bad input and malicious characters.

joaomcteixeira:
@amjjbonvin do you want to enforce paths to have the `./` prefix? If we do it becomes easier to define types as we read configuration files. I think there are no other drawbacks apart from having to write the `./` (current path), for example `run_dir = "./run1"`.

amjjbonvin:
No preferences - don’t we have fname in the variable of all input files? That should be enough to recognize the type

joaomcteixeira:
Yeah... I was thinking about `run_dir` or any other future parameter that could take a folder... but may become too clumsy do have to add the `./`  all the time.

joaomcteixeira:
Actually, you gave me an additional idea.

joaomcteixeira:
Updated the original message, read to review and merge.

amjjbonvin (approved):
I am not an expert on regex - I will trust you on this one or leave it to @rvhonorato to check carefully things

src/haddock/gear/config_reader.py
```diff
@@ -40,32 +46,117 @@
...
+    "cns_exec",
+    "executable",
+    "molecules",
+    RUNDIR,
+    ]
+
+_keys_files_regex = "|".join(_keys_that_accept_files)
 
-class _Path_re:
+_file_linux_re = re.compile(
```
rvhonorato, line 99: this won't take `par_fname = "../data/../other_folder/data/file.pdb"`, quite an odd scenario but people are very creative and in my opinion this should also be valid

src/haddock/gear/config_reader.py
```diff
@@ -40,32 +46,117 @@
...
+    r'''^ *(\w+_fname'''
+    f'''|{_keys_files_regex}) *= *'''
+    r'''("((\.{1,2}\/)*[\w\-\/]+[\w\-]\.?[\w\-]+)"'''
+    r'''|'((\.{1,2}\/)*[\w\-\/]+[\w\-]\.?[\w\-]+)')'''
+    )
+
+_file_windows_re = re.compile(
```
rvhonorato, line 106: a bit unclear to me, will it parse things **generated** with windows or **to be executed in windows**? does this mean we are adding windows support?

rvhonorato (approved):
Some minor clarifications but the PR looks ok to me.

But I have a wider range comment:
I understand why you decided to create this custom parser and format and everything that comes to it, i think its a great, very expertly crafted implementation but I'm quite concerned about the maintainability and long term survival of this custom file. 

There are at least 14 very complex regexes here, you needed to code every single behaviour, very little of this is clear to me that am actively working in the project. Imagine if you and me are no longer involved in the development and the next person needs to figure out why, based on:

```python
_string_re = re.compile(
    r'''^ *(\w+) *= *'''
    r'''("([\w\-]*?)"|'([\w\-]*?)')'''
    )
```

`value = "stringvalue"` is considered a string
but
`value = "string_value"` is not considered a string
and also
`value = "stringvalue.pdb"` is not considered a string

You probably have a very "obvious" answer to this..

imho the gain of using this custom parser - being able to repeat modules does not outweigh the maintainability and suitability of using a "normal" format.

But what is done is done, let's work hard on the documentation to minimize this. 👍🏽

joaomcteixeira:
Step by step,

> value = "stringvalue" is considered a string

Yes, it reads as a string.

> value = "string_value" is not considered a string

Wrong. It is considered a string as well because `_` is a word character.

> value = "stringvalue.pdb" is not considered a string

True, `value` would be read as a `Path` if `value` is defined in the `default.cfg` as a path. If `value` is not defined as a Path in the `defaults.cfg` the configuration reader will give an error. Why would we allow the user to insert a path value in a parameter that is not supposed to be a path?

> this won't take par_fname = "../data/../other_folder/data/file.pdb"

True, it won't take it. I thought about this. These paths are madness. Please me clean :-) But I can allow users to add them if @amjjbonvin agrees. Note that `../data/folder/data/file.pdb" is accepted.

> a bit unclear to me, will it parse things generated with windows or to be executed in windows? does this mean we are adding windows support?

No, we are not supporting Windows. But it was so straightforward to have it that I made it so just in case a user wants to path a windows path on a Linux machine she/he gets a warning about that. If we ever support windows, it will be there already.

Regarding maintainability. I think the file is super maintainable. There are very clear regexes for each type (`str`, `path`, `numbers`, etc). You can see from the previous PRs that adding additional features to the file did **not** require refactoring and only additions to the file. Also, the regexes automatically avoid users injecting unwanted characters in the parameter values. I know regular expressions are strange at first but they become very clear as you use them. With this setup, we avoid having to validate many things later on down the line.

I think the code is absolutely self-explanatory, but I will add more comments to the file. Maintainability comes from well thought and defined design. Keeping things nested and "simple" doesn't mean things are maintainable. Honestly, for me would be difficult to rewrite this file as if I had 1-year python experience and I was starting with the basics. Surely someone with higher experience than me would rewrite this file in a much better way, maybe with a single regex to parse the whole file ;-)

If we revert to using `toml` there are many functionalities we will lose.

amjjbonvin:
All good on my side

rvhonorato:
> > value = "string_value" is not considered a string
> 
> Wrong. It is considered a string as well because `_` is a word character.

Screenshot 2022-01-27 at 11 10 06

joaomcteixeira:
Sorry, @rvhonorato that link was outdated.

I have updated all the regex101 links and made them permanent. Also improved the regexes for `ASCII` chars only. The config reader alone has 174 unit tests. Added lots of comments explaining the whole implementation.

I think it is cool.

rvhonorato (approved):

PR #280, opened 2022-01-18
Clean default parameters, refactor CNS scripts and tweak examples

joaomcteixeira:
@amjjbonvin here's the branch to clean the `defaults.cfg` file plus attempting to implement what we discussed about types. It goes (should)  as follow:

- numbers should be numbers, not strings
- parameters that are numbers that you DONT want to write in the CNS should be `param = nan`, without quotes.
- parameters that are strings that you don't want to write to the CNS should be empty strings `param = ""`
- Remove every `None` you find around
- For now, paths that are not to be written to CNS should be empty strings as well.

Give it a try, let me know the feeling.


examples/docking-protein-protein/docking-protein-protein.cfg
```diff
@@ -4,7 +4,7 @@
 # directory in which the scoring will be done
 run_dir = "run1"
 ncores = 20
-# cns_exec = "path/to/bin/cns" # optional
+cns_exec = "/Users/abonvin/haddock3/bin/cns"
```
joaomcteixeira, line 7: @amjjbonvin  Care adding these paths. Use `cns_exec = ""` will read the `bin` installation by default.

examples/docking-protein-protein/docking-protein-protein.cfg
```diff
@@ -31,29 +31,29 @@ nhise = 1
 hise_1 = 15
 
 [rigidbody]
-ambig_fname = 'data/e2a-hpr_air.tbl'
+ambig_fname = "data/e2a-hpr_air.tbl"
```
joaomcteixeira, line 34: Now is the perfect time to enforce `./` as in `./data` for paths if you wish. Is an idea, not a necessity,

amjjbonvin:
Sorry - should comment it out again - was testing before if this was solving the CNS path problem

Remember - work in progress… :-)

joaomcteixeira:
Can't understand why that error. `cns_exec` is reaching subprocess as an empty string. I can't reprocue that here. Can you add a `print(cns_exec)` [here in this file](https://github.com/haddocking/haddock3/blob/1555fd00e3fd4f70ccbb406b7b0ed9d6aeeca1f1/src/haddock/core/defaults.py#L16) and run `python defaults.py` and tell me what it prints?

joaomcteixeira:
Hi @amjjbonvin 
I made a prototype code to identify the parameter blocks from the `deafults.cfg`. The code might not be complete or a bit buggy in some edge cases but may allow you to continue cleaning the `defaults.cfg` until I am back to the office (Monday). All cases I could think about for now are working.

An expandable parameter block is a group of parameters that share `first` and `_1`, for example the `c3sym` block. Also, the keys for each block are identified, for example `sta1`, `end1` etc, meaning if a user writes `c3sym_lala_1` results in a controlled error.

The code thus confirms the new block (for example `c3sym_*_2`) has the expected number of parameters.

There are some parameters that might break this rule, for example the `start_rair_*`. If these are blocks, could you rename those to `rair_1_start_1`, `rair_1_end_1` for example so that the block is defined by `rair_1` and not by `start`? Otherwise, how should we read these blocks?

Try it out, and let me know your feelings.

amjjbonvin:
Should be doable and makes sense. Will take into account when further cleaning the default.cfg files

> There are some parameters that might break this rule, for example the start_rair_*. If these are blocks, could you rename those to rair_1_start_1, rair_1_end_1 for example so that the block is defined by rair_1 and not by start? Otherwise, how should we read these blocks?
>

amjjbonvin:
Check the slack discussion - the expandable parameter implementation / check should allow for more flexibility following simply the following rule:

<unique key><anything>_1

And anything could even be empty.

examples/docking-protein-DNA/docking-protein-DNA-mdref.cfg
```diff
@@ -22,34 +22,36 @@ molecules =  [
 autohis = true
 
 [rigidbody]
-ambig_fname = 'data/cro_air.tbl'
+ambig_fname = "data/cro_air.tbl"
```
joaomcteixeira, line 25: FYI, haddock3 config reader accepts both single and double quotes

amjjbonvin (approved):
I have been working a lot on the CNS parts and the defaults.cfg files. Everything working fine now. Time to merge it!

But @rvhonorato should review it as well.

src/haddock/gear/config_reader.py
```diff
@@ -49,6 +49,8 @@
 # https://regex101.com/r/K6yXbe/1
 _none_re = re.compile(r'^ *(\w+) *= *([Nn]one|[Nn]ull)')
 
+_nan_re = re.compile(r'^ *(\w+) *= *(nan)')
```
rvhonorato, line 52: maybe also good to capture `NaN`?

src/haddock/modules/refinement/emref/cns/emref.cns
```diff
@@ -411,9 +408,10 @@ end if
 noe
     averaging  * sum
     potential  * soft
+    scale      * 1.0
     scale ambi $amb_scale 
-    scale dist $amb_scale
-    scale hbon $amb_scale
+    scale dist $unamb_scale
+    scale hbon $hbond_scale
```
rvhonorato, line 414: was this a hidden feature? 😉

src/haddock/modules/refinement/emref/cns/symmultimer.cns
```diff
@@ -21,11 +21,11 @@ if ($Data.flags.sym eq true) then
     while ($ncount < $data.numc2sym) loop c2symloop
 
         eval ($ncount = $ncount + 1)
-        evaluate ($i1start = decode($Toppar.c2sym_sta1_$ncount))
-        evaluate ($i1end   = decode($Toppar.c2sym_end1_$ncount))
+        evaluate ($i1start = $Toppar.c2sym_sta1_$ncount)
+        evaluate ($i1end   = $Toppar.c2sym_end1_$ncount)
         evaluate ($chain1  = $Toppar.c2sym_seg1_$ncount)
-        evaluate ($i2start = decode($Toppar.c2sym_sta2_$ncount))
-        evaluate ($i2end   = decode($Toppar.c2sym_end2_$ncount))
```
rvhonorato, line 28: what was the purpose of this `decode`?

src/haddock/modules/refinement/flexref/cns/flexref.cns
```diff
@@ -21,31 +21,29 @@ end if
...
-evaluate ($saprotocol.inter_fin_cool2=$fin_cool2)
-evaluate ($saprotocol.inter_init_cool3=$init_cool3)
-evaluate ($saprotocol.inter_fin_cool3=$fin_cool3)
+evaluate ($saprotocol.tadhigh_t=$temp_high)
+evaluate ($saprotocol.t1_init=$temp_cool1_init)
+evaluate ($saprotocol.t2_init=$temp_cool2_init)
+evaluate ($saprotocol.t3_init=$temp_cool3_init)
+evaluate ($saprotocol.t1_final=$temp_cool1_final)
+evaluate ($saprotocol.t2_final=$temp_cool2_final)
+evaluate ($saprotocol.t3_final=$temp_cool3_final)
```
rvhonorato, line 30: The name of the parameters are more intuitive like this, we need to remember when writing the documentation that these are different from 2.x

src/haddock/modules/refinement/flexref/defaults.cfg
```diff
@@ -1,200 +1,305 @@
...
+dihedrals_on = false
+dihedrals_hot = 5
+dihedrals_cool1 = 5
+dihedrals_cool2 = 50
+dihedrals_cool3 = 200
+
+
+### Automatic dihedral angle restraining
+# Options are: "none" "alpha" "alphabeta" "all"
+ssdihed = "none"
```
rvhonorato, line 184: Does this need to be NONE for CNS? else would be less prone to confusion if it was `""`

src/haddock/modules/sampling/rigidbody/cns/rigidbody.cns
```diff
@@ -23,7 +23,6 @@ end if
 
 evaluate ($saprotocol.crossdock=$crossdock)
 evaluate ($saprotocol.randorien=$randorien)
-evaluate ($saprotocol.rigidmini=$rigidmini)
```
rvhonorato, line 26: This has no use anymore, right?

src/haddock/modules/scoring/emscoring/cns/calc_free-ene.cns
```diff
@@ -13,12 +13,10 @@
 @MODULE:scale_intra_only.cns
 
 evaluate ($eintfree = 0.0)
-@MODULE:flex_segment_back.cns
```
rvhonorato, line 16: Just to confirm if this was not removed by mistake

src/haddock/modules/scoring/emscoring/cns/scale_inter_final.cns
```diff
@@ -20,7 +20,7 @@ igroup
         eval($nchain2 = $nchain1 )
         while ($nchain2 < $data.ncomponents) loop nloop2
             eval($nchain2 = $nchain2 + 1)
-            eval($scalfac = $kinter * $Toppar.int_$nchain1_$nchain2)
+            eval($scalfac = 1.0)
```
rvhonorato, line 23: won't this impact the calculation?

src/haddock/modules/scoring/emscoring/cns/scoring.cns
```diff
@@ -145,7 +156,6 @@ do (z = refz) (all)
 {* ===================== calculate complex internal energy *}
 
 evaluate ($kinter = 1.0)
-energy end
```
rvhonorato, line 148: Isn't this needed to finish the calculation?

tests/test_libcns.py
```diff
@@ -32,7 +33,7 @@ def test_empty_vars_error(value):
         3453.543,
         'str',
         Path('path'),
-        '',
+        EmptyPath(),  # mepty paths needs to be written as ""
```
rvhonorato, line 36: *typo here

rvhonorato (commented):
A big one! Have some small comments, just to confirm a few things.

examples/docking-protein-DNA/docking-protein-DNA-mdref.cfg
amjjbonvin: For consistency I switched everywhere to double quotes

src/haddock/gear/config_reader.py
amjjbonvin: Why not indeed

src/haddock/modules/refinement/emref/cns/emref.cns
amjjbonvin: Yes it was...

src/haddock/modules/refinement/emref/cns/symmultimer.cns
amjjbonvin: Before those variables were define with double quotes making them strings. CNS has a decode/encode function to convert strings to numbers and vice-versa

src/haddock/modules/refinement/flexref/cns/flexref.cns
amjjbonvin: Will have to document all variables in the yaml file once we convert the cfg to yaml

src/haddock/modules/refinement/flexref/defaults.cfg
amjjbonvin: none does has a meaning... Wouldn't change the behaviour since we only actually check in the code if that variable is equal to "alpha" "alphabeta" or "all"

src/haddock/modules/sampling/rigidbody/cns/rigidbody.cns
amjjbonvin: `ridigmini` has indeed no use anymore. You simply don't call that module if you want to skip it.  One example of this is the `refine-complex` example.

But what if we want to dock during the flexible refinement stage (we did it in the past for some special cases)?  We might need in that case a `randorient` module after `topoaa`

src/haddock/modules/scoring/emscoring/cns/scale_inter_final.cns
amjjbonvin: Nope - we only calculate the intermolecular interactions and the weight for scoring should simply be 1. I removed the interaction matrix from the `default.cfg` file to simplify things.

src/haddock/modules/scoring/emscoring/cns/scoring.cns
amjjbonvin: No - already included in `calc_free-ene.cns` - was redundant.

src/haddock/modules/sampling/rigidbody/cns/rigidbody.cns
rvhonorato: allright! seems quite specific, but now more then ever we have the "flexibility" to do it! 😄

rvhonorato (approved):

joaomcteixeira:
@amjjbonvin what do to with the `parameter_1` case? Should this be a group also? I imagine not because you accepted the PR. But just want to make sure.

I am going to write the documentation and some tests, and after that I will merge.

src/haddock/gear/config_reader.py
joaomcteixeira: Done

amjjbonvin:
Not needed indeed (for now)

PR #291, opened 2022-01-27
Added full sampling cfg files

amjjbonvin:
Moved the test config files to `-test.cfg`
Changed the python scripts to run those (test and full versions)
Settings of the full runs should be similar to the 2.4 examples


rvhonorato (approved):

joaomcteixeira (approved):
Nice! Tested also the examples to make sure `mode=hpc` exists with `1` if there's an error so that the `-b` captures it.

PR #297, opened 2022-01-28
Set tolerance of modules in the workflow integration tests to 20%

### LINKED ISSUE #296: Set the tolerance to 20% for the workflow integration tests
For testing purposes set the tolerance to 20%

_Originally posted by @amjjbonvin in https://github.com/haddocking/haddock3/issues/295#issuecomment-1024057662_

rvhonorato:
closes #296

### DIFF
--- a/examples/run_tests.py
+++ b/examples/run_tests.py
@@ -41,17 +41,17 @@ except Exception:
 # edit this dictionary to add or remove examples.
 # keys are the examples folder, and values are the configuration files
 examples = (
-    ("docking-protein-DNA"         , "docking-protein-DNA-test.cfg"),
-    ("docking-protein-DNA"         , "docking-protein-DNA-mdref-test.cfg"),
-    ("docking-protein-homotrimer"  , "docking-protein-homotrimer-test.cfg"),
+    ("docking-protein-DNA", "docking-protein-DNA-test.cfg"),
+    ("docking-protein-DNA", "docking-protein-DNA-mdref-test.cfg"),
+    ("docking-protein-homotrimer", "docking-protein-homotrimer-test.cfg"),
     ("docking-protein-ligand-shape", "docking-protein-ligand-shape-test.cfg"),
-    ("docking-protein-ligand"      , "docking-protein-ligand-test.cfg"),
-    ("docking-protein-peptide"     , "docking-protein-peptide-test.cfg"),
-    ("docking-protein-protein"     , "docking-protein-protein-test.cfg"),
-    ("docking-protein-protein"     , "docking-protein-protein-cltsel-test.cfg"),
-    ("docking-protein-protein"     , "docking-protein-protein-mdref-test.cfg"),
-    ("refine-complex"              , "refine-complex-test.cfg"),
-    ("scoring"                     , "scoring-test.cfg"),
+    ("docking-protein-ligand", "docking-protein-ligand-test.cfg"),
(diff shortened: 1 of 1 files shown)


rvhonorato:
here I've changed it in `run_test.py`, so its only a single point of change, should this be changed in the .cfg instead?

amjjbonvin:
I guess it’s fine - but if present in the cfg file, which one will prevail?

amjjbonvin (approved):

rvhonorato:
> I guess it’s fine - but if present in the cfg file, which one will prevail?

The one in `run_test.py` will prevail

examples/run_tests.py
```diff
@@ -88,6 +88,10 @@ def main(examples, break_on_errors=True):
         with working_directory(folder):
 
             params = read_config(file_)
+            # https://github.com/haddocking/haddock3/pull/295#issuecomment-1024057662
+            for p in params:
+                if isinstance(params[p], dict) and p != "topoaa":
+                    params[p]['tolerance'] = 20
```
joaomcteixeira, line 94: This does not apply tolerance `20%` to the runs. You are changing the `params` dict within the python session but that is **not** passed to the configuration file. You would need to resave the config file again after the change. 

This script only runs `haddock3 <config>` for each of the examples listed.  If you want to change some parameter you need to change it **in the configuration file itself**. There are configuration files for the `tests`; I don't see the need to add a parameter modification in the `run_tests.py`. Also, if you add parameters in the `run_tests.py` overwriting the configuration files then we lose control of the configuration for tests. I disagree with changing the parameter in the `run*.py` files. These files should be just engines to run the configs one after the other, nothing more.

I strongly suggest adding the tolerance **in** the configuration files themselves.

examples/run_tests.py
```diff
@@ -41,9 +41,9 @@
 # edit this dictionary to add or remove examples.
 # keys are the examples folder, and values are the configuration files
 examples = (
-    ("docking-protein-DNA"         , "docking-protein-DNA-test.cfg"),
-    ("docking-protein-DNA"         , "docking-protein-DNA-mdref-test.cfg"),
-    ("docking-protein-homotrimer"  , "docking-protein-homotrimer-test.cfg"),
+    ("docking-protein-DNA", "docking-protein-DNA-test.cfg"),
```
joaomcteixeira, line 44: This file is outside the `tox.ini` file. I added those spaces on purpose, don't remove them :wink:

joaomcteixeira (changes requested):
I don't like this PR. I really think these changes are not appropriate. I would close it, and add the tolerance to the config files themselves.

rvhonorato:
> I strongly suggest adding the tolerance in the configuration files themselves.

Closing this and will re-implement according to the feedback

examples/run_tests.py
rvhonorato: Please indicate the stuff that you added for vanity so I know what I can change or not 😉

joaomcteixeira: every PR I do and merge to the `main` is done on purpose. Nothing is left unnoticed :sunglasses: tox controlled lint files are defined here -> https://github.com/haddocking/haddock3/blob/593fa7d3f63a09d826056ed4b8e54e1955d0a35c/tox.ini#L66

rvhonorato: sure, made another PR adding the documentation not to remove them, are these also present on somewhere else? if so add them in https://github.com/haddocking/haddock3/pull/299

PR #298, opened 2022-01-28
[REDO] Set tolerance of modules in the workflow integration tests to 20% 

rvhonorato:
This PR is a redo of #297, this time adding `tolerance = 20` to all the examples that contain `-test.cfg` and keeping the trailing white spaces


joaomcteixeira (approved):
looks good to me.

PR #299, opened 2022-01-28
Add documentation about whitespaces in `run_test.py`

rvhonorato:
As pointed in https://github.com/haddocking/haddock3/pull/297#discussion_r794655989, these whitespaces must be kept but this information is not present in the document, this PR adds a minor comment about this so they are not removed by mistake.


joaomcteixeira:
Addressed it :+1:

PR #308, opened 2022-02-07
Added description to topoaa default.yaml file

amjjbonvin:
Removed unneeded parameters


rvhonorato (approved):

joaomcteixeira:
@amjjbonvin in the `yaml` files the `nan` value should be indeed input as `.nan`; otherwise, a `"nan"` string will be generated. I have re-added the dots in the `topoaa` yaml config and added tests.

For the `prot_segid` you added a max of four chars. Are you counting also with CIF files somehow? `segid` should accept only one char because `chainid` is also only one char in PDB format v3. Forcing `segid` to a single char will allow than to use `chainxseg` from `pdbtools` and swap that information as needed. Let me know.

amjjbonvin:
In CNS it is the segID which is relevant and it is officially four characters.

Of course it can not be copied to the chainID, only the first character would be passed.

> For the prot_segid you added a max of four chars. Are you counting also with CIF files somehow? segid should accept only one char because chainid is also only one char in PDB format v3. Forcing segid to a single char will allow than to use chainxseg from pdbtools and swap that information as needed. Let me know.

PR #311, opened 2022-02-09
Update deps versions - forcing exact numbers

joaomcteixeira:
Update dependency versions, forcing exact numbers. This is important because for now, HADDOCK3 is *under development* but also in the future HADDOCK3 is mostly to be used as a program instead of a library, so enforcing exact dependency numbers avoids possible discrepancies between installations.

Because we have little dependencies, we can, once in a while, review the dependencies updates manually. I am looking for stability instead of cutting-edge.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
-pdb-tools>=2.4.3
-jsonpickle==2.0.0
-numpy==1.20.2
-tox==3
-pyyaml==6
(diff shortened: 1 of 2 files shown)


rvhonorato (approved):

joaomcteixeira:
You may need to recreate the haddock3 environment to make sure these changes apply. Mostly the one referring to `pdb-tools`.

PR #314, opened 2022-02-10
Updated install instructions

### LINKED ISSUE #118: The `--restart` option should also evalute the state of the module
@amjjbonvin requested:

The `--restart` option after #88 restarts the run at a given stage. For example, `--restart 2` would maintain modules `00` and `01` and restart the run at module `02_`, meaning it would delete everything previously done in `02_`. That raises the problem that if `02_` crashes in the last calculation of 10,000 calculations we need to repeat all that again. Hence, `--restart` should **also** evaluate the state of the stage to avoid repeating calculations.

joaomcteixeira:
Closes #118

* updates the install instructions
* adds minimal usage instructions
* Keeps the need for `--recursive`
* do not consider updating `fcc` because that package is stable and not expected to change. Once we fully move to `fcc2` we will re-update the install instructions.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -2,6 +2,8 @@
 
 ## 1.1 Clone this repository:
 
+Mind the `--recursive` flag!
+
 ```bash
 git clone --recursive https://github.com/haddocking/haddock3.git
 cd haddock3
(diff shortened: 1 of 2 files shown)


rvhonorato (approved):

amjjbonvin (commented):
In the usage, may-be good to also add the cli command to view the list of parameters for the various modules.

PR #318, opened 2022-02-11
Add YAML for mandatory parameters

### LINKED ISSUE #304: Add description for `run_dir` and `molecules`
After #274 and https://github.com/i-VRESSE/workflow-builder/pull/22

> Main problem is, that I have to define molecules and run_dir parameters myself. I would like to fetch the type/title/description from haddock3. @sverhoeven

I will add those. It is not straightforward that those parameters should be in the `modules/defaults.yaml` file or in another one. Because these are parameters without defaults and mandatory. Well, `run_dir` could have a default, but not `molecules`.

joaomcteixeira:
Closes #304 

Add a `yaml` with the description of mandatory parameters.

@sverhoeven can you comment to see if it's okay?

### DIFF
--- a/src/haddock/core/mandatory.yaml
+++ b/src/haddock/core/mandatory.yaml
@@ -3,11 +3,11 @@ run_dir:
   title: Run directory
   short: Folder to store the HADDOCK3 run
   long: The new folder that will be created to save the HADDOCK3 run
-  group: mandatory
   explevel: easy
 molecules:
   type: list
-  itemtype: ".pdb"
+  accept:
+    - ".pdb"
   minitems: 1
   maxitems: 20
   title: Input Molecules
@@ -15,5 +15,4 @@ molecules:
   long:
     Molecules must be PDB files. These PDB files can be single molecules
     or ensembles.
-  group: mandatory
   explevel: easy
--- a/src/haddock/core/mandatory.yaml
+++ b/src/haddock/core/mandatory.yaml
@@ -13,6 +13,6 @@ molecules:
   title: Input Molecules
   short: The input molecules that will be used for docking.
   long:
-    Molecules must be PDB files. These PDB files can be single molecules
-    or ensembles.
+    Molecules must be provided in PDB format. These PDB files can be
+    single molecules or ensembles using the MODEL/ENDMDL statements.
   explevel: easy
--- a/src/haddock/core/mandatory.yaml
+++ b/src/haddock/core/mandatory.yaml
@@ -1,0 +1,13 @@
+run_dir:
+  type: directory
+  title: Run folder
+  short: Folder to store the HADDOCK3 run
+  long: The new folder that will be created to save the HADDOCK3 run
+  group: mandatory
+molecules:
+  type: list
+  title: Input molecules
+  short: The input molecules that will be used for docking.
+  long:
+    Molecules must be PDB files. These PDB files can be single molecules
+    or ensembles.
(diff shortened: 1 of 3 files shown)


src/haddock/core/mandatory.yaml
```diff
@@ -0,0 +1,19 @@
+run_dir:
+  type: dir
+  title: Run directory
+  short: Folder to store the HADDOCK3 run
+  long: The new folder that will be created to save the HADDOCK3 run
+  group: mandatory
```
sverhoeven, line 6: This seem double. 
Also by specifying a group the builder will move the parameter into a collapsible group and not as the first parameter.

run_dir and molecules do not have a default value that makes them required for the builder already.

Can the group be removed?
```suggestion
```

sverhoeven (changes requested):
The Python code looks good to me, just a small nitpick.

I have some issues with the yaml file, see inline suggestions.

src/haddock/core/mandatory.yaml
joaomcteixeira: Yes, I can remove the group.

src/haddock/core/mandatory.yaml
```diff
@@ -0,0 +1,19 @@
...
+  explevel: easy
+molecules:
+  type: list
+  itemtype: ".pdb"
```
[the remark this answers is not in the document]
joaomcteixeira: I can make accept a list already

sverhoeven (approved):

joaomcteixeira:
@amjjbonvin if you agree also, you can merge it directly :+1:

src/haddock/core/mandatory.yaml
```diff
@@ -0,0 +1,18 @@
...
+  title: Run directory
+  short: Folder to store the HADDOCK3 run
+  long: The new folder that will be created to save the HADDOCK3 run
+  explevel: easy
+molecules:
+  type: list
+  accept:
+    - ".pdb"
+  minitems: 1
+  maxitems: 20
```
amjjbonvin, line 12: This 20 is what we now advertise. But in principle, e.g. in a scoring scenario, or virtual screening it could be more, correct?

amjjbonvin: Or should these particular case use instead an ensemble of models?

src/haddock/core/mandatory.yaml
```diff
@@ -0,0 +1,18 @@
...
+  title: Input Molecules
+  short: The input molecules that will be used for docking.
+  long:
+    Molecules must be PDB files. These PDB files can be single molecules
```
amjjbonvin, line 16: Rather `Molecules must be provided in PDB format. ...`

src/haddock/core/mandatory.yaml
```diff
@@ -0,0 +1,18 @@
...
+    or ensembles.
```
amjjbonvin, line 17: `or ensembles using the MODEL/ENDMDL statements.`

amjjbonvin:
If this max applies to the split models from an ensemble then capri scoring would no longer be possible…

joaomcteixeira:
No, it does not. It only applies to the maximum number of PDB files that the IVRESSE workflow builder allows the user to input. Currently, we have no `max` for the HADDOCK3. Nontheless, that `20` won't affect the number of models *inside* a PDB file.

amjjbonvin:
All good then!

Issue #321, opened 2022-02-16
Hidden parameters


joaomcteixeira:
As suggested by @amjjbonvin parameters in yaml should accept the `explevel = hidden`. Those parameters should be filtered out from the `haddock-cfg` CLI and from the workflow web interface @sverhoeven

PR #324, opened 2022-02-17
Corrects #322

joaomcteixeira:
In #322 I removed `hidden` from `CNS` input and still considered for `haddock-cfg`. What a lapse. This corrects it properly. Related to #321

### DIFF
--- a/src/haddock/gear/yaml2cfg.py
+++ b/src/haddock/gear/yaml2cfg.py
@@ -132,7 +132,8 @@ def flat_yaml_cfg(cfg):
             # users should not edit yaml files. So this error triggers
             # only during development
             if "explevel" not in values:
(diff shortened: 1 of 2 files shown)


src/haddock/gear/yaml2cfg.py
```diff
@@ -56,7 +57,7 @@ def _yaml2cfg_text(ycfg, module, explevel):
     params = []
     exp_levels = {
         _el: i
-        for i, _el in enumerate(config_expert_levels + ("all",))
+        for i, _el in enumerate(config_expert_levels + ("all", _hidden_level))
```
joaomcteixeira, line 60: It's interesting to see that the actual addition that prevents parameters to be written to `cfg` files is this single addition.

amjjbonvin (approved):

PR #332, opened 2022-02-20
Mdref yaml documentation and CNS scripts updates

amjjbonvin:
And few edits to other modules (small corrections).

All tests passing. Note that some changes might be observed in the mdref output since I changed the default behaviour for random removal of restraints to true (the proper default - before was set to false).


joaomcteixeira:
It looks great!
Question: are the `fix_origin_` and the `shape_` potential candidates to be expandable parameters?

amjjbonvin:
It’s one param per molecule for those

joaomcteixeira (approved):

rvhonorato (approved):

PR #334, opened 2022-02-22
Further improves CONTRIBUTING

joaomcteixeira:
Completes the `CONTRIBUTING` guidelines. This should be enough. The `CONTRIBUTING` file is not an explanation of how to use github or how to dev code, but how to interact with the project. Do you find something missing? You can read the rendered markdown version of the file [here](https://github.com/haddocking/haddock3/blob/contri/CONTRIBUTING.md).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,7 +41,8 @@ To contribute to the HADDOCK3's Python shell, follow these steps:
     1.  `tox -e py39` runs tests in Python 3.9 environment.
     2.  `tox -e lint` shows you errors in the code style.
     3.  `tox -e build` simulates building the HADDOCK3 package.
-    4.  If you want to submit high-quality code, you may wish to run
+    4.  Run the above altogether with the simple `tox` command
+    5.  If you want to submit high-quality code, you may wish to run
     `tox -e radon` to assess your code cyclomatic complexity.
 
 5.  You can work on these `tox` tests until they all pass green before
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -42,17 +46,20 @@ To contribute to the HADDOCK3's Python shell, follow these steps:
 
 5.  You can work on these `tox` tests until they all pass green before
 submitting your PR.
-6.  We also have an `examples` folder with `test` cases that you can run
+6.  We also have an `examples` folder with test cases that you can run
```
amjjbonvin, line 49: Add some explanation that the tests perform a reduced sampling and are only meant to test the machinery. The example folder also contain full config files for production runs.

amjjbonvin: And may-be we should give an estimate of the time required to run those, e.g. on an 8 core computer (i.e. using 7 cores). Could use my laptop for this

rvhonorato (changes requested):
Looks good overall, would also be helpful if there was a checklist (maybe as a template). Based on what you wrote, would be something like
* * *
### Conditions for PR to be approved:
- [ ] PR contains only Python code
- [ ] PR does not add dependency
- [ ] PR does not contain Python Classes
- [ ] PR follows code style
- [ ] PR passes tests 
- [ ] PR does not break license 

If **any** of these criteria are not met, add a detailed justification below:
...
* * *

joaomcteixeira:
Thanks Rodrigo. I will add those in the PR template instead of the contributing, as is usually done in these cases. Okay Alexandre, will add that info too.

rvhonorato (approved):
Changed the grammar of some bits, see if you agree.

> `-test.cfg` examples

This is real-world `double-think` 🐑

joaomcteixeira:
Great :+1: 
Sorry, I normally format `.md` files to max 72 lines. But that's no reason for that, or `80`, or whatever, as long as they are not 400 chars lines. It's perfectly fine as you did. If @amjjbonvin agrees we can merge :wink:

rvhonorato (commented):
Actually I just thought of something else; I am fine with following arbitrary rules, but since we want to have a community developing for haddock3, would be nice if the rules had some sort of motivation - if you think its relevant, of course.

For example: 

- Only python code - why? 
- Avoid use of classes - many of us when we learn python, classes are there on lesson 2. So why not use them? How do they affect HADDOCK in a bad way? 
- Does not add dependencies - if we are working on an open source and community driven project, why not use things that the community have developed? doesn't this go against `Don't Repeat Yourself` paradigm? 
- Why attach testing to `tox`? (This makes tox a dependency)
- Shouldn't different modules be allowed different styles? They should be indeed modular  and separate from the main code.
- What is considered "well documented", docstrings with parameters and return or just a lot of # on top of each line?

Just a few points that I can think off, of course if you don't agree with answering this or don't think this is needed for the community, also fine.

amjjbonvin:
> Only python code - why?
I guess only for the module machinery, but the module might call software written in different languages - we already have Fortran with CNS. 
And some other components might be in different languages. E.g. if we build a gromacs module
Performance is important here.
> Does not add dependencies - if we are working on an open source and community driven project, why not use things that the community have developed? doesn't this go against Don't Repeat Yourself paradigm?

Any new module will add dependencies if it integrates some new software… Unavoidable.

rvhonorato:
Based on what we discussed elsewhere, I've added a new step 6 that explains the motivations behind the contribution rules so that it is also clear for someone outside the group, think this one can be merged! Thanks for making this clear @joaomcteixeira ;)

joaomcteixeira:
Made some improvements for clarity

amjjbonvin (approved):

PR #338, opened 2022-02-25
Add new rule for expandable parameters

joaomcteixeira:
New rule for expandable parameters is defined:

```yaml
nfl1: ...
fle_sta_1_1: ...
fle_end_1_1: ...
nfl2: ...
fle_sta_2_1: ...
fle_end_2_1: ...
```

The above will create groups `fle1` and `fle2`. The pattern is `<param>_<anything>_<N>_<G>`, where `N` is the number of molecule and `G` the group. See how I modified the `yaml` files.

It seems working here. Please test carefully.

Implementation is dirty. Need to clean it next week.


rvhonorato:
I tried my best but don't understand the python code of this part, so I will trust @amjjbonvin CNS judgment on it!

joaomcteixeira:
Merging PR now. I will clean the code in a PR later.

PR #344, opened 2022-03-01
Added new MD refinement module

amjjbonvin:
This PR adds a new scoring module performing a short MD refinement (adapted the mdref module protocol).
The module is based on the emscoring one.

The impact of that MD refinement will need to be properly evaluated, e.g. on the CAPRI set.

A new test/example has been added.

The tests revealed an issue with the contactairs script that has been updated in all relevant modules.

tox passing, but no new module-specific python tests added.


joaomcteixeira (approved):
Fantastic! I am very happy to see these integrations. They mean the code is very much self-explanatory and it communicates.

I have added the new `scoring examples` to the `run_test.py`.

Don't worry about the python tests. there will be a time in the future where we will add plenty of those. For now, already all PRs I am making already covered by tests.

rvhonorato (approved):

PR #346, opened 2022-03-02
Consider cluster centers in `clustfcc`

### LINKED ISSUE #345: selecttopcluster issue in number of models selected
I performed a test run with clustering after rigid body with a threshold of 4 cluster members.

In the selection step I require 20 models from each cluster. If a cluster only has 4 models, all should be used, but only three are selected. My guess is that the cluster center is discarded, which should not be the case.  Models should be selected based on the HADDOCK score. 

Here is an output from the log file:

```
[2022-03-01 22:51:02,647 __init__ INFO] [seletopclusts] Selecting rigidbody_1504.pdb Cluster 827 Model 1
[2022-03-01 22:51:02,647 __init__ INFO] [seletopclusts] Selecting rigidbody_9748.pdb Cluster 827 Model 2
[2022-03-01 22:51:02,648 __init__ INFO] [seletopclusts] Selecting rigidbody_2613.pdb Cluster 827 Model 3
[2022-03-01 22:51:03,707 __init__ INFO] Module [seletopclusts] finished.
```

Cluster 827 in that case does have 4 members, but only 3 are selected.

rvhonorato:
# Pull Request template

with the following criteria:

  those tricky parts.
- [x] `tox` tests pass. Run `tox` command inside the repository folder.
  `python run_tests.py -b`
- [X] You avoided structuring the code into classes as much as possible, using
  by the HADDOCK team

* * *
* 
Fix #345 by adding the cluster centers to `ModuleIO` considering the cluster centers

### DIFF
--- a/src/haddock/modules/analysis/clustfcc/__init__.py
+++ b/src/haddock/modules/analysis/clustfcc/__init__.py
@@ -137,12 +137,17 @@ class HaddockModule(BaseHaddockModule):
             clt_centers = {}
             for clt in clusters:
                 cluster_id = clt.name
+                cluster_center_id = clt.center.name - 1
+                cluster_center_pdb = models_to_cluster[cluster_center_id]
+
                 clt_dic[cluster_id] = []
-                clt_centers[cluster_id] = models_to_cluster[clt.center.name - 1]
+                clt_centers[cluster_id] = cluster_center_pdb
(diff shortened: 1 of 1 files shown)


joaomcteixeira (approved):
Haven't tested it to confirm it works as expected. But the implementation seems clean. Trusting you :wink: :+1:

amjjbonvin (approved):

PR #347, opened 2022-03-02
Verbosity

amjjbonvin:
# Pull Request template

with the following criteria:

  those tricky parts.
  `python run_tests.py -b`
- [ ] You have stick to Python. Talk with us before if you really need to add
- [ ] You avoided structuring the code into classes as much as possible, using
  by the HADDOCK team

Added choices to the verbosity parameter: verbose, normal or quiet
Adapted CNS scripts
Set default verbosity to quiet

### DIFF
--- a/src/haddock/modules/refinement/emref/cns/emref.cns
+++ b/src/haddock/modules/refinement/emref/cns/emref.cns
@@ -11,6 +11,8 @@
 
 if ( $log_level = "verbose" ) then
     set message=normal echo=on end
+elseif ( $log_level = "normal" ) then
+    set message=normal echo=off end
 else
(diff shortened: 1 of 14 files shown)


rvhonorato (approved):

joaomcteixeira (approved):
It is not needed to add the quotes `"` around simple strings. In `yaml` quotes are needed only when writing special characters that may have meanings beyond strings. I removed the quotes for consistency.

amjjbonvin:
Yes but not all modules might use the same params even if they have the same. Doesn't mean the workflow recipe should reuse the same settings.  But things like the verbosity could in principle be defined in the general parameters and not the module specific ones.

PR #348, opened 2022-03-02
Remove hardcoded `top_n` value in `clustfcc`

rvhonorato:
# Pull Request template

with the following criteria:

  those tricky parts.
  `python run_tests.py -b`
- [X] You avoided structuring the code into classes as much as possible, using
  by the HADDOCK team

* * *

### DIFF
--- a/src/haddock/modules/analysis/clustfcc/__init__.py
+++ b/src/haddock/modules/analysis/clustfcc/__init__.py
@@ -145,13 +145,12 @@ class HaddockModule(BaseHaddockModule):
                     clt_dic[cluster_id].append(pdb)
 
             # Rank the clusters
-            #  they are sorted by the top4 models in each cluster
-            top_n = 4
+            #  they are sorted by the topX (threshold) models in each cluster
             score_dic = {}
             for clt_id in clt_dic:
                 score_l = [p.score for p in clt_dic[clt_id]]
                 score_l.sort()
-                top4_score = sum(score_l[:top_n]) / float(top_n)
+                top4_score = sum(score_l[:threshold]) / float(threshold)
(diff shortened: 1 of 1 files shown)


joaomcteixeira (commented):
Before `top_n=4` always and now `threshold` is defined by the `for` loop in line 107. From what I understand `threshold` will be the first value for which clusters are identified. Is this what is intended (honest question)?

src/haddock/modules/analysis/clustfcc/__init__.py
```diff
@@ -145,13 +145,12 @@ def _run(self):
...
             # Rank the clusters
-            #  they are sorted by the top4 models in each cluster
-            top_n = 4
+            #  they are sorted by the topX (threshold) models in each cluster
             score_dic = {}
             for clt_id in clt_dic:
                 score_l = [p.score for p in clt_dic[clt_id]]
                 score_l.sort()
-                top4_score = sum(score_l[:top_n]) / float(top_n)
+                top4_score = sum(score_l[:threshold]) / float(threshold)
```
amjjbonvin, line 153: Is there somewhere a check that the threshold should be >= 1?

amjjbonvin (approved):

amjjbonvin:
> 

Yes this is the intended behaviour.

src/haddock/modules/analysis/clustfcc/__init__.py
rvhonorato: just added!

joaomcteixeira: Is defined by the for loop and the `range`

joaomcteixeira:
:+1:

PR #350, opened 2022-03-02
make `topoaa.molX` and `mol_` parameter expandable

joaomcteixeira:
- [x] Your code is well documented. Functions and classes have proper docstrings as explained in contributing guidelines. You wrote explanatory comments for those tricky parts.
- [x] You wrote tests for the new code.
- [x] `tox` tests pass. Run `tox` command inside the repository folder.
- [x] `-test.cfg` examples execute without errors. Inside `examples/` run `python run_tests.py -b`
- [x] You have stick to Python. Talk with us before if you really need to add other programming languages to HADDOCK3
- [x] You avoided structuring the code into classes as much as possible, using small functions instead. But, you can use classes if there's a purpose.
- [x] PR does not add any *install dependencies* unless permission was granted by the HADDOCK team

---

Define `mol` parameters as expandable parameters. These include the `molX` in `topoaa` and `mol_fix_origin` and `mol_shape` in some other modules.

For `topoaa` the `prot_segid` is defined by the alphabet progression **unless** the user has defined a value.

**todo:**
- [ ] implement forced progression for params except `mol_`


joaomcteixeira:
Define `mol` parameters as expandable parameters.

These include the `molX` in `topoaa` and `mol_fix_origin` and `mol_shape` in some other modules.

For `topoaa` the `prot_segid` is defined by the alphabet progression **unless** the user has defined a value.

`capri.tsv` values are the same as in `main` for all the `test.cfg` cases.

amjjbonvin (commented):
How does it work? Will it create by default all values (e.g. mol_fix_origin and mol_shape), set those to the default values and write them to the CNS inp files? And of course take any value the user might define in the config file?

Also how will that connect to the i-vresse workflow builder/editor? 

Also note that there are other molecule-related parameters that could be handled that way, e.g. `nseg_1` `nfle_1` `nrair_1` and all associated parameters.

rvhonorato:
Approved but I don't understand anything about this part, trusting you @joaomcteixeira :)

joaomcteixeira:
I am making a presentation for tomorrow. I think it's best to explain to all

amjjbonvin:
Ok - all clear 

> Yes, but those are different. Those are associated with the rair_sta_1_1 parameters, for example. nseg_1 (and others) are a different kind of parameters that we can decide to populate automatically from the number of groups defined in the rair_sta_1_1 type of parameters. The rair_sta_1_1 type of parameter are expandable, as we have seen before.

These are molecule specific as the first index relates to the molecule and the second to the segment.

joaomcteixeira:
> These are molecule specific as the first index relates to the molecule and the second to the segment.

Yes. One thing I can do is to block to the maximum number of molecules also.

joaomcteixeira:
Okay, I will add that in another issue, because this one was about cleaning the implementation and adding the `mol`

amjjbonvin (commented):
Looking good - one last comment:

Shouldn't the yaml file for the topoaa module be update so that the documentation (short and long) explains that this block and parameters can be duplicated for additional molecules?

joaomcteixeira:
Added the documentation.
FYI, in the python shell, I treat these special cases of nested dictionaries with having or not the `default` key. That is, actual parameters have the `default` key, and if `default` is not present, the entry is treated as a subdictionary, as is the case of mol1. Need not to worry about this, was just to let it here stated. :+1:
If you agree it can be merged.

amjjbonvin (approved):

PR #351, opened 2022-03-02
update PR template

joaomcteixeira:
Now that we have used the PR template for a couple of PRs... this updates the PR template.


rvhonorato (approved):

amjjbonvin:
May-be since a lot of the protocols are related to CNS code/scripts, add an option for this in the template?

joaomcteixeira:
Wasn't expecting others to edit CNS, but added a line now. Surely this file is going to be re-edited many times in the future as we interact with it, it's normal.

PR #213, opened 2021-12-20
Implement rank column in `caprieval`

rvhonorato:
This PR implements a new ranking column in `caprieval` and closes #212.

It adds two new parameters.

```
rankby = 'score'
rank_ascending = true
```

With the changes on this PR its possible to use one column to do the ranking and another for the sorting, let me know if this is not the desired behavior of the feature.

PR #353, opened 2022-03-03
Impelement cluster-based output in `caprieval`

### LINKED ISSUE #352: Add cluster-based analysis to `caprieval`

rvhonorato:
# Pull Request template

with the following criteria:

  those tricky parts.
- [X] You wrote tests for the new code.
  `python run_tests.py -b`
- [x] You avoided structuring the code into classes as much as possible, using
  by the HADDOCK team
- [X] You get extra bonus if you write tests for already existing code :godmode:

* * *

Closes #352 by implementing a cluster-based output to `caprieval`;

it now has two output files: `capri_ss.tsv` for single-structure and `capri_clt.tsv` for cluster-based, if there is no cluster information in the models, then only `capri_ss.tsv` is generated. 

The models inside each cluster are ranked by score; so by setting `clt_threshold=4` (new parameter) the average metrics will be calculated over the top4 of each cluster.

There is a scenario in which `clt_threshold > total_number_of_clusters` that results in the metrics being under-evaluated; ex: each cluster has 2 models but if `clt_threshold=4`, the values of these 2 models will be divided by 4. When this happens, the module will set `under_eval=yes` in `capri_clt.tsv`.

## Example 1: `capri_clt.tsv`

```tsv
########################################
# `caprieval` cluster-based analysis
#
# > rankby_key=score
# > rank_ascending=True
# > sortby_key=score
# > sort_ascending=True
# > clt_threshold=4
#
# NOTE: if under_eval=yes, it means that there were less models in a cluster than
#    clt_threshold, thus these values were under evaluated.
#   You might need to tweak the value of clt_thresholdor change some parameters
#    in `clustfcc` depending on your analysis.
#
########################################
caprieval_rank	cluster_rank	cluster_id	n	under_eval	score	irmsd	fnat	lrmsd	dockq
1	1	2	2	yes	-11.379	1.608	0.146	2.915	0.192
2	2	1	2	yes	-8.921	2.616	0.056	4.638	0.108
```
## Example 2: `capri_clt.tsv`

```tsv
########################################
# `caprieval` cluster-based analysis
#
# > rankby_key=score
# > rank_ascending=True
# > sortby_key=score
# > sort_ascending=True
# > clt_threshold=4
#
# NOTE: if under_eval=yes, it means that there were less models in a cluster than
#    clt_threshold, thus these values were under evaluated.
#   You might need to tweak the value of clt_threshold or change some parameters
#    in `clustfcc` depending on your analysis.
#
########################################
caprieval_rank	cluster_rank	cluster_id	n	under_eval	score	irmsd	fnat	lrmsd	dockq
1	1	1	4	-	-33.461	2.358	0.354	4.406	0.479
2	2	3	4	-	-26.601	10.927	0.111	17.980	0.104
3	3	5	4	-	-20.855	5.686	0.111	10.384	0.193
4	4	6	4	-	-20.489	6.591	0.076	11.419	0.161
5	5	2	4	-	-18.540	9.972	0.069	16.070	0.104
6	6	8	4	-	-16.757	3.774	0.278	7.214	0.333
7	7	4	4	-	-14.531	2.544	0.361	4.915	0.457
8	8	7	4	-	-14.480	10.449	0.104	17.195	0.107
```

## Example 3 `capri_clt.tsv` `rankby/sortby=irmsd`

```
########################################
# `caprieval` cluster-based analysis
#
# > rankby_key=irmsd
# > rank_ascending=True
# > sortby_key=irmsd
# > sort_ascending=True
# > clt_threshold=4
#
# NOTE: if under_eval=yes, it means that there were less models in a cluster than
#    clt_threshold, thus these values were under evaluated.
#   You might need to tweak the value of clt_threshold or change some parameters
#    in `clustfcc` depending on your analysis.
#
########################################
caprieval_rank	cluster_rank	cluster_id	n	under_eval	score	irmsd	fnat	lrmsd	dockq
1	1	1	4	-	-33.461	2.358	0.354	4.406	0.479
2	7	4	4	-	-14.531	2.544	0.361	4.915	0.457
3	6	8	4	-	-16.757	3.774	0.278	7.214	0.333
4	3	5	4	-	-20.855	5.686	0.111	10.384	0.193
5	4	6	4	-	-20.489	6.591	0.076	11.419	0.161
6	5	2	4	-	-18.540	9.972	0.069	16.070	0.104
7	8	7	4	-	-14.480	10.449	0.104	17.195	0.107
8	2	3	4	-	-26.601	10.927	0.111	17.980	0.104
```

## `capri_ss.tsv` (unchanged)
```tsv
model	caprieval_rank	score	irmsd	fnat	lrmsd	ilrmsd	dockq	cluster-id	cluster-ranking	model-cluster-ranking
../01_rigidbody/rigidbody_16.pdb	1	-24.729	3.051	0.361	5.911	4.083	0.410	2	1	1
../01_rigidbody/rigidbody_12.pdb	2	-21.579	5.714	0.111	10.224	8.199	0.195	1	2	1
../01_rigidbody/rigidbody_20.pdb	3	-20.785	3.383	0.222	5.748	4.111	0.358	2	1	2
../01_rigidbody/rigidbody_2.pdb	4	-14.105	4.751	0.111	8.330	6.732	0.237	1	2	2
```

### DIFF
--- a/examples/docking-protein-protein/docking-protein-protein-cltsel-test.cfg
+++ b/examples/docking-protein-protein/docking-protein-protein-cltsel-test.cfg
@@ -31,18 +31,22 @@ hise_1 = 15
 [rigidbody]
 tolerance = 20
 ambig_fname = "data/e2a-hpr_air.tbl"
-sampling = 20
+sampling = 25
+
+[caprieval]
+#tolerance = 20
+reference_fname = "data/e2a-hpr_1GGR.pdb"
 
 [clustfcc]
 #tolerance = 20
-threshold=1
+threshold = 4
 
 [seletopclusts]
 #tolerance = 20
 ## select the best 4 clusters
 # top_cluster = [1, 2, 3, 4]
 ## select all the clusters
-top_cluster = [1,2,3,4,5,6,7,8,8,10]
+top_cluster = [1, 2, 3, 4]
 ## select the best 4 models of each cluster
 top_models = 4
 ##select all the models
@@ -50,6 +54,7 @@ top_models = 4
 
 [caprieval]
 #tolerance = 20
+clt_threshold=4
 reference_fname = "data/e2a-hpr_1GGR.pdb"
 
 [flexref]
--- a/examples/docking-protein-protein/docking-protein-protein-cltsel-test.cfg
+++ b/examples/docking-protein-protein/docking-protein-protein-cltsel-test.cfg
@@ -31,7 +31,7 @@ hise_1 = 15
 [rigidbody]
 tolerance = 20
 ambig_fname = "data/e2a-hpr_air.tbl"
-sampling = 25
+sampling = 100
 
 [caprieval]
 #tolerance = 20
@@ -46,7 +46,7 @@ threshold = 4
 ## select the best 4 clusters
 # top_cluster = [1, 2, 3, 4]
 ## select all the clusters
-top_cluster = [1, 2, 3, 4]
+top_cluster = []
 ## select the best 4 models of each cluster
 top_models = 4
 ##select all the models
@@ -54,6 +54,7 @@ top_models = 4
 
 [caprieval]
 #tolerance = 20
+sortby="irmsd"
 clt_threshold=4
 reference_fname = "data/e2a-hpr_1GGR.pdb"
 
--- a/src/haddock/modules/analysis/caprieval/defaults.yaml
+++ b/src/haddock/modules/analysis/caprieval/defaults.yaml
@@ -155,3 +155,13 @@ lovoalign_exec:
   long: No long description yet
   group: ''
   explevel: easy
+clt_threshold:
+  default: 4
+  type: integer
+  min: 0
+  max: 1000
+  title: No title yet
+  short: No short description yet
+  long: No long description yet
+  group: ''
+  explevel: easy
--- a/src/haddock/modules/analysis/caprieval/defaults.yaml
+++ b/src/haddock/modules/analysis/caprieval/defaults.yaml
@@ -1,167 +1,180 @@
 reference_fname:
   default: ''
   type: file
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Reference structure
+  short: Structure to be used when calculating the CAPRI metrics.
+  long: Structure to be used when calculating the CAPRI metrics.
+    If none is defined then the lowest scoring structure is selected by default.
+  group: 'analysis'
   explevel: easy
 irmsd:
   default: true
   type: boolean
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Calculate the I-RMSD
+  short: Perform I-RMSD calculations.
+  long: Perform I-RMSD calculations.
+  group: 'analysis'
   explevel: easy
 fnat:
   default: true
   type: boolean
   title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  short: Perform FNAT calculations.
+  long: Perform FNAT calculations.
+  group: 'analysis'
   explevel: easy
 lrmsd:
   default: true
   type: boolean
   title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  short: Perform L-RMSD calculations.
+  long: Perform L-RMSD calculations.
+  group: 'analysis'
   explevel: easy
 ilrmsd:
   default: true
   type: boolean
   title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  short: Perform I-L-RMSD calculations.
+  long: Perform I-L-RMSD calculations.
+  group: 'analysis'
   explevel: easy
 dockq:
   default: true
   type: boolean
   title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  short: Evaluate DockQ.
+  long: Evaluate DockQ.
+  group: 'analysis'
   explevel: easy
 irmsd_cutoff:
   default: 10.0
   type: float
-  min: -9999
+  min: 0.0
   max: 9999
   precision: 3
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Distance cutoff (Å) to consider the interface
+  short: Distance cutoff (Å) to consider the interface.
+  long: Distance cutoff (Å) to consider the interface.
+  group: 'analysis'
   explevel: easy
 fnat_cutoff:
   default: 5.0
   type: float
-  min: -9999
+  min: 0.0
   max: 9999
   precision: 3
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Distance cutoff (Å) to consider the interface
+  short: Distance cutoff (Å) to consider the interface.
+  long: Distance cutoff (Å) to consider the interface.
+  group: 'analysis'
   explevel: easy
 lrmsd_cutoff:
   default: 10.0
   type: float
-  min: -9999
+  min: 0.0
   max: 9999
   precision: 3
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Distance cutoff (Å) to consider the interface
+  short: Distance cutoff (Å) to consider the interface.
+  long: Distance cutoff (Å) to consider the interface.
+  group: 'analysis'
   explevel: easy
 receptor_chain:
   default: A
   type: string
-  minchars: 0
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  minchars: 1
+  maxchars: 1
+  title: Receptor ChainID
+  short: Receptor ChainID.
+  long: Receptor ChainID.
+  group: 'analysis'
   explevel: easy
 ligand_chain:
   default: B
   type: string
-  minchars: 0
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  minchars: 1
+  maxchars: 1
+  title: Ligand ChainID
+  short: Ligand ChainID.
+  long: Ligand ChainID.
+  group: 'analysis'
   explevel: easy
 rankby:
   default: score
   type: string
-  minchars: 0
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  choices:
+    - score
+    - irmsd
+    - lrmsd
+    - ilrmsd
+    - fnat
+    - dockq
+  title: Rank the structures in the output file by this value
+  short: Which value should be used to rank the structures in the output file.
+  long: Which value should be used to rank the structures in the output file, this will determine the `caprieval_rank`
+  group: 'analysis'
   explevel: easy
 rank_ascending:
   default: true
   type: boolean
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Rank in ascending order
+  short: Rank in ascending order.
+  long: Rank in ascending order.
+  group: 'analysis'
   explevel: easy
 sortby:
   default: score
   type: string
-  minchars: 0
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  choices:
+    - score
+    - irmsd
+    - lrmsd
+    - ilrmsd
+    - fnat
+    - dockq
+  title: Sort the structures in the output file by this value
+  short: Which value should be used to sort the output.
+  long: Which value should be used to sort the output, the output table will be sorted accordingly.
+  group: 'analysis'
   explevel: easy
 sort_ascending:
   default: true
   type: boolean
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Sort in ascending order
+  short: Sort in ascending order.
+  long: Sort in ascending order.
+  group: 'analysis'
   explevel: easy
 alignment_method:
   default: sequence
   type: string
-  minchars: 0
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  choices:
+    - structure
+    - sequence
+  title: Alignment method used to match the numbering.
+  short: Which alignment method should be used to match the numbering.
+  long: Alignment method used to match the numbering. Sequence alignment is haddock3-friendly but might not produce best results
+    for structures that are too different, for that use
+  group: 'analysis'
   explevel: easy
 lovoalign_exec:
   default: ''
   type: string
   minchars: 0
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  maxchars: 999
+  title: Location (path) of the LovoAlign executable
+  short: Location (path) of the LovoAlign executable.
+  long: Location (path) of the LovoAlign executable.
+  group: 'analysis'
   explevel: easy
 clt_threshold:
   default: 4
   type: integer
-  min: 0
+  min: 1
   max: 1000
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Threshold (n) used for the average evaluation in the cluster-based output
+  short: Threshold (n) used for the average evaluation in the cluster-based output.
+  long: Threshold (n) used for the average evaluation in the cluster-based output.
+  group: 'analysis'
   explevel: easy
--- a/src/haddock/modules/analysis/caprieval/defaults.yaml
+++ b/src/haddock/modules/analysis/caprieval/defaults.yaml
@@ -7,38 +7,47 @@ reference_fname:
     If none is defined then the lowest scoring structure is selected by default.
   group: 'analysis'
   explevel: easy
+
 irmsd:
   default: true
   type: boolean
   title: Calculate the I-RMSD
-  short: Perform I-RMSD calculations.
-  long: Perform I-RMSD calculations.
+  short: Perform Interface RMSD calculations.
+  long: Calculate the Interface Root Mean Square Deviation.
+    This calculation evaluates the RMSD between two interfaces, not considering the rest of the complex.
   group: 'analysis'
   explevel: easy
+
 fnat:
   default: true
   type: boolean
-  title: No title yet
+  title: Calculate the FNAT.
   short: Perform FNAT calculations.
-  long: Perform FNAT calculations.
+  long: Calculate the Frequency of Native Contacts, this calculation evaluates how many contacts observed
+    in the reference are present in the model.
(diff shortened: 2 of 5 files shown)


amjjbonvin (commented):
What is the difference between caprieval_rank and cluster_rank?  
To me these should be the same, i.e. only the rank based on the HADDOCK score is relevant here.

The cluster_id is I assume simply the cluster number returned by cluster_fcc.

Also, what are the RMSD dockQ and other metrics reported? The average, or the best within the number of selected models within one cluster?

And may-be also good to add the standard deviations to the average values.

amjjbonvin:
Also the single structure file is sorted based on the HADDOCK score. Why not do the same for the cluster-based file?

rvhonorato:
> 1. What is the difference between caprieval_rank and cluster_rank?

`cluster_rank` is always based on the score and its topN is given by a parameter in `clustfcc`, `caprieval_rank` can vary according to the metric; let's say you want caprieval to display the clusters with lowest i-rmsd first (this is the example 3 I posted above), but in in detail below:

Cluster ID 1 is also the best scoring cluster, and it contains the lowest i-rmsds, then:
```
cluster_id = 1
cluster_rank = 1 (best score)
caprieval_rank = 1 (lowest i-rmsd)
```

Cluster ID 4 has the second lowest i-rmsds and is ranked 7 in relation to scores, then:
```
cluster_id = 4
cluster_rank = 7 (7th best in relation to score)
caprieval_rank = 2 (second lowest in i-rmsd)
```

> 2. To me these should be the same, i.e. only the rank based on the HADDOCK score is relevant here.

Sure, I can disable this ranking according to the capri metrics in the cluster-based output.

> 3. Also, what are the RMSD dockQ and other metrics reported? The average, or the best within the number of selected models within one cluster?

The metrics reported for the clustering are given according to the `clt_threshold` parameter, example:

```
clt_threshold = 4
Cluster #1 has 20 elements -> the top4 elements will be selected and divided by 4
Cluster #5 has 5 elements -> the top4 elements will be selected and divided by 4
Cluster #10 has 2 elements -> the top2 elements will be selected and divided by 4
```

When the number of elements is smaller than the `clt_threshold`, the cluster will be be under sampled, and it will be flagged with `yes` in the `under_eval` column in the output file, these can then be easily filtered out during more advanced analysis.

> 4. And may-be also good to add the standard deviations to the average values.

Indeed good point, will add that.

> 5. Also the single structure file is sorted based on the HADDOCK score. Why not do the same for the cluster-based file?

Se the explanation above (point 1), but to clarify - `caprieval` has the parameters `rankby` and `sortby`. By default all models are ranked and sorted by `score`, however they can be ranked and sorted by any of the metrics. 

For example you want to `rankby=score` and `sortby=irmsd`, or for parametrization purposes, `rankby=irmsd` and `sortby=score`

I have explained this before when the ranking was implemented in https://github.com/haddocking/haddock3/pull/213#issuecomment-998021129, but again here for completion:

Using the following parameters, this is how the single-structure would look like:
```
rankby = 'score'
sortby = 'irmsd'

             model            caprieval_rank     score     irmsd      fnat     lrmsd    ilrmsd
01_rigidbody/rigidbody_15.pdb           7   -22.486     1.367     0.500     3.158     2.285
01_rigidbody/rigidbody_4.pdb            1   -33.049     1.621     0.389     3.589     2.908
01_rigidbody/rigidbody_9.pdb           17   -13.026     2.549     0.417     7.081     4.720
01_rigidbody/rigidbody_16.pdb           3   -24.729     2.950     0.361     7.472     4.502
01_rigidbody/rigidbody_20.pdb          10   -20.785     3.289     0.222     6.922     4.261
```

The same logic applies to the cluster-based, see explanation above (point 1).

amjjbonvin:
> What is the difference between caprieval_rank and cluster_rank?
Ok - clear - need to add the documentation to the yaml file to make things clear then

> To me these should be the same, i.e. only the rank based on the HADDOCK score is relevant here.
> Sure, I can disable this ranking according to the capri metrics in the cluster-based output.
> 
No need for that. But the default behaviour for the caprieval_rank should be the score.

> Also, what are the RMSD dockQ and other metrics reported? The average, or the best within the number of selected models within one cluster?
> The metrics reported for the clustering are given according to the clt_threshold parameter, example
> 
I.e. the average in that case.

> 
> And may-be also good to add the standard deviations to the average values.
> Indeed good point, will add that.
> 
> Also the single structure file is sorted based on the HADDOCK score. Why not do the same for the cluster-based file?
> Se the explanation above (point 1), but to clarify - caprieval has the parameters rankby and sortby. By default all models are ranked and sorted by score, however they can be ranked and sorted by any of the metrics.
> 
Default should always be score (because we might use it also when we don’t have a reference)

rvhonorato:
Yes, its always `score` by default - both ranking and sorting! Will add the standard deviations.

rvhonorato:
Added the standard deviation and the documentations to the `defaults.yml`

examples/docking-protein-protein/docking-protein-protein-cltsel-test.cfg
```diff
@@ -31,25 +31,30 @@ hise_1 = 15
...
-threshold=1
+threshold = 4
 
 [seletopclusts]
 #tolerance = 20
 ## select the best 4 clusters
 # top_cluster = [1, 2, 3, 4]
 ## select all the clusters
-top_cluster = [1,2,3,4,5,6,7,8,8,10]
+top_cluster = [1, 2, 3, 4]
```
amjjbonvin, line 49: Why not simply: `top_cluster = []` since it comes after the comment select all the clusters...

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
 irmsd:
   default: true
   type: boolean
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Calculate the I-RMSD
+  short: Perform I-RMSD calculations.
+  long: Perform I-RMSD calculations.
```
amjjbonvin, line 15: Could here expand to define what is the i-RMSD

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
   explevel: easy
 fnat:
   default: true
   type: boolean
   title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  short: Perform FNAT calculations.
+  long: Perform FNAT calculations.
```
amjjbonvin, line 23: Could here expand to define what it the Fnat

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
   explevel: easy
 lrmsd:
   default: true
   type: boolean
   title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  short: Perform L-RMSD calculations.
+  long: Perform L-RMSD calculations.
```
amjjbonvin, line 31: Could here expand to define what it the l-RMSD

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
   explevel: easy
 ilrmsd:
   default: true
   type: boolean
   title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  short: Perform I-L-RMSD calculations.
+  long: Perform I-L-RMSD calculations.
```
amjjbonvin, line 39: Could here expand to define what it the i-l-RMSD

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
   explevel: easy
 dockq:
   default: true
   type: boolean
   title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  short: Evaluate DockQ.
+  long: Evaluate DockQ.
```
amjjbonvin, line 47: Could here expand to define what it the DockQ score

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
+  group: 'analysis'
   explevel: easy
 irmsd_cutoff:
   default: 10.0
   type: float
-  min: -9999
+  min: 0.0
```
amjjbonvin, line 53: Does a minimum value of 0 make sense?  May-be set it to 3?

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
+  title: Distance cutoff (Å) to consider the interface
+  short: Distance cutoff (Å) to consider the interface.
+  long: Distance cutoff (Å) to consider the interface.
+  group: 'analysis'
   explevel: easy
 fnat_cutoff:
   default: 5.0
   type: float
-  min: -9999
+  min: 0.0
```
amjjbonvin, line 64: Same comment as for i-RMSD
And I would put an upper limit of say 20A to avoid bad choices and heavy calculations

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
   max: 9999
   precision: 3
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Distance cutoff (Å) to consider the interface
+  short: Distance cutoff (Å) to consider the interface.
+  long: Distance cutoff (Å) to consider the interface.
```
amjjbonvin, line 69: "to consider the interface" is rather poorly defined.

May-be better: used to define interface contacts

And same comment for the RMSD values, but in that case rather "to define interface residues"

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
   type: float
-  min: -9999
+  min: 0.0
   max: 9999
   precision: 3
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Distance cutoff (Å) to consider the interface
```
amjjbonvin, line 78: For the l-RMSD there is not interface required to calculate it...  I.e. this lrmsd_cutoff should not be needed... Is it used???

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  minchars: 1
+  maxchars: 1
+  title: Receptor ChainID
+  short: Receptor ChainID.
+  long: Receptor ChainID.
```
amjjbonvin, line 90: Relevant for the l-RMSD and i-l-RMSD calculations. So this should be explained. After all this is a "long" explanation...

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  minchars: 1
+  maxchars: 1
+  title: Ligand ChainID
+  short: Ligand ChainID.
+  long: Ligand ChainID.
```
amjjbonvin, line 100: Same comment as for receptor but for the ligand

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,186 @@
...
+  group: 'analysis'
   explevel: easy
 rankby:
```
amjjbonvin, line 103: There is confusion in these explanations between rank and sort.
E.g. there is no single mention of the score (e.g. HADDOCK score). 

To me the rank should be defined by the scoring function used (and nothing else) - and this should be related to the "cluster_score" or "structure/model_score" values. And the sorting can be made on different values, including the score, and this would be related to the caprieval_rank value.

amjjbonvin (changes requested):

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: wow this makes no sense, it's not used and should not be there indeed

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: ok, I will remove the ranking logic then

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: changed

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: changed

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: changed

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: updated

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: updated

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: updated

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: updated

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: updated

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: updated

src/haddock/modules/analysis/caprieval/defaults.yaml
rvhonorato: updated

amjjbonvin (approved):
Please check that the descriptions I edited in the yaml file match what happens in the code

examples/docking-protein-protein/docking-protein-protein-cltsel-test.cfg
```diff
@@ -31,25 +31,31 @@ hise_1 = 15
 [rigidbody]
 tolerance = 20
 ambig_fname = "data/e2a-hpr_air.tbl"
-sampling = 20
+sampling = 100
```
joaomcteixeira, line 34: This is a `test.cfg` made to execute fast. Isn't `sampling = 100` too much?

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -1,157 +1,175 @@
...
   default: score
   type: string
-  minchars: 0
-  maxchars: 100
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
-  explevel: easy
+  minchars: -999999999999
```
joaomcteixeira, line 116: A bit too much even if it is not needed. `(0, 20)` is better.

examples/docking-protein-protein/docking-protein-protein-cltsel-test.cfg
amjjbonvin: Rigid body is fine - fast anyway - and for the cluster selection you need enough models

joaomcteixeira:
this is not allowed:

```yaml
long: Lorem ipsum dolor sit amet, consectetur adipiscing elit. Negabat igitur ullam esse artem,
quae ipsa a se proficisceretur; Si longus, levis. Nihilo beatiorem esse Metellum quam Regulum.
``` 

give spaces;

```yaml
long: Lorem ipsum dolor sit amet, consectetur adipiscing elit. Negabat igitur ullam esse artem,
   quae ipsa a se proficisceretur; Si longus, levis. Nihilo beatiorem esse Metellum quam Regulum.
```

mgiulini:
Having a look at the `capri_clt.tsv` files I noticed that the clusters with a population lower than `clt_threshold` are correctly flagged with `under_eval = yes`. Nonetheless, the fact that the values of irmsd, fnat, dockq, lrmsd and ilrmsd are still divided by `clt_threshold` could be misleading. I would keep the `under_eval` flag, but correcting the values by introducing an additional check here:
https://github.com/haddocking/haddock3/blob/ed55a99e729bbd9b365de78341911a3fed6e9a52/src/haddock/modules/analysis/caprieval/capri.py#L611-L616
what do you think?

amjjbonvin:
Is n then set to `clt_threshold` ? If that's the case then indeed this should be corrected.

rvhonorato:
Yes n is `clt_threshold` and yes, that's exactly why the `under_eval` is there because its divided by a larger number.

Because if you don't mark them, and make the correct you suggest above, they'll be artificially better, for example:

```
# cluster 1 
model1 -25
model2 -30
average score = -27,5

# cluster 2
model1 -35
model2 -25
model3 -30
model4 -10 
average score = -25
```

Here we'd select **cluster 1** instead of **cluster 2** and its what will happen if we correct the `clt_threshold`. 

Is this the desired behavior?

amjjbonvin:
But if the threshold is 4, why would you even consider cluster1 in this case?

mgiulini:
For low-level analysis purposes I would keep the under-populated clusters in the `capri_clt.tsv` file together with the flag and the correct averages. I would exclude them from plot routines and other high-level analysis. 

Does this sound meaningful?

amjjbonvin:
Yes - the reported scores should be the correct one, i.e. divided by the correct number of models.

rvhonorato:
> But if the threshold is 4, why would you even consider cluster1 in this case?

I think in the v2 version we automatically exclude those at some point, here I'm considering them nonetheless, maybe not the best

> For low-level analysis purposes I would keep the under-populated clusters in the capri_clt.tsv file together with the flag and the correct averages. I would exclude them from plot routines and other high-level analysis.

Yes sounds correct!

rvhonorato:
Just for clarity, which should we do then?
1) not consider a given cluster if N < clt_threshold 
2) correct the scores of the cluster if N < clt_threshold

amjjbonvin:
> Just for clarity, which should we do then?
> 
> not consider a given cluster if N < clt_threshold
If we define a threshold then we should use it!
Thus indeed not consider clusters that have less than the threshold (even if they might be reported in the file)
> correct the scores of the cluster if N < clt_threshold
The correct average score should be reported (i.e. based on the real number of models in the cluster)

PR #363, opened 2022-03-11
Googleform

joaomcteixeira:
Adds a section for users to sign the google forms supporting HADDOCK3. Since the form is on the group's page, I thought it could be useful having it here as well. What do you think?


rvhonorato:
Don't think support is the correct word?

joaomcteixeira:
Don't know. Really need your help with the proper wording.

amjjbonvin (approved):
Made a few edits. Check it out

joaomcteixeira:
Sounds perfect!

PR #378, opened 2022-04-06
Zerofill

joaomcteixeira:

- [ ] Your PR is about CNS

---

Modules folders are now zerofilled according to to the number of modules. If there are less than 10 modules, module folder will be prefix with only one digit. More than 10 modules, would be prefixed with 0, and so on. Thanks @douweschulte for the input :wink:

### DIFF
--- a/src/haddock/libs/libutil.py
+++ b/src/haddock/libs/libutil.py
@@ -281,6 +281,16 @@ def get_number_from_path_stem(path):
     return int(number)
 
 
+def get_number_of_digits(num):
+    """
+    Get the number of digits of a number.
+
+    10 has two digits.
+    100 has three digits.
+    """
+    return len(str(num))
+
+
 def sort_numbered_paths(*paths):
     """
     Sort input paths to tail number.
--- a/src/haddock/libs/libutil.py
+++ b/src/haddock/libs/libutil.py
@@ -7,6 +7,7 @@ import subprocess
 import sys
 from copy import deepcopy
 from functools import partial
+from math import log10, ceil
 from os import cpu_count
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


src/haddock/libs/libutil.py
```diff
@@ -281,6 +281,16 @@ def get_number_from_path_stem(path):
...
 
 
+def get_number_of_digits(num):
+    """
+    Get the number of digits of a number.
+
+    10 has two digits.
+    100 has three digits.
+    """
+    return len(str(num))
```
douweschulte, line 291: `log10(num+1)` also gives this result, and 'should' be faster (although that does not matter here). But I feel like a numerical problem should not be solved by using strings ;-)

src/haddock/modules/__init__.py
```diff
@@ -198,7 +198,8 @@ def _load_previous_io(self):
 
     def previous_path(self):
         """Give the path from the previous calculation."""
-        previous = sorted(list(self.path.resolve().parent.glob('[0-9][0-9]*/')))
+        # https://regex101.com/r/4qRTjn/1
+        previous = sorted(list(self.path.resolve().parent.glob('[0-9]*/')))
```
douweschulte, line 202: Why does this need to be changed? It already parsed one or more digits which is what is needed. In this new version it parses zero or more digits which I think is not what is needed here. Maybe `\d+` is more logical, which is equivalent to `[0-9]+` which is equivalent to `[0-9][0-9]*`.

tests/test_libutil.py
```diff
@@ -165,3 +166,16 @@ def test_extract_keys_recursive(inp, expected):
...
     assert result == expected
+
+
+@pytest.mark.parametrize(
+    "num,expected",
+    [
+        (0, 1),
+        (1, 1),
+        (22, 2),
+        (335, 3),
```
douweschulte, line 177: Best to test here are around the border of an extra number so `99` and `100`. As these are the most likely to cause problems.

douweschulte (changes requested):
See my inline comments.

douweschulte (approved):

src/haddock/libs/libutil.py
joaomcteixeira: n1!

Issue #382, opened 2022-04-06
Continue previous runs in new folder


joaomcteixeira:
We want to take an haddock3 run and continue it in a different direction in another run folder. There are several possibilities this can be done, and they can all be implemented together:

- [ ] take the whole run or part of it, copy it to a new folder, and start it from a new configuration that defines only the new steps.
- [ ] take only the topology folder and the desired step and go forward as in the previous point
- [ ] give a folder through an  `haddock3` command parameter and a config file so the run continues.

joaomcteixeira:
Although we have implemented #378 and #383, I wonder if that flexibility in the module's folder number prefixes will perturb the functionalities we want to add here. The problem may arise if we have a run with 8 modules and than we extend it for a run with 12 modules. How are the folders and files from the previous run be handled? We will have a mixture of folders starting with `0` and without. This example would result in,

```
0_topoaa
1_...
1_...
1_...
1_...
1_...

**Comment by rvhonorato** [comment]
This already exists, doesn't it?

The below works just fine:

```bash
$ haddock3 workflow-with-two-steps.cfg
---
$ mkdir ~/other-folder
$ cp -r data ~/other-folder
$ cp workflow-with-two-steps.cfg ~/other-folder/workflow-with-five-steps.cfg
$ cd ~/other-folder
# edit workflow-with-five-steps.cfg and add the new steps
$ haddock3 --restart 2 workflow-with-five-steps.cfg
```

joaomcteixeira:
Yes. That part exists. But there are some variations of that protocol that I am unsure will work straight with the current implementation. I want to use this issue to review all the possibilities, implement what's left, and make docs about it also. The example you show definitively works already.

joaomcteixeira:
# Case 1

*describing the behavior for case 1*

1. take a completed run, for example `protein-protein` docking:

```
0_topoaa
1_rigidbody
2_caprieval
3_seletop
4_flexref
5_caprieval
6_emref
7_caprieval
8_clustfcc
data
log
```

Now we want to take `4_flexref` to another place and continue a different run from there. How should we handle it? A possibility is:

Copy `0_topoaa` and `4_flexref` to a new folder:

```
0_topoaa
4_flexref
```

Reprepare the configuration file only with the steps we need onwards:

```toml
[caprieval]
reference_fname = "data/e2a-hpr_1GGR.pdb"

[emref]
tolerance = 20
ambig_fname = "data/e2a-hpr_air.tbl"

[clustfcc]
[...]

PR #383, opened 2022-04-06
Corrects edge case in zero filling modules folder names

joaomcteixeira:

- [ ] Your PR is about CNS

---

Corrects edge cases in #378 when having 10 modules because module count starts at `0`.

### DIFF
--- a/src/haddock/libs/libutil.py
+++ b/src/haddock/libs/libutil.py
@@ -293,6 +293,21 @@ def get_number_of_digits(num):
     return max(ceil(log10(num + 1)), 1)
 
 
+def get_zerofill_for_modules(modules):
+    """
+    Get a the prefix zerofill for modules.
+
+    If there are 5 modules, zerofill digits is 1.
+    If there are 10 modules, zerofill digits is 1 because counting
+    starts at 0 (for topoaa).
+    If there are 100 modules, zerofill digits is 2 because counting
+    starts at 0 (for topoaa).
+
+    This function is used in combination with `zero_fill`.
+    """
+    return max(1, get_number_of_digits(len(modules) - 1))
+
+
 def sort_numbered_paths(*paths):
     """
     Sort input paths to tail number.
--- a/src/haddock/libs/libutil.py
+++ b/src/haddock/libs/libutil.py
@@ -305,7 +305,7 @@ def get_zerofill_for_modules(modules):
 
     This function is used in combination with `zero_fill`.
     """
-    return max(1, get_number_of_digits(len(modules) - 1))
+    return get_number_of_digits(len(modules) - 1)
 
 
 def sort_numbered_paths(*paths):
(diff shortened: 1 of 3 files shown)


joaomcteixeira:
all tests pass... accept it @douweschulte if you don't find any error.
but now I am concerned with https://github.com/haddocking/haddock3/issues/382#issuecomment-1090874872

src/haddock/libs/libutil.py
```diff
@@ -293,6 +293,21 @@ def get_number_of_digits(num):
...
+
+    If there are 5 modules, zerofill digits is 1.
+    If there are 10 modules, zerofill digits is 1 because counting
+    starts at 0 (for topoaa).
+    If there are 100 modules, zerofill digits is 2 because counting
+    starts at 0 (for topoaa).
+
+    This function is used in combination with `zero_fill`.
+    """
+    return max(1, get_number_of_digits(len(modules) - 1))
```
rvhonorato, line 308: Why not simply define everything with 3 digits? its already not realistic that 100 modules will be executed anyway

rvhonorato: would also make it easier to continue-restart. remember that the main guideline is sense and simplicity :P

joaomcteixeira: Yeah. That was the original idea. But @douweschulte yesterday rose a point on that, and he was right. There are drawbacks to having alternating leading zeros. But yesterday night I thought that it might not be that problem even for #383 after all. Anyway, now we need to accept this PR to finalize and correct #378. 

If in the future we decide to remove the alternating leading zeros, it is just a matter of changing a function name, nothing more.

douweschulte, line 308: Nit-picking here, is the `max(1, ..)` needed? Because the `get_number_of_digits` already returns at least one.

douweschulte: I would say it is nice to make sure that it can run in any at all times, even with stupidly high numbers of module, because this would be a very stupid reason to crash the execution of Haddock if everything else is just working. In this implementation if there are 100 modules it starts naming at `000` , `001`, .. . But even if you start a run with 2 modules and so get `0` and `1` it would still be very possible to just continue with `002` if you continue-restart with a run with many more modules.

joaomcteixeira: Yes. But some of these `000` numbers are hardcoded in the modules run files, for which if the name of the folders change, we would need to change also those paths. With `sed` most likely.

joaomcteixeira: But, I realized we need to address this path name changing for some examples, even if we use always two leading zeros. Therefore we can safely use the alternating zerofill; it should not hinder the other implementations.

src/haddock/libs/libutil.py
```diff
@@ -293,6 +293,21 @@ def get_number_of_digits(num):
...
+
+    If there are 5 modules, zerofill digits is 1.
+    If there are 10 modules, zerofill digits is 1 because counting
+    starts at 0 (for topoaa).
+    If there are 100 modules, zerofill digits is 2 because counting
+    starts at 0 (for topoaa).
+
+    This function is used in combination with `zero_fill`.
+    """
+    return get_number_of_digits(len(modules) - 1)
```
joaomcteixeira, line 308: Note: the crucial part in this implementation is the `-1` here.

douweschulte (approved):

PR #391, opened 2022-04-09
Lifts `zero fill` logic to `gear`

joaomcteixeira:

- [ ] Your PR is about CNS

---

A single sentence from @douweschulte the other day already resulted in #378 #383 and this one. :wink:

While implementing #382, I noticed the "zero fill" information managing the folders prefix needs to be stored lib-wide.

This PR lifts the `zero_fill` logic to a new `gear`.

Parts now on `haddock3` depend on the `gear.zero_fill` class state. However, the `zero_fill` gear module has the functions needed to decouple those dependencies in the future if needed.


src/haddock/gear/zerofill.py
```diff
@@ -0,0 +1,110 @@
...
+
+        Zero fill number if available through the attribute: `zfnum`.
+
+        Parameters
+        ----------
+        iterable
+            Usually a dictionary or a list of the modules names.
+            Any object implementing `len`.
+        """
+        self.zfnum = get_zerofill_for_modules(modules)
```
douweschulte, line 57: The called function is one line, is there a reason the define it externally? If there is none I would vote for the definition to be written inline.

src/haddock/gear/zerofill.py
```diff
@@ -0,0 +1,110 @@
...
+        >>> zerofill.zfum
+        1
+
+        >>> zerofill.fill("topoaa", 0)
+        "0_topoaa"
+
+        >>> zerofill.fill("rigidody", 1)
+        "1_rigidbody"
+        """
+        return make_zero_fill(num, self.zfnum) + "_" + name
```
douweschulte, line 74: The same comment is true here for `make_zero_fill`.

douweschulte (approved):

joaomcteixeira:
Yes  @douweschulte 

The reason is that I want to have methods logic outside the class' methods. Why? Because in this way, I can use both FP and OOP depending on the scenario. Is a design decision on purpose. More and more, I use classes just as APIs with just minimal logic inside, and I place all the logic in separated functions outside the class and inside the same module. Is a strategy I started using a while back while developing my project from Toronto. I know few people use it, but so far, is giving me great results.

douweschulte:
That is a valid strategy, I was also thinking of the ease of testing here witch is easier with separate functions. The only major downside is that with python other people could start using the functions instead of the class, which could be bad thing. Otherwise just do it this way.

PR #392, opened 2022-04-09
adds docs

joaomcteixeira:
This Pull Request adds documentation to #382.

@amjjbonvin @rvhonorato let's brainstorm how the HADDOCK3 could edit, continue, and extend a run. Here, I wrote the documentation as if the functionalities were already implemented so we can discuss them properly.

**Questions:**

1. Are the functionalities described here the desired ones? Personally, I like the proposal.
2. If yes, is the documentation clearly explained?
3. If not, what is missing, wrong, or unnecessary?

Let me know your comments. Once we decide on the way to go I will implement it.

p.s. - The first functionalities described should work already. But the last ones require a bit more of coding. That's why I wanted to clarify the requirements before proceeding.

Everyone else is also invented to the discussion.

### DIFF
--- a/docs/continuing_runs.md
+++ b/docs/continuing_runs.md
@@ -1,0 +1,195 @@
+# Continuing and editing runs
+
+The modularity of HADDOCK3 allows users to take parts of a run and use them as
+starting points for new runs using different configuration files. There are
+several ways users can continue or expand previous runs. The index below lists
+those possibilities:
+
+1. Restart a previous run from a given step 1. Extend the original run by
+extending the configuration file with additional modules 1. Take a specific step
+and continue the run onwards, giving a new configuration file
+
+# Restart a previous run from a given step
+
+This option allows you to restart a run from a specific step. It is useful when
+you want to repeat part of the run or to continue a broken run without needing
+to rerun all successful steps.
+
+Let's imagine you execute a run with six steps (modules).
+
+```
+haddock3  my-haddock3-config-file.cfg
+```
+
+If the run completes successfully, the run directory shows as follows:
+
+```
+0_topoaa/
+1_rigidbody/
+2_caprieval/
+3_seletop/
+4_flexref/
+5_caprieval/
+data
+log
+```
+
+You noticed you introduced the wrong values for the `4_flexref` module.
+Therefore, you want to repeat the run only for `4_flexref` onwards without
+repeating the previous successful steps. For that, you should do the following:
+
+```
+haddock3  my-haddock3-config-file.cfg --restart 4
+```
+
+This operation will delete modules' folders `4_` onwards and repeat those steps
+following the information in the configuration file. You must run the `haddock3`
+command from within the same directory as you did for the original run.
+
+# Expanding a previous run from a given step
+
+The `--restart` functionality explained above allows some exciting variations.
+We will see here how to expand a previous run with additional modules.
+
+Run 1 completed successfully, but now we want to complement our run with a
+`mdref` step and perform a final CAPRI score evaluation. For that, add the
+parameters for the new modules to the end of the original configuration file.
+You keep the rest of the configuration file the same.
+
+Navigate back to the same folder you executed the previous runs and:
+
+``` haddock3 expanded-configuration-file.cfg --restart 6 ```
+
+In this case, the `--restart` option takes the value `6` because we want to
+continue the run from step 6, the new `mdref` step.
+
+# Editing and expanding a previous run from a given step
+
+Following the same rationale from the previous example, we can edit and expand a
+previous run. For that, edit the original configuration file by altering
+(add/edit/remove) the information for the new modules you want to execute. For
+example, delete steps `4_` onwards and perform a `mdref` only. You should delete
+the parameters for those modules in the original configuration file and replace
+them by the parameters for a `mdref` execution. In the newly edited
+configuration file, the`mdref` is now the `4_` module, so you should restart
+from four.
+
+```
+haddock3 edit-config-file.cfg --restart 4
+```
+
+This setup would result in:
+
+```
+0_topoaa/
+1_rigidbody/
+2_caprieval/
+3_seletop/
+4_mdref
+data
+log
+```
+
+As you can see, we have reused the first four modules - up to `3_seletop` and
+replaced the `4_flexref` and `5_caprieval` with a `4_mdref` module. You could
+execute as many modules as you wish after `4_mdref`.
+
+# Copy a run to a new folder
+
+Importantly to know, you can execute all previous examples on copies of the
+original run. There is no need to rewrite the initial run if you wish to keep
+it. For that, copy the original run to a new folder:
+
+```
+cp -r run1-test run2-test
+```
+
+In the configuration file, edit the `run_dir` parameter to `run2-test` so it
+matches with the newly created folder. Then, proceed by editing the
+configuration file as discussed in the previous examples. Finally,
+
+```
+haddock3 new-config-file-run2.cfg --restart N
+```
+
+where `N` is the step you wish to start from, as explained previously. Notably,
+you do not need to copy the whole folder when copying the original run. Instead,
+you can copy only the modules folders you wish to maintain plus the `data`
+folder.
+
+# Extend an unknown run
+
+HADDOCK3 allows you to extend a run without previous knowledge of the run
+configuration. For example, you have several runs for which you want to perform
+a CAPRI evaluation to the last step. How to do this?
+
+Create a new configuration file containing just the `caprieval` module, or
+whatever and many modules you desire. For example:
+
+```toml
+[caprieval]
+reference_fname = "reference-structure.pdb"
+```
+
+Now, run the configuration file by giving the `--restart` flag the path of the
+previous run directory:
+
+```
+haddock3 caprieval.cfg --restart <previous-run-directory>
+```
+
+In this case, the configuration file does not need to contain any information
+about the molecules used as input or the previous modules.
+
+# Use a module as a seed for a new run
+
+Imagine you want to take a specific module of a successful run as the starting
+point of a new run. Copy the `data`, the `topoaa`, and the desired module to a
+new folder. You can also use the `topoaa` as the initial module. At this point,
+you can start the new run using both strategies we have detailed so far:
+
+1) edit the original configuration file keeping only the information for the
+modules you used as seed and editing the `run_dir` parameter to the new folder,
+and execute:
+
+```
+haddock3 new-config.cfg --restart 2
+```
+
+If your seed module is directly the `topoaa` module, use `--restart 1`.
+
+2) Create a new configuration file containing only the parameters for the
+modules you want to run and execute:
+
+```
+haddock3 new-config.cfg --restart <run_dir>
+```
+
+# Final considerations
+
+When extending or continuing a run from a new configuration file, all the
+modules' folders and their content will be updated to the corresponding position
+and leading zeros. For example, if the new run takes as initial input from a run
+with more than 10 modules:
+
+```
+00_topoaa/
+04_flexref/
+data/
+```
+
+When you want to extend it with  `caprieval`, `emref`, `caprieval`, and
+`clustfcc`. The final output will be:
+
+```
+0_topoaa/
+1_flexref/
+2_caprieval/
+3_emref/
+4_caprieval/
+5_clustfcc/
+```
+
+Where `0_topoaa` and `1_flexref` are the original folders renamed. The files
+within these folders were also edited to ensure paths compatibility and the
+consistency of the whole run.


docs/continuing_runs.md
```diff
@@ -0,0 +1,204 @@
...
+The modularity of HADDOCK3 allows users to take parts of a run and use them as
+starting points for new runs using different configuration files. There are
+several ways users can continue or expand previous runs. The index below lists
+those possibilities:
+
+1. Restart a previous run from a given step
+1. Expanding a previous run from a given step
+1. Editing and expanding a previous run from a given step
+1. Copy a run to a new folder
+1. Extend an unknown run
```
amjjbonvin, line 12: What is an unknown run???

docs/continuing_runs.md
```diff
@@ -0,0 +1,204 @@
...
+1. Use a module as a seed for a new run
```
amjjbonvin, line 13: Kind of related to copy a run to a new folder - i.e. copy partially a run to a new folder

docs/continuing_runs.md
```diff
@@ -0,0 +1,204 @@
...
+```
+haddock3 new-config-file-run2.cfg --restart N
+```
+
+where `N` is the step you wish to start from, as explained previously. Notably,
+you do not need to copy the whole folder when copying the original run. Instead,
+you can copy only the modules folders you wish to maintain plus the `data`
+folder.
+
+# Extend an unknown run
```
amjjbonvin, line 126: Again unknown in this case is not the proper naming. But this use case should be useful indeed.

Ideally if you want to only run caprieval, you might want to run it for various stages of the run... i.e. insert it between existing steps.

docs/continuing_runs.md
```diff
@@ -0,0 +1,204 @@
...
+1_flexref/
+2_caprieval/
+3_emref/
+4_caprieval/
+5_clustfcc/
+```
+
+Where `0_topoaa` and `1_flexref` are the original folders renamed. The files
+within these folders were also edited to ensure paths compatibility and the
+consistency of paths and the run as a whole.
```
amjjbonvin, line 204: Additional consideration:

Provide an option to copy specific modules from an existing run to a new run. E.g.

I want to take over modules 0 1 4 and 8. Copy those and renumber then sequentially (probably needed)

joaomcteixeira: Okay. That's a good functionality. I will add it.

docs/continuing_runs.md
joaomcteixeira: Therefore, we need the capacity to add intermediate **analysis** stages. Which would require the renaming of the consequent folders. Sounds good?

docs/continuing_runs.md
joaomcteixeira: Yes. I had them separate to avoid overloading each example. So that all the possibilities become clear and are explained in separate steps. But we can change if in the end we see it's best explained differently

docs/continuing_runs.md
amjjbonvin: Yes - but in that case a restart should not delete existing steps, only add the missing ones...

docs/continuing_runs.md
amjjbonvin: So you mean a run for which the config file is not present?  Which will kind of never happen since the config file is copied  in the run dir, right?  You have to have some knowledge of what is in a run in order to be able to setup the next steps.

This seems to me an unlikely scenario

docs/continuing_runs.md
joaomcteixeira: Agree. Likely we would need another flag different from `--restart`.

docs/continuing_runs.md
joaomcteixeira: No, no. The config is surely there. And as users (human beings), we know what the run looks like and all the parameters. But at the moment of running an extension, the command line itself:

`haddock3 new-extension.cfg --restart <folder>`

would not know about the previous run parameters because these are not defined in `new-extension.cfg`, which would automatically be extracted from the run instead. Sometimes I overuse certain words, sorry. Is it clear the difference between:

`haddock3 new-extension.cfg --restart <folder>`

and

`haddock3 new-extension.cfg --restart <number>`

Those are the two option I wanted to implement in these docs.

joaomcteixeira:
> What about then a more explicit option like —restart-from-run

In this case, which configuration file would we give? The whole config or just a partial config with the steps you want to perform?

docs/continuing_runs.md
```diff
@@ -0,0 +1,205 @@
...
+
+1) edit the original configuration file keeping only the information for the
+modules you used as seed and editing the `run_dir` parameter to the new folder,
+and execute:
+
+```
+haddock3 new-config.cfg --restart 2
+```
+
+If your seed module is directly the `topoaa` module, use `--restart 1`.
```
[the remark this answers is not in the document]
joaomcteixeira: Yes, I thought about the `seed` terminology but didn't have a better term at that time. I will look for a better word, indeed.

docs/continuing_runs.md
```diff
@@ -0,0 +1,205 @@
+# Continuing and editing runs
+
+The modularity of HADDOCK3 allows users to take parts of a run and use them as
```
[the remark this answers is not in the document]
joaomcteixeira: Does it sound like that? But yes, I meant what you wrote. I will reword it as you suggested.

docs/continuing_runs.md
```diff
@@ -0,0 +1,205 @@
...
+3_seletop/
+4_flexref/
+5_caprieval/
+data
+log
+```
+
+Now, you noticed you introduced the wrong values for the `4_flexref` module.
+Therefore, you want to repeat the run only for `4_flexref` onwards without
+repeating the previous successful steps. For that, you should do the following:
```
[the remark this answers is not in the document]
joaomcteixeira: Sounds good :+1:

docs/continuing_runs.md
```diff
@@ -0,0 +1,205 @@
...
+```
+haddock3 new-config-file-run2.cfg --restart N
+```
+
+where `N` is the step you wish to start from, as explained previously. Notably,
+you do not need to copy the whole folder when copying the original run. Instead,
+you can copy only the modules folders you wish to maintain plus the `data`
+folder.
+
+# Extend a run with a partial config
```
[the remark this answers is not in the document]
joaomcteixeira: It means you can write a config just with the information of the new modules you want to add to the run and execute those additions. I will try to rephrase

joaomcteixeira:
Thanks for your feedback @rvhonorato 

Maybe on some parts, I went too straightforward on the text. I didn't want the text to be wordy. But surely I can explain those better.

docs/continuing_runs.md
```diff
@@ -0,0 +1,205 @@
...
+
+```
+haddock3 caprieval.cfg --restart <previous-run-directory>
+```
+
+In this case, the configuration file does not need to contain any information
+about the molecules used as input or the previous modules. HADDOCK3 will just
+execute the new modules on top of the previous run.
+
+# Use a module as a seed for a new run
```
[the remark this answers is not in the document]
joaomcteixeira: I hope this is solved in aa8bcce30bdd264d5bba8136c0181dad64b83185

joaomcteixeira:
@rvhonorato @amjjbonvin 

I have addressed Rodrigo's comments and made extensive changes to the document by simplifying several explanations while adding more detailed examples. Can you have a quick review?

Later, I will address @amjjbonvin comments.

docs/continuing_runs.md
```diff
@@ -0,0 +1,225 @@
...
+haddock3 new-config-file-run2.cfg --restart N
+```
+
+where `N` is the step you wish to start from, as explained previously.
+
+Notably, you do not need to copy the whole folder when copying the
+original run. Instead, you can copy only the steps' folders you wish to
+maintain plus the `data` folder. For example, if the original run has
+ten modules, but you wish to keep only the first three, you can copy
+only those first three modules to the new run folder.
```
mgiulini, line 132: In my opinion this sentence is a bit convoluted.
```suggestion
There is no need to keep the whole original run when copying, 
but only those folders you wish to maintain plus the `data` directory.
```

docs/continuing_runs.md
```diff
@@ -0,0 +1,225 @@
...
+information about the molecules used as input or the previous modules.
+Instead, HADDOCK3 will execute the new modules following the run's last
+module.
+
+# Final considerations
+
+When extending or continuing a run from a new configuration file, all the
+modules' folders and their content will be updated to the corresponding position
+and leading zeros. For example, the new takes steps 0 and 4 from a run with more
+than 10 modules (note the leading double digit numbers):
```
mgiulini, line 203: This is a bit verbose and unclear to me..Could you rephrase it?

docs/continuing_runs.md
```diff
@@ -0,0 +1,225 @@
...
+configuration file according to your new preferences, and run:
+
+```
+haddock3 my-haddock3-config-file.cfg --restart 4
+```
+
+This operation will delete modules' folders `4_` onwards and repeat
+those steps following the configurations in the (new) configuration
+file. You must run the `haddock3` command from within the same directory
+as you did for the original run.
```
mgiulini, line 53:
```suggestion
in which you launched the original run.
```

mgiulini (commented):
Added some comments and suggestions. In general I think it's a bit too verbose, but that's my personal taste :)

docs/continuing_runs.md
```diff
@@ -0,0 +1,225 @@
...
+
+As you can see, this option allows you to **restart** or **repeat** a
+run from a specific step using the **original** configuration file or to
+**modify** a run by updating the parameters for the steps you want to
+rerun.
```
rvhonorato, line 58: This is still confusing, if you use the original you restart, if you use the modified you repeat? What is the difference between restart and repeat or here is just a synonymy?

docs/continuing_runs.md
```diff
@@ -0,0 +1,225 @@
...
+
+Likewise, we can use the `--restart` functionality to **extend** a run
```
rvhonorato, line 60: So its, `restart`, `repeat` and `extend` all with the same option?

docs/continuing_runs.md
```diff
@@ -0,0 +1,225 @@
...
+```
+cp -r run1 run2
+```
+
+You can also copy the run to different folders because HADDOCK3 treats
+internal paths as relative paths. Therefore, modules can communicate
+with each other regardless of the absolute location of the run
+directory. The same applies to the installation requirements. Provided
+the installation process was the same, you can copy HADDOCK3 run over
+different machines.
```
rvhonorato, line 110: Are you sure about this one? I've tried it once with a previous version - maybe its fixed now, from my workstation to the cluster and it did not work.

rvhonorato (approved):
Still very confusing to me, but I guess some documentation is better than none

docs/continuing_runs.md
joaomcteixeira: It should work now, provided the haddock3 environment is installed. all haddock3 libraries are taken from the system installation. The run does not care about where things are installed, as long as they are accessible in the system, which they will be if installed following the INSTALL.md instructions.

docs/continuing_runs.md
```diff
@@ -0,0 +1,225 @@
...
+internal paths as relative paths. Therefore, modules can communicate
+with each other regardless of the absolute location of the run
+directory. The same applies to the installation requirements. Provided
+the installation process was the same, you can copy HADDOCK3 run over
+different machines. 
+
+In the configuration file, edit the `run_dir` parameter to `run2` to
+match the newly created folder. `run1` and `run2` are arbitrary names.
+
+**Important:** The `run_dir` path must point to the new run directoy
```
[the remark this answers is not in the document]
joaomcteixeira: Otherwise, how does haddock knows where the run dir is?

joaomcteixeira:
I find it surprising that the text is still verbose and unclear for you. But likely it is. I will rephrase it completely, then. The aim is to get clear docs and functionalities. Maybe we can find a way to make the functionality easier also.

joaomcteixeira:
@amjjbonvin @rvhonorato @mgiulini, thanks for all your comments!

I completely rewrote the documentation page and simplified it a lot. Let me know your comments. The whole text is different. I hope now is clear and straightforward.

docs/continuing_runs.md
```diff
@@ -0,0 +1,100 @@
...
+
+You can profit from the `--restart` option described above to (re)run modules
+with modified parameters. For example, if you noticed a parameter was set to
+`true` when you wanted it to `false`, or if you wish to increase the sampling
+for a given refinement module. Simply edit the configuration file modifying the
+parameters for the modules you want to (re)run, in this case, from the fourth
+module onwards.
+
+```
+haddock3 my-run-config.cfg --restart 3
```
mgiulini, line 36: Here the command is equal to the previous one, although the concept is quite different. I would emphasize the fact that the config file has been modified (as you do later)
```suggestion
haddock3 my-edited-config.cfg --restart 3
```

docs/continuing_runs.md
```diff
@@ -0,0 +1,100 @@
+# Editing and extending runs
+
+The modularity of HADDOCK3 allows restarting halted runs from a certain step,
+extending previous runs with additional steps, and using successful steps as
+starting points for new runs. Let's see some examples.
+
+## Restarting a run
+
+Use the'- restart' flag to restart a HADDOCK3 run. For example, imagine you
```
mgiulini, line 9:
```suggestion
Use the'--restart' flag to restart a HADDOCK3 run. For example, imagine you
```

docs/continuing_runs.md
```diff
@@ -0,0 +1,100 @@
...
+
+Create a configuration file for the new run. **Note:** In this case, you
+**don't** need to define the `run_dir` and `molecules` parameters because these
+will be taken from the initial step folder. You should not define the parameters
+for the module you took as the starting point.
+
+To start the new run:
+
+```
+haddock3 my-new-config.cfg --restart-from-run <path-to-run-dir>
```
mgiulini, line 71: should we be overprecise and specify what is `<path-to-run-dir>`?

docs/continuing_runs.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+## Additional considerations
+
+You can now combine the functionalities explained above in many different
+variations.
+
+Below is some quick Q&A regarding these options:
```
mgiulini, line 79: too colloquial in my opinion, but it's subjective ;)

mgiulini (approved):
Great job! I find it much more readable and clear. I just added some super minor comments.

docs/continuing_runs.md
rvhonorato: think so too

rvhonorato (approved):

docs/continuing_runs.md
```diff
@@ -0,0 +1,100 @@
...
+configuration file).  You want now to continue the run from where it stopped.
+Use the `--restart` followed by `N`, where `N` is an integer corresponding to
+the step you which to restart from. **Remember**, haddock steps are 0-indexed
+(0, 1, 2, 3, ...); so, `--restart 3` is the fourth step.
+
+```
+haddock3 my-run-config.cfg --restart 3
+```
+
+## Restarting a run with modified parameters
```
amjjbonvin, line 26: I would add here a comment that this will cause all steps >= N to be deleted from the current run

docs/continuing_runs.md
amjjbonvin: May-be path-to-old-run-dir instead?

docs/continuing_runs.md
```diff
@@ -0,0 +1,100 @@
...
+
+```
+haddock3 my-edited-config.cfg --restart 5
+```
+
+We use `--restart 5` because the first new step is the sixth in the new
+workflow: the five initial steps and three new steps. Remember `--restart` is
+0-indexed.
+
+## Starting new runs from previous modules
```
amjjbonvin, line 55: Unclear to me what will happen. In this example you copy 0_topoaa and 4_flexref in a new run directory (what should be the name of it and how will haddock know its name?  I.e. somewhere you need to specify `run_dir`...

Also what should the new config file contain? Only the additional steps? E.g. I want to run `emref`, should the config file only contain that?  And what will be the numbering of the steps in the new run?  0_topoaa, 4_flexref and 5_emref, or are the steps renumbered sequentially from 0?

docs/continuing_runs.md
joaomcteixeira: Why? How would you rephrase it?

docs/continuing_runs.md
joaomcteixeira: Addressed

joaomcteixeira:
> May-be path-to-old-run-dir instead?

No. The path is the **new** run directory that was manually created. I will update the option to `--restart-from-dir`. Updated the text also.

amjjbonvin (approved):

mgiulini (approved):

joaomcteixeira:
Perfect. Thanks for all the feedback!

Now that we have defined the requirements, I will prepare the code implementing all the functionalities here described.

I will keep this PR open. Once the code is ready, we will merge both PRs together.

mgiulini (approved):

joaomcteixeira:
While developing #425 (`--restart-from-dir`) I realized that preparing the new run by manually copying folders may create some uncontrolled scenarios because we need to copy the `data/` folder and we lose track of the original run directory. @amjjbonvin had suggested at some point to use a CLI to handle these cases. What about:

```
haddock-copy -i <original_run_dir> -m 0 4 -o <new_run_dir>
```

Where `-m` are the number of models to copy. What do you think?

joaomcteixeira:
By developing #425, I realized the strategy described here was not "perfect". I updated the `--restart-from-dir` part. Now it is `--start-from-copy` and uses the `haddock3-copy` commando also. Much cleaner and easy to handle in code. Have a look, please.

joaomcteixeira:
Perfect. Merging this one before merging #425 :rocket:

amjjbonvin:
Where -m are the stages to copy rather than the model?

amjjbonvin:
Sense and simplicity!

Keep things simple, short sentences to the point and not too much verbosity

PR #396, opened 2022-04-12
Updates manual integration tests with documentation

joaomcteixeira:

---

Because more of us are now participating in the development of haddock3, I push my scripts to run the integration tests manually. This is how I am doing, but I think they can become part of the repository so that everyone has the guidelines and the tools to use them. Please read the documentation to see if it is clear.

Added more `-test.cfg` examples for antibody-antigen cases.

Added `ncores=40` to all tests. Anyway, haddock internally will decide which is the best `ncores` value.

Added a `run-test.job` to send tests to slurm.

Even when we automate theses tests, having the capacity to run them locally is always good.


mgiulini (commented):
Very nice @joaomcteixeira. I just added two minor comments.

docs/integration_tests.md
```diff
@@ -0,0 +1,95 @@
...
+```
+some-folder/
+    - haddock3/
+    - haddock3main/
+```
+
+## 2. Install HADDOCK3 main branch in a separate folder and python environment
+
+```bash
+git clone https://github.com/haddocking/haddock3 haddock3main
```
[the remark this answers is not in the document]
joaomcteixeira: Totally right!

joaomcteixeira:
Done it @mgiulini :+1: thanks for your feedback

mgiulini (approved):

docs/integration_tests.md
```diff
@@ -0,0 +1,103 @@
+# Running integration tests manually
+
+The integration tests ensure HADDOCK3 workflows work properly by
+performing short docking simulations on benchmark systems. With these
```
rvhonorato, line 4: Not everything is docking and these are not benchmark systems

examples/diff_capri.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+# execute this script from inside the `examples/` folder.
+
+# HADDOCK main relative path
+HM="../../haddock3main/examples"
+
+diff_capri () {
```
rvhonorato, line 8: Why is this being done with shell and not with python? Can't we parse these and compare if the values are within a tolerable range? How can the developer tell if the numbers "match"?

examples/run-tests.job
```diff
@@ -0,0 +1,16 @@
...
+
+# make sure anaconda is activated
+source $HOME/miniconda3/bin/activate
+conda activate haddock3
+
+# go to the example directory
+cd $HOME/haddock3/examples
+
+# execute
+python run_tests.py -b
```
rvhonorato, line 16: I'd change this to be a template instead, this is very particular to our cluster - other's will not have a `short` queue. Best to avoid people just blindly trying to run things.

rvhonorato (changes requested):

examples/diff_capri.sh
joaomcteixeira: The quick answer is that I just wanted to share my scripts with the main dev team. Since we are not yet on an official release. I can make it on Python, but for testing, is that needed?

Regarding comparing the values within a tolerable range. Do we want a tolerable range for integration tests? Remember tests are run on the same machines and the numbers should be equal.

The developer knows the numbers match because `diff` gives no output. Otherwise `diff` shows the different lines.

examples/run-tests.job
joaomcteixeira: Yeah. Okay. I wrote in the docs the developer should edit it. But I can make it a template. But what parts should I transform the template? Should I remove the `queue` line?

examples/diff_capri.sh
rvhonorato: > I can make it on Python, but for testing, is that needed?

"The project guidelines" say haddock3 is python exclusive (: 

> Remember tests are run on the same machines and the numbers should be equal.

So maybe I did not understand what this part is doing, is it not comparing values obtained by the tests with an expected one?

examples/run-tests.job
rvhonorato: Is this is just for internal usage I'd actually remove it. People running HPC jobs are advanced users and can write simple scripts

examples/diff_capri.sh
joaomcteixeira: Okay, better make it in python. It might be useful for the future.

> So maybe I did not understand what this part is doing, is it not comparing values obtained by the tests with an expected one?

No, is just using the bash `diff` command. If files are equal, no output. If numbers are different, files are different, `diff` outputs the difference. Just that.

examples/run-tests.job
joaomcteixeira: Yes. It is as stated in the documentation. But, yes. Maybe is better to remove that mention and the `run-test.job`.

docs/integration_tests.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+If you are on a Slurm supported system, you can use the `run-tests.job`
+file.  You might need to edit the file according to your system
+specifications and to update paths and environment names.
+
+## 5. Comparing both runs
+
+Once both runs completed, that is, all `*-test.cfg` files were executed,
+you can compare the different results for the CAPRI scoring modules.
```
rvhonorato, line 89: Sorry, now I understand: you make a branch, make modifications, compare yours with the main and check the diff, ok.

But remember that due to the non-deterministic nature of haddock, changes in the CNS level might produce models with slight variations and these will be propagated to the capri metrics (ex 0.147 instead of 0.152). This is what I meant about values being within a certain range.

rvhonorato:
Also, since you are at it, is it not better if we move all the test-related scripts to the `tests/` folder instead? I think its a bit odd that we have examples and test together

docs/integration_tests.md
joaomcteixeira: Yes. I actually suggest keeping two separate haddock3 installations: one for development and the other for the `main`, just for this testing purposes. That's how I am doing.

Regarding the numbers, when running on the same machine, there should be not these different numbers, I think. So far, numbers have been always the same in all the PR tests I have made in the last months.

Obviously, if the changes affect the CAPRI scores on purpose, we cannot use these as validity metrics.

I will have a look further at your comments and edit accordingly.

joaomcteixeira:
> Also, since you are at it, is it not better if we move all the test-related scripts to the tests/ folder instead? I think its a bit odd that we have examples and tests together.

To what I explained in the ISGC meeting, and I am writing it now in the form of text, I would say it is okay to have tests in the `examples/`. The `tests/` folder are for unnittests or API tests. And in this case the `examples/*/*-test.cfg` are integration tests. I would keep them separated.

The `-test.cfg` in `examples/` also serves to demonstrate to users before running the actual `-full.cfg`. We had discussed this before, as of today, I still agree with this rationale. However, your comment is important. :+1:

docs/integration_tests.md
rvhonorato: > Regarding the numbers, when running on the same machine, there should be not these different numbers, I think. 

You are wrong. I'm speaking from experience, but fine - we can cross that bridge when the times comes to it.

joaomcteixeira: Even for 3 decimal numbers? Okay. I can make it to test `close` numbers, that's not a problem. I was planning to make in Python as you suggested.

rvhonorato: As I said already, its not numerical fluctuation, its a result of a **slightly different model**, but nevermind that, let's not worry with future problems hehe

joaomcteixeira: Is not a future problem, is understanding haddock's behavior. So far, I see deterministic results: that is, running the examples several times always gives the exact same numbers. I am taking note of your comments :+1:

docs/integration_tests.md
joaomcteixeira: for docking :+1:
Okay, I can reword the "benchmark". But a *benchmark* is anything used as a reference. These are benchmarks because they are used as test cases. But I will reword it because HADDOCK uses "benchmark" for a very specific purpose. I think you were referring to that. :+1:

amjjbonvin:
In this context benchmark should be replaced by test systems 

The goal is not to get the best results but to ensure the workflow works properly

joaomcteixeira:
> In this context benchmark should be replaced by test systems. The goal is not to get the best results but to ensure the workflow works properly

Understood, done it :+1:

joaomcteixeira:
Okay, ready for review:

- Made the comparison script in Python
- Update the docs. Are they clear?
- Added three more `*-test.cfg` cases for `antibody-antigen`
- added tests
- removed the `.job` file. Agree with @rvhonorato; advanced users should know what to do.

docs/integration_tests.md
```diff
@@ -0,0 +1,113 @@
...
+
+```bash
+python run_tests.py -b
+```
+
+This script will run all the test cases in the `examples/` directory and
+stop if any of the tests break. For help, see `python
+run_tests.py -h`.
+
+## 4. Running the tests on your branch
```
amjjbonvin, line 82: Is this the usual way to do it?  I am creating branches in my main installation and switching between branches.

examples/docking-antibody-antigen/docking-antibody-antigen-CDR-accessible-clt-test.cfg
```diff
@@ -0,0 +1,95 @@
+# ====================================================================
+# Protein-protein docking example with NMR-derived ambiguous interaction restraints
```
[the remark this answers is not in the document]
joaomcteixeira: @mgiulini implementation was not merged yet -> #397
Whoever gets merged first, the other needs to update :wink:

docs/integration_tests.md
joaomcteixeira: That's also possible. If you have two separate installations, you can run the tests simultaneously: main and developing. I thought about it, but I preferred to document the two installations because I think it is easier than asking to copy the `examples/` folder to a new one and be switching between branches. But both ways work.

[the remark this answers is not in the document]
joaomcteixeira: I use slurm batch submission to run the `*-test.cfg`, and I have the `job` files for that in my account. Initially, I had an explanation for using Slurm, but removed it because I agreed with @rvhonorato comments (https://github.com/haddocking/haddock3/pull/396#discussion_r849123487). But I can add a comment for that now. Should I?

rvhonorato (approved):

docs/integration_tests.md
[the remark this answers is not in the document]
joaomcteixeira: Added short message in the docs

joaomcteixeira:
> "- if I just want to test if proteinA interacts with proteinB, should I use docking-protein-protein-mdref-full.cfg or docking-protein-protein-mdref-test.cfg?"

@rvhonorato 

Added option to `compare_runs.py` to compare two specific run directories.

docs/integration_tests.md
```diff
@@ -0,0 +1,130 @@
...
+working on incorporating these tests into our automatic CI pipelines.
+
+Running the integration tests for a single branch takes about 1 hour using 8 CPU
+cores in a AMD EPYC 7451 24-Core Processor. In case you do not have the computer
+power to run these tests, ask us and we will run them for you.
+
+**How to correctly set up and run the integration tests?**
+
+The easiest way is to have two installations of haddock3:
+one fixed at the `main` branch and other with your developments.
```
mgiulini, line 24:
```suggestion
one fixed at the `main` branch and another with your developments.
```

docs/integration_tests.md
```diff
@@ -0,0 +1,130 @@
...
+run_tests.py -h`.
+
+## 4. Running the tests on your branch
+
+Navigate to your development repository and haddock3 environment - you
+are likely on a different terminal window.
+
+Run the tests the same way:
+
+Navigate to the `examples/` folder and execute the tests:
```
mgiulini, line 89: I would lump these two sentences in a single one

examples/compare_runs.py
```diff
@@ -0,0 +1,347 @@
...
+    ("docking-protein-ligand"      , "docking-protein-ligand-test.cfg"),  # noqa: E203, E501
+    ("docking-protein-peptide"     , "docking-protein-peptide-test.cfg"),  # noqa: E203, E501
+    ("docking-protein-peptide"     , "docking-protein-peptide-mdref-test.cfg"),  # noqa: E203, E501
+    ("docking-protein-protein"     , "docking-protein-protein-test.cfg"),  # noqa: E203, E501
+    ("docking-protein-protein"     , "docking-protein-protein-cltsel-test.cfg"),  # noqa: E203, E501
+    ("docking-protein-protein"     , "docking-protein-protein-mdref-test.cfg"),  # noqa: E203, E501
+    ("refine-complex"              , "refine-complex-test.cfg"),  # noqa: E203, E501
+    ("scoring"                     , "emscoring-test.cfg"),  # noqa: E203, E501
+    ("scoring"                     , "mdscoring-test.cfg"),  # noqa: E203, E501
+    )
```
mgiulini, line 80: I am a bit skeptical about hard-coding all these examples here, as we have them hard-coded in another script (`run_tests.py`). Wouldn't be better to create a list and import it in both cases? so that when we add a an example we only need to modify the file with the list

joaomcteixeira: Yes, you are right. These were hard coded at first when we were prototyping them. But now it makes little sense. I will arrange that in another pull request.

joaomcteixeira:
Yes, updated with all your comments. I think it is fine now. `compare_runs.py` can be used to compare `examples` folders as a whole or individual run directories.

rvhonorato (approved):

PR #397, opened 2022-04-12
added input output dictionary in scoring modules

### LINKED ISSUE #394: run scoring modules after modules other than topoaa
Currently, it is impossible to run scoring modules after modules other than topoaa. 

The following error appears:
```
File "haddock3/src/haddock/modules/scoring/mdscoring/__init__.py", line 93, in _run
    original_name = pdb.topology.file_name.replace('.psf', '.pdb')
AttributeError: 'list' object has no attribute 'file_name'
[2022-04-11 17:48:34,152 libutil ERROR] 'list' object has no attribute 'file_name'
```

mgiulini:

- [ ] Your PR is about CNS

---

now a dictionary is used to keep track of the original pdb name in emscoring/mdscoring. This allows to run these modules anywhere in the workflows.

***

(edit by @rvhonorato for clarity) This PR fixes #394 by taking the original filename from the `.pdb` input file instead of the `.psf` generated by the `topoaa` module


joaomcteixeira:
Great!

Do you think you can add some examples in the `examples/` folder? One or two should suffice. Thanks so much! Great correction :rocket:

mgiulini:
> Great!
> 
> Do you think you can add some examples in the `examples/` folder? One or two should suffice. Thanks so much! Great correction 🚀

I am not sure it makes sense in this context..we would like the users to be able to apply scoring modules whenever they like, but I would not encourage them to do so by creating examples. what do you think?

joaomcteixeira:
I meant to add these modules to some of the `-test.cfg` in the `examples/` folder so that we add them to the integration tests.

mgiulini:
They are already there! `examples/scoring/emscoring-test.cfg` and `examples/scoring/mdscoring-test.cfg` are executed when launching `run_tests.py` :)

amjjbonvin:
You could simply modify the antibody example to use emscoring

joaomcteixeira:
> They are already there! examples/scoring/emscoring-test.cfg and examples/scoring/mdscoring-test.cfg are executed when launching run_tests.py :)

yes, but they run after `topoaa` :wink: As a reviewer, how do I know your code works, and why is #394 important in the first place? (playing :smiling_imp: side here :wink:)

mgiulini:
yess, understood, I will do that!

mgiulini:
I added the two tests as requested. It is important to underline that it is currently impossible to run `caprieval` after `mdscoring`, as the latter produces a pdb file with some water inside

src/haddock/modules/scoring/emscoring/__init__.py
```diff
@@ -56,6 +57,9 @@ def _run(self):
             expected_pdb = prepare_expected_pdb(
                 model, model_num, ".", "emscoring"
                 )
+            # fill the input.pdb - output.pdb dictionary
+            io_table[expected_pdb.file_name] = model.file_name
```
rvhonorato, line 61: The `PDBFile` class has `ori_name`, its best to use it instead:

```suggestion
            expected_pdb.ori_name = model.file_name
```

then you don't need the `io_table` dictionary

src/haddock/modules/scoring/emscoring/__init__.py
```diff
@@ -87,10 +91,8 @@ def _run(self):
...
             fh.write(f"structure\toriginal_name\tscore{linesep}")
             for pdb in self.output_models:
-                # temporary solution to get the original name
-                #  here one .pdb = one .psf and the .psf is not changed by
-                #  the module so use its name as the original one
-                original_name = pdb.topology.file_name.replace('.psf', '.pdb')
+                # retrieve the original name
+                original_name = io_table[pdb.file_name]
                 fh.write(
                     f"{pdb.file_name}\t{original_name}\t{pdb.score}{linesep}"
```
rvhonorato, line 97: using `PDBFile.ori_name` instead:

```suggestion
                fh.write(
                    f"{pdb.file_name}\t{pdb.ori_name}\t{pdb.score}{linesep}"
```

rvhonorato (changes requested):
With this PR instead of getting the original name from the topology, we recover it from the input file name, good addition!

See my comments below, you could use `PDBFile.ori_name` (that was added exactly for this purpose) instead of a dictionary.

Also I see you added two new examples, one using `emscoring` and other using `mdscoring`, since the difference between these two modules is at the CNS level and this PR deals with the python shell, I'm not sure these two examples are necessary.

src/haddock/modules/scoring/emscoring/__init__.py
mgiulini: perfect!

joaomcteixeira:
Question. Is it possible to have `emscoring` following `mdscoring` in a hd3 pipeline? In other words, having multiple consecutive scoring and analysis modules? Something to implement if not. Interesting how can one can predict, or not, the behavior (the good one) of code.

mgiulini:
> With this PR instead of getting the original name from the topology, we recover it from the input file name, good addition!
> 
> See my comments below, you could use `PDBFile.ori_name` (that was added exactly for this purpose) instead of a dictionary.
> 
> Also I see you added two new examples, one using `emscoring` and other using `mdscoring`, since the difference between these two modules is at the CNS level and this PR deals with the python shell, I'm not sure these two examples are necessary.

I actually agree, the examples do not add much..but, as @joaomcteixeira said, that part of the code should be somehow tested..should I create a `test_module_scoring.py` inside the `tests` directory?

mgiulini:
> Question. Is it possible to have `emscoring` following `mdscoring` in a hd3 pipeline? In other words, having multiple consecutive scoring and analysis modules? Something to implement if not. Interesting how can one can predict, or not, the behavior (the good one) of code.

It is already possible to run `mdscoring` after `emscoring` and the other way around.

rvhonorato:
From my point of view integration tests should be a way to check if a given **useful and unique workflow** is being properly executed.

The docking of antibody and antigen considering with random restraints for CDRs is the workflow, if you add a module to this, does it become a new workflow or just a variation? 

IMHO the integration tests should be as complex as possible (to cover the wider range of features) not variations of the same. So in this specific case, I'd change `docking-antibody-antigen-ranairCDR-test.cfg` to use `mdscoring` (and later even `caprieval`) and add specific tests for the modules; `test_module_emscoring` and `test_module_mdscoring`.

joaomcteixeira:
I agree with @rvhonorato. If you can add `mdscoring` and `emscoring` to some of the already existing `*-test.cfg`, that's okay. But we need to have some kind of testing in this PR :wink:

mgiulini:
> > I actually agree, the examples do not add much..but, as @joaomcteixeira said, that part of the code should be somehow tested..should I create a test_module_scoring.py inside the tests directory?
> 
> From my point of view integration tests should be a way to check if a given **useful and unique workflow** is being properly executed.
> 
> The docking of antibody and antigen considering with random restraints for CDRs is the workflow, if you add a module to this, does it become a new workflow or just a variation?
> 
> IMHO the integration tests should be as complex as possible (to cover the wider range of features) not variations of the same. So in this specific case, I'd change `docking-antibody-antigen-ranairCDR-test.cfg` to use `mdscoring` (and later even `caprieval`) and add specific tests for the modules; `test_module_emscoring` and `test_module_mdscoring`.

Sounds reasonable. At this point I would also include `emscoring` in `docking-antibody-antigen-ranairCDR-test.cfg`, so that the workflow becomes as comprehensive as possible. Will implement the tests asap.

mgiulini:
@joaomcteixeira @rvhonorato I tried to address your comments and suggestions. Briefly:

- I deleted the unnecessary cfg files in the examples;
- I added an example in the `examples/scoring` directory that ensures that it is always possible to run `mdscoring` after `emscoring`
- I created two files for testing the scoring modules, namely `test_module_emscoring.py` and `test_module_mdscoring.py`. They both instantiate the `HaddockModule` and test the output function. I see that such a procedure has never been implemented, let me know if you deem it reasonable and portable;
- inside the tests folder I created the `params` folder, which is supposed to contain the `.yaml` file necessary to test other modules in a similar way (yes caprieval, this is for you)

src/haddock/modules/scoring/mdscoring/__init__.py
```diff
@@ -84,15 +99,6 @@ def _run(self):
...
-            for pdb in self.output_models:
-                # temporary solution to get the original name
-                #  here one .pdb = one .psf and the .psf is not changed by
-                #  the module so use its name as the original one
-                original_name = pdb.topology.file_name.replace('.psf', '.pdb')
-                fh.write(
-                    f"{pdb.file_name}\t{original_name}\t{pdb.score}{linesep}"
-                    )
-
+        self.output(self.output_models, output_fname)
```
joaomcteixeira, line 102: If `self.output` is a method of the module and `self.output_models` is defined, better to have `output` as a normal method and internally use the `self.output_models`.
If you want to keep it in a separate function, better have a single function inside the `scoring/` folder that is used by both modules. I see `self.output` is the same for both `emscoring  and `mdscoring`. I can handle it if you want.

tests/params/default_emscoring.yaml
```diff
@@ -0,0 +1,297 @@
+#
```
joaomcteixeira, line 1: Are these the same as the `defaults.yaml`? If so, there is no need to repeat the text here. You can import the defaults directly:

```python
from haddock.modules.scoring.emscoring import DEFAULT_CONFIG as emscoring_default_params
from haddock.gear.yaml2cfg import read_from_yaml_config
```

joaomcteixeira (changes requested):

joaomcteixeira:
Well done, Marco. Made just some comments.

rvhonorato (approved):

src/haddock/modules/scoring/mdscoring/__init__.py
mgiulini: agreed and implemented! I went for the first option, as probably there will not be other functions shared by both the scoring modules. If those will be implemented, the second option is definitely better :)

joaomcteixeira (approved):
@mgiulini, I have made the corrections I found pertinent. I agree to merge this PR. :rocket: well done!

PR #409, opened 2022-04-15
Modules output `params.cfg`

### LINKED ISSUE #407: Writing down a copy of the module's parameters
I know we want to write to the disk the less possible files. But I think it is useful that each module writes a file with all the parameters it used, that is, the defaults updated with the user parameters.

This will facilitate reporting and quick traceability. Maybe we can write this information somewhere inside another already used file, `io.json`?

What are your thoughts?

joaomcteixeira:

- [ ] Your PR is about CNS

---

Closes #407

Creates a `params.cfg` file for each step folder. The file is created just after the folder is created.

**notes:**
Needed to add `toml` as a **testing**  dependency; it won't get installed by installing haddock, is just for testing.


joaomcteixeira:
Meanwhile integration tests run, you can have a look :wink:

joaomcteixeira:
@sverhoeven 

I think this PR will be useful for the workflow builder. You can now save dictionaries to haddock3 user config files.

```python
from haddock.modules import save_config
save_config(params_dict, fpath)
```

If the dictionary is a nested dictionary where the main keys are the names of the modules, it will save a user configuration file as expected. If the main keys are directly the parameters of a specific module, you need to give the module name, for example:

```python
save_config(params_dict, fpath, module_name="topoaa")
```

otherwise it will save the config flat without the `[topoaa]` header, for example.

rvhonorato (approved):
I don't agree with this implementation because I cannot see why/how this extra file would be needed, specially since everything is already in the `run.cfg` and since it would add unnecessary complexity and increase the learning curve of basic users trying to replicate their simulations - instead of having just one .cfg file now they have many and need to learn which should be shared/published together.

But the implementation is ok 🙂

joaomcteixeira:
We can turn **off** saving the `params.cfg` for each module file during the haddock run just by commenting a line a in `libs.workflow` if we wish in the future. Nonetheless, I think it is important because despite having the run config, the defaults may very from version to version, or there might be some error. This file is a way to verify everything is okay. Users can just ignore this file.

The implementation to save a dictionary to a haddock user config is needed, though.

imho

amjjbonvin (approved):
I can see it can be useful indeed in case we change some defaults in module cfg files. Would allow finding out the differences.

joaomcteixeira:
Tests are all passing and are easy to compare after #397 :relaxed:

sverhoeven:
> @sverhoeven
> 
> I think this PR will be useful for the workflow builder. You can now save dictionaries to haddock3 user config files.
> 
> ```python
> from haddock.modules import save_config
> save_config(params_dict, fpath)
> ```
> 
> If the dictionary is a nested dictionary where the main keys are the names of the modules, it will save a user configuration file as expected. If the main keys are directly the parameters of a specific module, you need to give the module name, for example:
> 
> ```python
> save_config(params_dict, fpath, module_name="topoaa")
> ```
> 
> otherwise it will save the config flat without the `[topoaa]` header, for example.

Nice, I need to read/write the config with TypeScript so good to have Python code I can compare functionality with. 

Does the function also write `[topoaa.mol1]` header?

joaomcteixeira:
> Does the function also write [topoaa.mol1] header?

Yes. Note that for that the dict needs to be `{"topoaa": {params...}}`. If the dict is just `{params...}` you will have `[mol1]`. Still, if it becomes too complicated to rewrite such a function in Typescript, let me know. We can think about something "in between". The code you need is in `gear.config_writer.py`. Some recursive functions and type checking.

joaomcteixeira:
Yes. any parameter that is a list or tuple, say `mylist = [1, 2, 3, 4]` will be saved in the cfg text files as:

```toml
mylist = [
    1,
    2,
    3,
    4
    ]

# the empty line above is mandatory
```

Does this answer your question? :relaxed:

amjjbonvin:
Doesn’t seem to apply to the list of input pdb files from the provided examples… referring here to the mandatory empty first line

sverhoeven:
Yes it does, it did some additional testing

```python
from haddock.gear.config_reader import read_config
from haddock.modules import save_config
c = read_config("./examples/docking-protein-protein/docking-protein-protein-test.cfg")

save_config({'topoaa': c['topoaa']}, '/tmp/topoaa.cfg', module_name="topoaa")

c_without_modules = dict(filter(lambda kv: kv[0] in {"run_dir", "ncores", "mode", "molecules"}, c.items()))
save_config(c_without_modules, '/tmp/general.cfg')
```
Both config file look like I expected.

topoaa..cfg
```ini
[topoaa]
[topoaa]
autohis = false

[topoaa.mol1]
nhisd = 0
nhise = 1
hise_1 = 75

[topoaa.mol2]
nhisd = 1
hisd_1 = 76
nhise = 1
hise_1 = 15
```
general.cfg 
```ini
run_dir = "/home/verhoes/git/vre4imbc/haddock3/run1-test"
ncores = 40
mode = "local"
molecules = [
    "data/e2aP_1F3G.pdb",
    "data/hpr_ensemble.pdb"
    ]
```

joaomcteixeira:
I am away replying from my mobile. Sorry for the quick text. I see two
`topoaa` headers. There should be only one. Cheers

sverhoeven:
The double topoaa header disappears when I do
```python
save_config(c["topoaa"], "/tmp/topoaa.cfg", module_name="topoaa")
```

joaomcteixeira:
Perfect, that's the behavior. Otherwise, if `c` contains only the `topoaa` information and the `topoaa` main key, you can give:

```python
save_config(c, "/tmp/topoaa.cfg")
```

`save_config` should use the main key to write the header.

PR #423, opened 2022-05-03
Reduce `pytest` verbosity

joaomcteixeira:
As the number of unit tests increases, the current `pytest` command in `tox` is becoming very verbose. This PR reduces the verbosity of `tox -e py39` command. If you want fully verbosity, run `tox -e py39 -- -vv`. Also updated the docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,7 +38,9 @@ To contribute to the HADDOCK3's Python shell, follow these steps:
     (should) use our `tox` environments to test your code. Use the
     following commands from the main repository folder:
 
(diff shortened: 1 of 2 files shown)


rvhonorato:
I'm not qualified for reviewing this, that's why I removed my assignment - I don't know why it was there in the first place or why its important for it to be removed, thanks for asking :)

douweschulte (approved):
Nice change. This removes around 95% of the tox output in the action runs, so it should help in making it easier to spot which ones are actually failing.

joaomcteixeira:
Yep! Thanks :wink:

PR #424, opened 2022-05-03
update `pdb-tools version` to `2.4.5`

joaomcteixeira:
Update `pdb-tools version` to `2.4.5`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-pdb-tools==2.4.3
+pdb-tools==2.4.5
 biopython==1.79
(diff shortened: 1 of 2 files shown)


douweschulte (approved):
Looks fine, looking through the changes for pdb-tools (as there is no changelog) nothing should be breaking.

joaomcteixeira:
Yes. We need to update the changelog in pdb-tools and make it generate automatically. https://github.com/haddocking/pdb-tools/pull/108 is driving me crazy :-x.

In pdb-tools we strickly follow SV2.0, no API breaks with this upgrade.

Thanks @douweschulte !

PR #425, opened 2022-05-03
Implements `--start-from-copy` option defined in #392

joaomcteixeira:
You are about to submit a new Pull Request. Before continuing make sure you read the [contributing guidelines](CONTRIBUTING.md) and you comply with the following criteria:

- [x] You have stick to Python. Talk with us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [ ] Your code is well documented: proper docstrings and explanatory comments for those tricky parts
- [x] You structured the code into small functions as much as possible. You can use classes if there's a (state) purpose
- [x] code follows our coding style
- [x] You wrote tests for the new code
- [x] `tox` tests pass. *Run `tox` command inside the repository folder*
- [x] `-test.cfg` examples execute without errors. *Inside `examples/` run `python run_tests.py -b`*
- [x] PR does not add any *install dependencies* unless permission granted by the HADDOCK team
- [x] PR does not break licensing
- [ ] Your PR is about writing documentation for already existing code :fire:
- [ ] Your PR is about writing tests for already existing code :godmode:

---

* implements `--start-from-copy` option defined in #392

PR #381, opened 2022-04-06
Faster fuzzy matching on user supplied parameters and paths

douweschulte:
Closes #373.

You are about to submit a new Pull Request. Before continuing make sure you read the [contributing guidelines](CONTRIBUTING.md) and you comply with the following criteria:

- [x] You have stick to Python. Talk with us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [x] Your code is well documented: proper docstrings and explanatory comments for those tricky parts
- [x] You structured the code into small functions as much as possible. You can use classes if there's a (state) purpose
- [x] code follows our coding style
- [x] You wrote tests for the new code
- [x] `tox` tests pass. *Run `tox` command inside the repository folder*
- [x] `-test.cfg` examples execute without errors. *Inside `examples/` run `python run_tests.py -b`*
- [x] PR does not add any *install dependencies* unless permission granted by the HADDOCK team
- [x] PR does not break licensing
- [ ] Your PR is about writing documentation for already existing code :fire:
- [ ] Your PR is about writing tests for already existing code :godmode:

---

Same as previous PR but now faster hopefully.

PR #427, opened 2022-05-04
Improves method to find step folders in a run directory

joaomcteixeira:

- [ ] Your PR is about CNS

---

While preparing #425, I realized the way we search for step folders (`0_topoaa`, `1_rigidbody`, etc...) in a run directory can be improved because before we were considering only folders starting with a number. Now, we consider folders starting with a number, plus `_`, and that are named the same as modules.

### DIFF
--- a/src/haddock/modules/__init__.py
+++ b/src/haddock/modules/__init__.py
@@ -427,7 +427,7 @@ def get_module_steps_folders(folder):
     list of str
         List containing strings with the names of the step folders.
     """
-    folders = [p for p in Path(folder).glob() if p.is_dir()]
+    folders = [p for p in Path(folder).iterdir() if p.is_dir()]
     fr = re.compile(step_folder_regex)
-    steps = sorted(f for f in folders if fr.search(f))
+    steps = sorted(f for f in folders if fr.search(str(f)))
     return steps
--- a/src/haddock/modules/__init__.py
+++ b/src/haddock/modules/__init__.py
@@ -437,7 +437,7 @@ def get_module_steps_folders(folder):
     list of str
         List containing strings with the names of the step folders.
     """
-    folders = [p.name for p in Path(folder).iterdir() if p.is_dir()]
+    folders = (p.name for p in Path(folder).iterdir() if p.is_dir())
     steps = sorted(
         (f for f in folders if step_folder_regex_re.search(f)),
         key=lambda x: int(x.split("_")[0]),
--- a/src/haddock/modules/__init__.py
+++ b/src/haddock/modules/__init__.py
@@ -1,11 +1,12 @@
 """HADDOCK3 modules."""
+import re
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from functools import partial
 from pathlib import Path
 
 from haddock import EmptyPath, log, modules_defaults_path
-from haddock.core.defaults import MODULE_IO_FILE, modules_folder_prefix
+from haddock.core.defaults import MODULE_IO_FILE
 from haddock.core.exceptions import ConfigurationError
 from haddock.gear.config_reader import read_config
 from haddock.gear.config_writer import convert_config as _convert_config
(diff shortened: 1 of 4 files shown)


src/haddock/modules/__init__.py
```diff
@@ -53,6 +54,30 @@
     ".cfg": read_config,
     }
 
+_step_folder_regex = tuple(
+    r"[0-9]+_" + mod_name
+    for mod_name in modules_category.keys()
+    )
+step_folder_regex = "(" + "|".join(_step_folder_regex) + ")"
```
douweschulte, line 61: I do not think you need the group here (braces) as the group is not used anywhere and does not change the behaviour of the pipe `|`.

douweschulte (approved):
Nice improvement. But for the future, if the goal is to allow users to define their own modules the use of a predefined set of modules may be detrimental as that means that any module added has to change the codebase of the whole haddock3.

Maybe accepting any folder starting with `\d+_` is acceptable as creating a folder named this way is clearly trying to interfere with the execution, and so could result in `Undefined Behaviour`.
Or change the behaviour to storing a list of all generated module folders somewhere in Haddock so that you do not have to query the filesystem to know what modules you ran.

src/haddock/modules/__init__.py
douweschulte: Another thing, here on line 58 in the regex instead of `[0-9]` you could use `\d` as this does the same.

mgiulini (approved):
I agree with @douweschulte here, I wouldn't use the hard-coded modules' categories in the future. As for the rest, nice!

joaomcteixeira:
> Nice improvement. But for the future, if the goal is to allow users to define their own modules the use of a predefined set of modules may be detrimental as that means that any module added has to change the codebase of the whole haddock3.

Wow, that's crazy forward-thinking. But correct. First, the regex is not hard coded. New modules added to the `modules/` folder will be represented by the regex, **without** the need to modify any code. But what you say applies to someone developing a module **outside** the haddock3 code base, for example in Jupyter, and somehow trying the module in a pipeline. However, many other parts of haddock3 rely on the modules defined in the `modules/` folder to verify configuration files and other parts. So, you have a point in what you are saying. But I don't know how much it will take for us to reach that capacity nor how it would apply. For now, this regex addition avoids errors for someone writing a numbered prefix folder in a run directory, thus breaking the functionalities of the `--restart` flag.

> Maybe accepting any folder starting with \d+_ is acceptable as creating a folder named this way is clearly trying to interfere with the execution, and so could result in Undefined Behaviour.

Yeah... or no? Good question. Was a way to make things more sure.

> Or change the behaviour to storing a list of all generated module folders somewhere in Haddock so that you do not have to query the filesystem to know what modules you ran.

I will think about it in the future. It's a good idea. I did not consider it for #425 and #394.

Your points are really good. Changing the regex in the future back to a `\d+` is very easy. But for now. We can go this way I think. About registering a file with the modules that run. It's possible, but there should be a good reason why. The reasons you gave are good, but now it's maybe too early? I like your comments @douweschulte :wink:

Regarding the braces (`()`), I don't know. I have to see. I have ideas to use the regex somewhere else, where braces may be usefull.

douweschulte:
Yes they were quite forward facing points, but I think it is always good to think about such things when writing code as it helps writing nice modular extensible code. 
About the braces, they will capture the full match in a group, which does not add anything over just taking the full match after matching. Do with this information what you want but I prefer writing thing in the least complex way possible as a general rule of thumb, and these two characters do not add anything.
It is nice to hear that the names will automatically be 'loaded' from the modules folder.

joaomcteixeira:
I will test the braces. The braces "capture" the match. Otherwise, it just matches. I will need that capture for #425. If later I see they are not needed, I will remove them.

Yes. New modules are automatically capture here:

https://github.com/haddocking/haddock3/blob/c98f4a5e2d5bf96c499f124c272f00635099729d/src/haddock/modules/__init__.py#L26-L31

The modules' names and categories are then used throughout to validate many points. For example, the names you implemented in #381.

The capacity of an advanced user to design a module in a Jupyter and run it stand-alone is possible. However, to integrate the module in a pipeline **without** passing through the `haddock3` CLI is way advanced; possible, but way advanced. And would work. It would only "break" if the user later wants to use `haddock3` CLI in a run made in that advanced manner I discussed before. It's a very unlikely scenario. And I think the problems of matching a `1_nothing` folder are higher at this point. Again, creating a file with a list of the generated modules or reverting the regex back to a `\d+_` is fast. Or making it a way to register  a new module manually, also possible. Thanks @douweschulte, really. Very nice discussion. Your previous comment on the zero fill prefix made me make 3 PRs :smile:

douweschulte:
No problem, thanks for letting me have these discussions!

PR #185, opened 2021-12-07
Add option to sort-by in `caprieval`

rvhonorato:
This is a remake of #182, now [in pure-python implementation since we should not use `pandas`](https://github.com/haddocking/haddock3/pull/182#issuecomment-987864556) which again closes #153.

PR #441, opened 2022-05-30
Refactor `caprieval`

### LINKED ISSUE #403: implement model-wise sequence alignment in caprieval
Currently, the CAPRI class, when instantiated, performs a single sequence alignment between the reference and the first input model. This is wrong, as the set of input models might have different alignments. Therefore, we must do every reference-model sequence alignment at the beginning of the CAPRI module.

Discussed with @rvhonorato

### LINKED ISSUE #437: Refactor `align_seq` to bypass lack of chain parity
During `align_seq` function, since there's no chain matching yet (#179) there's a specific `assert` to make sure the chains from `seqdic_ref` and `seqdic_model` match. This should be changed to throw an exception instead;

```
[2022-05-28 19:49:12,853 libutil ERROR]
  File "/trinity/login/abonvin/haddock_git/haddock3/src/haddock/libs/libutil.py", line 341, in log_error_and_exit
    yield
  File "/trinity/login/abonvin/haddock_git/haddock3/src/haddock/clis/cli.py", line 180, in main
    workflow.run()
  File "/trinity/login/abonvin/haddock_git/haddock3/src/haddock/libs/libworkflow.py", line 31, in run
    step.execute()
  File "/trinity/login/abonvin/haddock_git/haddock3/src/haddock/libs/libworkflow.py", line 105, in execute
    module.run()
  File "/trinity/login/abonvin/haddock_git/haddock3/src/haddock/modules/__init__.py", line 204, in run
    self._run()
  File "/trinity/login/abonvin/haddock_git/haddock3/src/haddock/modules/analysis/caprieval/__init__.py", line 98, in _run
    capri_obj = CAPRI(
  File "/trinity/login/abonvin/haddock_git/haddock3/src/haddock/modules/analysis/caprieval/capri.py", line 59, in __init__
    self.model2ref_numbering = align_func(reference, model, path)
  File "/trinity/login/abonvin/haddock_git/haddock3/src/haddock/libs/libalign.py", line 554, in align_seq
    assert ref_chain == model_chain
AssertionError
```

rvhonorato:

- [ ] Your PR is about CNS

---

This PR refactors `caprieval` to properly handle scoring scenarios.

* Closes #437 
* Closes #403 
* Keeps functionality of #385, I basically moved the `run` method into the `CAPRI` class, please have a look @mgiulini 

If a model fails the capri-evaluation it will not show up on `capri_ss.txt` or `capri_clt.txt`, is that the expected behaviour @amjjbonvin ?


src/haddock/libs/libio.py
```diff
@@ -35,6 +37,101 @@ def read_from_yaml(yaml_file):
...
+    output_fname : str or Path
+        Name of the output file.
+
+    Notes
+    -----
+    This function is used to write nested dictionaries.
+    {int: {key: value}}, the int key will be discarded.
+    """
+    first_key = list(data_dict.keys())[0]
+    header = "\t".join(data_dict[first_key].keys())
```
joaomcteixeira, line 109: Suggestion. You may wish to use `json.dumps` here.

joaomcteixeira:
Naive question. Why `caprijob` is no longer needed?

src/haddock/libs/libio.py
```diff
@@ -35,6 +37,101 @@ def read_from_yaml(yaml_file):
...
+    with open(output_fname, 'w') as fout:
+        fout.write(info_header)
+        fout.write(str(data_dict))
+
+    header = "\t".join(data_dict.keys())
+
+    if info_header:
+        header = info_header + os.linesep + header
+
+    with open(output_fname, "w") as out_fh:
```
joaomcteixeira, line 67: `open(..., 'w')` is already used above. I think here you wanted to use `a`, instead?

src/haddock/libs/libio.py
```diff
@@ -35,6 +37,101 @@ def read_from_yaml(yaml_file):
...
+    data_dict : dict
+        Dictionary to write.
+    output_fname : str or Path
+        Name of the output file.
+    info_header : str
+        Header to write before the data.
+    """
+    with open(output_fname, 'w') as fout:
+        fout.write(info_header)
+        fout.write(str(data_dict))
```
joaomcteixeira, line 60: Here. You might want to use `json.dumps` instead of `str`.

src/haddock/libs/libio.py
```diff
@@ -35,6 +37,101 @@ def read_from_yaml(yaml_file):
...
 
 
+def write_dic_to_file(
+        data_dict,
+        output_fname,
+        info_header="",
+        sep="\t",
+        ):
+    """
+    Write a dictionary to a file.
```
joaomcteixeira, line 47: To what I understand, this function does more than write a dictionary to a file. To simply write a dict to a file there's `json.dump` function already. I think it is something more like: "Create a table from a dictionary?"

rvhonorato:
> Naive question. Why `caprijob` is no longer needed?

It was acting as a wrapper for `CAPRI`, if we move the `run` method into it, then the `Scheduler` can pick it up

mgiulini:
Hi, before reviewing the technicalities I'd like to have a small discussion on the overall problem.

In this context, I think that having one micro-job for each model is suboptimal, as it would mean around 100k additional jobs for a standard haddock3 run with 5-10 calls to caprieval => big overhead.

My question is: is there a way to keep the efficient dispatching of models provided by `caprijob` and to account for heterogeneous ensembles? It seems doable to me, as we would only need a list of alignments per job, one per structure. In this way `self.model2ref_numbering` becomes a list, and we use the model-specific alignment for each calculation.

What do you think?

rvhonorato:
> My question is: is there a way to keep the efficient dispatching of models provided by `caprijob` and to account for heterogeneous ensembles? It seems doable to me, as we would only need a list of alignments per job, one per structure. In this way `self.model2ref_numbering` becomes a list, and we use the model-specific alignment for each calculation.

I agree, even made a note in the code about it. 

> It seems doable to me, as we would only need a list of alignments per job, one per structure. 

That's the main issue, if the ensembles are heterogeneous we cannot re-use the alignment; a possibility would be to group all models that are equal into the same "batch", but I think its out of scope of this PR

src/haddock/libs/libio.py
rvhonorato: ?

mgiulini:
> 
> I agree, even made a note in the code about it.
> 

> That's the main issue, if the ensembles are heterogeneous we cannot re-use the alignment; a possibility would be to group all models that are equal into the same "batch", but I think its out of scope of this PR.
> 
> *Also, how are we going to figure out the models are the same without aligning them? 🤔

What I propose here is to do _always_ the assumption that all the models are different. The important key-point is that the reference must be unique, and we can give it for granted I guess. So the only thing to change in the previous code is the object `self.model2ref_numbering`, that becomes a list and the fact that the alignment must be done for each model. But those changes should be easy to implement without deleting the caprijob class.

joaomcteixeira:
Some months ago I had improved the style of the caprieval table (https://github.com/haddocking/haddock3/pull/185). However, now reading the `main` it disappeared somehow and some time ago. It would be nice to recover it as the output was more human readable.

joaomcteixeira:
Should we then have **temporarily** two modules: `caprieval` and `caprieval2`? I think that is **not** a problem. Therefore, instead of discarding the original `caprieval`, make a new module `caprieval2` where both coexist until we are certain of which to discard.

rvhonorato:
As it is now it's completely functional, works for both the docking scenarios and the complex capri sets. The issue raised by @mgiulini is that it will then create a _large_ set of tasks with low load to be passed to the multiprocessor schedule and it would be better to have it a _smaller_ set of tasks with a higher load. However can only be sure if the overhead is really a problem by doing some profiling/benchmark. 

Besides that, neither this nor the previous implementation support HPC/MPI execution and that's something we also need to amend. 

But again the purpose of this PR is to refactor the `caprieval` module so it can work for _real_ scoring sets, which it now does whilst the previous does not.

joaomcteixeira (approved):
Sounds good!

mgiulini (approved):

Issue #329, opened 2022-02-17
Error running example configs


jordantwells42:
Hello,

I am getting an error across my own config files and the example config files in the topology and rigidbody modules.

For topoaa I get the following error:

[2022-02-17 13:18:38,065 __init__ INFO] [topoaa] CNS jobs have finished
[2022-02-17 13:18:38,072 libutil ERROR] 100.00% of output was not generated for this module and tolerance was set to 0.00%.
Traceback (most recent call last):
  File "/mnt/c/Users/jorda/desktop/projects/haddock3/src/haddock/libs/libutil.py", line 320, in log_error_and_exit
    yield
  File "/mnt/c/Users/jorda/desktop/projects/haddock3/src/haddock/clis/cli.py", line 148, in main
    workflow.run()
  File "/mnt/c/Users/jorda/desktop/projects/haddock3/src/haddock/libs/libworkflow.py", line 27, in run
    step.execute()
  File "/mnt/c/Users/jorda/desktop/projects/haddock3/src/haddock/libs/libworkflow.py", line 100, in execute
    module.run(**self.config)
  File "/mnt/c/Users/jorda/desktop/projects/haddock3/src/haddock/modules/base_cns_module.py", line 48, in run
    self._run()
  File "/mnt/c/Users/jorda/desktop/projects/haddock3/src/haddock/modules/topology/topoaa/__init__.py", line 206, in _run
    self.finish_with_error(_msg)
  File "/mnt/c/Users/jorda/desktop/projects/haddock3/src/haddock/modules/__init__.py", line 186, in finish_with_error
    raise RuntimeError(reason)
RuntimeError: 100.00% of output was not generated for this module and tolerance was set to 0.00%.
[2022-02-17 13:18:38,082 libutil ERROR] 100.00% of output was not generated for this module and tolerance was set to 0.00%."

If I set the tolerance to 100% I then get this error, showing that no models were being generated in the topoaa step.

[2022-02-17 13:19:31,112 base_cns_module INFO] Running [rigidbody] module
[2022-02-17 13:19:31,115 __init__ INFO] [rigidbody] crossdock=true
[2022-02-17 13:19:31,115 libutil ERROR] division by zero
Traceback (most recent call last):
  File "/mnt/c/Users/jorda/desktop/projects/haddock3/src/haddock/libs/libutil.py", line 320, in log_error_and_exit
    yield
[...]

PR #469, opened 2022-06-07
Add CNS installation instructions

cvnoort:
General instructions for installing/recompiling CNS for HADDOCK added to INSTALL.md.

- [x] Your PR is about CNS

---

Added section to INSTALL.md that lists the basic steps for installing CNS. 

For now, it links to the CNS routines to add on Dropbox. I would be happy to also add these to the repository and update the instructions accordingly, would just need to know where.


amjjbonvin:
I would suggest to create a separate document just for CNS and refer to it in the INSTALLATION.md file

docs/INSTALL.md
```diff
@@ -4,6 +4,58 @@ Open a `terminal` window and navigate to the folder where you want to
...
+
+```
+# CHANGE THE NEXT LINE TO POINT TO THE LOCATION OF THE CNSsolve DIRECTORY
+
+        CNS_SOLVE=PATH/TO/cns_solve_1.3
+```
+
+#### Add HADDOCK routines
+
+HADDOCK requires some adjustments in CNS to function properly. Download the [set of routines to add](https://www.dropbox.com/s/wliubqovuusqdvr/cns.tgz?dl=0). Uncompress the archive and move all of its files into the CNS source directory (`PATH/TO/cns_solve_1.3/source/`).
```
amjjbonvin, line 44: Where is the code coming from? Is this something we are providing? If not, may-be we better create our own archive and provide it under SURF drive for example (I can handle that)

docs/INSTALL.md
```diff
@@ -4,6 +4,58 @@ Open a `terminal` window and navigate to the folder where you want to
...
+
+#### Recompile CNS
```
amjjbonvin, line 46: What about the compiler requirements?   And any OS-specific instructions?

docs/INSTALL.md
cvnoort: This is a link that was provided to a user under #329 , but can of course be replaced with anything more appropriate.

docs/INSTALL.md
cvnoort: Will add more details once I've moved the CNS instructions to a separate file!

amjjbonvin:
Should be indeed.
A good idea to package it together I would say. 
> Is this the same as https://github.com/haddocking/haddock/tree/haddock2.4/cns1.3? Since these might be crucial for the proper work of cns and haddock3, would it not be better if we package it together with the repo

docs/CNS.md
```diff
@@ -0,0 +1,166 @@
...
+source cns_solve_env
+```
+
+For sh/bash:
+
+```bash
+source .cns_solve_env_sh
+```
+
+To avoid having to repeat this step each time you open a terminal window to run CNS/HADDOCK, you may want to add the relevant `source` line to your `~/.cshrc`, `~/.tcshrc`, `~/.bashrc`, or `~/.bash_profile`. When doing this, make sure to use the full path instead of only the file name.
```
joaomcteixeira, line 139: I think this is not needed to run HADDOCK. At least, I don't have it in my `.bashrc`. It doesn't hurt, though :relaxed:

docs/INSTALL.md
```diff
@@ -52,10 +54,10 @@ python setup.py develop --no-deps
 mkdir -p bin/
 
 # on mac
-ln -s /PATH/TO/cns_solve-1.31-UU-MacIntel.exe bin/cns
+ln -s /PATH/TO/cns_solve_1.3/mac-intel-darwin/source/cns_solve-2206031450.exe bin/cns
 
 # on linux
-ln -s /PATH/TO/CNS_FOLDER/intel-x86_64bit-linux/source/cns_solve-2002171359.exe bin/cns
+ln -s /PATH/TO/cns_solve_1.3/intel-x86_64bit-linux/source/cns_solve-2002171359.exe bin/cns
```
joaomcteixeira, line 60: I suggest keeping the `CNS_FOLDER` instead of `cns_solve_1.3` because it better explains the command, as it is more general.

docs/CNS.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+wget https://www.dropbox.com/s/wliubqovuusqdvr/cns.tgz
+tar -xvf cns.tgz
+mv cns1.3/* source/
+```
+
+## 3 Compile CNS
+
+#### Start compilation
```
joaomcteixeira, line 48: This head should have 3 `#` instead of 4.

```markdown
### Start compilation
```

docs/CNS.md
```diff
@@ -0,0 +1,166 @@
...
+
+You should see the progress of the compilation, followed by a confirmation that an executable has been created. For example (with gfortran):
+
+```
+created executable file cns_solve-2206031450.exe
+```
+
+If this worked, you can continue to the [next step](#4-Source-CNS).
+
+#### Install a compiler
```
joaomcteixeira, line 72: I suggest moving this section and the next one to a `Troubleshooting` section at the end of the file. Like that, we can continue adding examples and solutions to possible future problems without perturbing the reading flow.

rvhonorato (approved):

PR #473, opened 2022-06-10
Support toml with repeated modules

sverhoeven:

- [ ] Your PR is about CNS

---

In https://github.com/i-VRESSE/workflow-builder I use a [toml library](https://www.npmjs.com/package/@ltd/j-toml) to write the config file. The toml library does not support repeats of the same header. To repeat headers I need to make them unique by appending an index and putting quotes around the module name. 

This PR adds support to haddock3 to parse config file like:

```toml
[headerone]
val = 10

['headerone.2']
val = 20

['headerone.3']
val = 30
```
to
```python
{
  "headerone": {'val': 10},
  "headerone.1": {'val': 20},
  "headerone.2": {'val': 30},
}
```

This config file can be written or loaded by any standard compliant toml library and should make it easier for haddock3 config files to be produced or consumed by other programs.

### DIFF
--- a/src/haddock/gear/config_reader.py
+++ b/src/haddock/gear/config_reader.py
@@ -79,7 +79,7 @@ from haddock.libs.libfunc import false, give_same, nan, none, true
 
 # Captures the main headers.
 # https://regex101.com/r/9urqti/1
-_main_header_re = re.compile(r'^ *\[(\w+)\]', re.ASCII)
+_main_header_re = re.compile(r'^ *\[\'?(\w+)(?:\.\d+\')?\]', re.ASCII)
 
 # Captures sub-headers
 # https://regex101.com/r/6OpJJ8/1
--- a/tests/test_config_reader.py
+++ b/tests/test_config_reader.py
@@ -34,6 +34,7 @@ def on_linux():
     [
         ('[header]', 'header'),
         ('[under_scores_work]', 'under_scores_work'),
+        ("['header.3']", 'header')
         ],
     )
 def test_main_header_re(line, expected):
@@ -647,6 +648,24 @@ _config_example_dict_3 = {
         }
     }
 
+_config_example_4 = """
+[headerone]
+val = 10
+
+['headerone.2']
+val = 20
+
+['headerone.3']
+val = 30
+
+"""
+
+_config_example_dict_4 = {
+    "headerone": {'val': 10},
+    "headerone.1": {'val': 20},
+    "headerone.2": {'val': 30},
+    }
+
 
 @pytest.mark.parametrize(
     'config,expected',
@@ -654,6 +673,7 @@ _config_example_dict_3 = {
         (_config_example_1, _config_example_dict_1),
         (_config_example_2, _config_example_dict_2),
         (_config_example_3, _config_example_dict_3),
+        (_config_example_4, _config_example_dict_4),
         ],
     )
 def test_read_config(config, expected):


joaomcteixeira:
Hi @sverhoeven 

Good addition!

The only caveat I see on this is the difference between the two config styles for users. But I perfectly understand the restraints required by the web interface. It's just a matter that in the official examples we advertise only the one without integers. FYI, we have already a [config writer module](https://github.com/haddocking/haddock3/blob/main/src/haddock/gear/config_writer.py) in case it helps you.

Some technical aspect I am trying to sort out is that our config requires the multiline lists to have an empty line afterwards. I think I need to sort that out before this PR can go in because `toml` does not have such a requirement.

Another thing, you also need to operate in this function:

https://github.com/haddocking/haddock3/blob/7d4bb3419e0ad460e608274b438f984877f6e57c/src/haddock/gear/config_reader.py#L378

You need to `key.split(".")[0]` before the function applies the index. You can add that inside the function itself. I would rather not skip the function in case there is a mix of "with index" and "without index" in the config file. In the later case, the index should be ignored and haddock3 should handle it internally (by the `_update_key_number` function).

And, I think this is all. :wink:

p.s. - an alternative would that the web interface create a toml string (internally) and then quickly remove the header integers before saving the string to the file. I don't know if this is possible in the web interface. But, it's an idea. With the python `json` module, for example, it is possible to generate the string before saving it to the file.

Cheers,

sverhoeven:
The test_read_config test has been expanded to test mix of "with index" and "without index".
Added a short description in the module doc string that headers with index is a thing.

haddock3 and the workflow-builder both need to read and write config files.
it would be nice that both can read "with index" or "without index" and haddock3 only writes "without index" and the workflow-builder only writes "with index"  .

joaomcteixeira:
Okay, I will accommodate that. We can have:

1. reading with and without index **automatically** parsed by the function
2. writing with or without index by **boolean** flag.

Sounds good? :relaxed:

Internally, the dictionaries entering the `writing` function **have** indexes (by obviously reasons). In writing, indexes are removed here: https://github.com/haddocking/haddock3/blob/7d4bb3419e0ad460e608274b438f984877f6e57c/src/haddock/gear/config_writer.py#L106

for example:

```python
>>> get_module_name("topoaa.1")
"topoaa"
```

rvhonorato:
> it would be nice that both can read "with index" or "without index" and haddock3 only writes "without index" and the workflow-builder only writes "with index" .

@joaomcteixeira are you working on this or should I try?

joaomcteixeira:
You can work on it.

joaomcteixeira (approved):
I have now reviewed this pull request. Sorry for the long delay, as it was still "draft" I thought it was still a work in progress. It seems very okay to merge. Thanks!

joaomcteixeira:
> it would be nice that both can read "with index" or "without index" and haddock3 only writes "without index" and the workflow-builder only writes "with index" .

I would leave this for another Pull Request following https://github.com/haddocking/haddock3/discussions/518

We can work with `toml` reading edited strings instead of files directly, i suggest.

joaomcteixeira:
lint issues above are addressed here: https://github.com/haddocking/haddock3/pull/523

sverhoeven:
Thanks for fixing the linting errors, I have merged main and now CI and `tox -e lint` are green.

joaomcteixeira:
Hi @sverhoeven

a quick question before merging. Why do the headers need to be quoted to have the indexes? Is it because to avoid conflicts with parameters that are dictionaries? If so, I think we can safely edit the regexes because no parameters start with digits. Let me know, Cheers.

sverhoeven:
> Hi @sverhoeven
> 
> a quick question before merging. Why do the headers need to be quoted to have the indexes? Is it because to avoid conflicts with parameters that are dictionaries? If so, I think we can safely edit the regexes because no parameters start with digits. Let me know, Cheers.

TOML libraries treat `.` in table name with and without quotes differently.

For example
```toml
[foo]
x = 1

['foo.1']
x = 2

[bar]
y = 1

[bar.1]
y = 2
```

will be parsed as

```python
{
    "foo": {"x": 1}, 
    "foo.1": {"x": 2}, 
    "bar": {
        "y": 1, 
        "1": {"y": 2}
    }
 }
```
Having `[bar]` and `[bar.1]` will give keys on different levels, while with quotes the keys are on same level.
So using quotes gives a data structure, more like what the parser does with repeated table names (without digits or qoutes).

joaomcteixeira:
closed by https://github.com/haddocking/haddock3/pull/539
Thanks!

PR #476, opened 2022-06-13
Add general parameters to `haddock3-cfg` client

joaomcteixeira:

- [ ] Your PR is about CNS

---

Now, `haddock3-cfg` can also output the general parameters for modules combined with those of the requested module.


joaomcteixeira:
Global parameters do not vary per module. We can have a separate listing, but it should be in the same command line. For example:

`haddock-cfg -m topoaa -g` -> outputs **also** the general parameters
`haddock-cfg -G` -> outputs **only** the general parameters

I think these options should be kept in the same command-line because in this way there's a single place to read the docs. What do you think? I can make it in this same PR.

amjjbonvin:
Sounds like a good idea!

joaomcteixeira:
Add what we discussed @amjjbonvin . But Instead of `haddock-cfg -G` we can use:

`haddock-cfg -g`
`haddock-cfg -m MODULE`
`haddock-cfg -m MODULE -g`

amjjbonvin (approved):

PR #482, opened 2022-06-20
Add OpenMM module

joaomcteixeira:
The code presented in this commit was entirely done by **Ivar de Leeuw** during his Master course: Major Research Project (GSLS), academic year 2021-2022.

I am pushing the code after Ivar's request and in its exact original form.

In the next commits, the team will work on Ivar's code to accommodate it fully to the HADDOCK3 project.

Thanks, Ivar, for all your great work and best luck with your future studies!


docs/INSTALL.md
```diff
@@ -135,3 +135,18 @@ export GDOCK_PATH=some-folder
 instructions as here only the model generation is used. Please check the
 [repository page](https://github.com/rvhonorato/gdock) for more
 information.
+
+## `openmm`
+
+1. Install the latest version of openmm:
```
joaomcteixeira, line 141: Since we have all in the same `## openmm` header, what do you think about suggesting a single command? `conda install -c conda-forge openmm pdbfixer`? It would be easier for `conda` to evaluate requirements also, i think.

mgiulini: yes, agreed

mgiulini:
tried a run with the example file `openmm-test.cfg` and commit ID #c79f47a. It's already a great work. A few "style" improvements from my user experience:

- [ ] we should clarify if the simulation uses implicit or explicit solvent at the beginning
- [ ] we should report the difference between equilibration and the actual simulation
- [ ] the logging during the actual simulation should be more frequent to allow the user to track the progress, we should inspect a bit the core openmm functions
- [ ] we could split the function runOpenMM in two or more pieces for enhancing the readability
- [ ] a warning appeared `Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.` probably we can safely ignore it, but I still have to check

joaomcteixeira:
@mgiulini, the pre-last commit solves the `globals()` problem. It was not necessary because the value was the same as the key, so we can simply avoid calling the `globals()` dictionary.

Also corrected some defaults that were lists to single strings

mgiulini:
As for the reproducibility: it is possible to assign a seed to the integrators, but not to the modeller functions. Therefore, the solvation boxes specified at the beginning of the simulations are bound to differ.

mgiulini:
First prettified, working version of the openmm module.

Still a lot of things to fix/implement:

- [ ] check for GPU support
- [ ] capability of running on complexes (for ex. coming from a topoaa ensemble)
- [ ] tests: how to implement them?
- [ ] the check for the correct installation must be improved and it should be executed only if the openmm module is part of the chosen workflow

joaomcteixeira:
Hi @mgiulini I have addressed what we discussed on slack; tests are now passing. Can you test the workflows to see if they work as expected?

We can have the tests for the `openmm` module itself in a separate pull request.

joaomcteixeira:
okay @mgiulini , reverted the parameters to their original name and considered them as strings because considering them as `_fname` would force us to add some exception in the `prepare_run`. We will see in the future if we need to consider them as paths for any reason. But surely not. For example, in the future, in the case the user can provide the path to a specific forcefield file, we can have it as a separate parameter (`custom_forcefield_fname`, for example).
Cheers

joaomcteixeira:
All examples run. :+1:

mgiulini:
@joaomcteixeira I see that the topoaa module, if it's not the first module in the .cfg file, creates the topology only for the first input model. I remember we somehow discussed this in the chat with Ivar. In the context of openmm we should generate the topology for each input model, what do you think?

try to run the example file `openmm-topoaa-flexref-test.cfg`, you will see that only a complex gets to the second `topoaa` and then to `flexref`

joaomcteixeira:
Yes, I remember that. The part which controls that is here:

https://github.com/haddocking/haddock3/blob/c404a4ab309e2f7eac181dfc0226e4e6d499bf3c/src/haddock/modules/topology/topoaa/__init__.py#L118-L128

And the culprit is that `[0]` index. We can make `[0]` a parameter or a *guesser* that behaves differently according to the situation that is presented to `topoaa`. In case we chose it t be a parameter the user needs to know what he/she's doing and know the parameter exists (which is not bad in the end). A guesser can be defined following the module preceding `topoaa`, but that would imply that the behaviour is constant for each module.

joaomcteixeira:
It makes sense what you are saying. Maybe the `[0]` was something made just as a draft. I think it is fine to remove the `[0]` and update the `topoaa` operations when performed in the middle of the workflow. I think you can do that directly here in this PR.

mgiulini:
> It makes sense what you are saying. Maybe the `[0]` was something made just as a draft. I think it is fine to remove the `[0]` and update the `topoaa` operations when performed in the middle of the workflow. I think you can do that directly here in this PR.

will do, thanks for the feedback! I think it's a good improvement, as it allows to reduce the complexity of the overall logic. maybe I will create a separate PR, as this is not strictly related to openmm and could be relevant in other scenarios.

joaomcteixeira:
Perfect. I agree with using a new PR! :+1:

mgiulini:
hi all! I would say that the first alpha version of the openmm module is ready for review. Although I changed the code substantially, the overall structure is the same as the one created by Ivar. There are still some major improvements to be made, such as

1. the addition of GPU support
2. the creation of a test suite
3. a more resources-friendly creation of the solvation box

Still, I would say that these can be addressed separately in other issues.

src/haddock/modules/refinement/openmm/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+# allow general testing when OpenMM is not installed
+with suppress(ImportError):
+    from haddock.modules.refinement.openmm.openmm import OPENMM
+
+
+RECIPE_PATH = Path(__file__).resolve().parent
+DEFAULT_CONFIG = Path(RECIPE_PATH, "defaults.yaml")
+
+
+def runSubprocess(command_to_run):
```
joaomcteixeira, line 21: I suggest moving this function to `libsubprocess` with the name `run_subprocess`.

src/haddock/modules/refinement/openmm/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+                    identificator=i,
+                    model=model_to_be_simulated,
+                    path=Path("."),
+                    params=self.params,
+                    directory_dict=directory_dict
+                    )
+                )
+
+        # running jobs
+        # currently it only accepts local and mpi.
```
joaomcteixeira, line 92: Any clue why `hpc` mode is not accepted?

src/haddock/modules/refinement/openmm/__init__.py
```diff
@@ -0,0 +1,114 @@
...
+        for pdb in output_pdbs:
+            pdbPath = os.path.join(directory_dict["openmm_output"], pdb)
+            pdbToExport = PDBFile(pdbPath)
+            models_to_export.append(pdbToExport)
+        
+        self.output_models = models_to_export
+        self.export_output_models()
+
+        self.log('Completed OpenMM module run.')
+        topoaa_reminder = ['If you want to continue the haddock3 workflow',
```
joaomcteixeira, line 110: Very curious :wink: :coffee: :snake:

```python
self.log(
    'If you want to continue the haddock3 workflow'
    ' after the OpenMM module, the next module should'
    ' be topoaa, to rebuild the molecular topologies.'
    )
```

src/haddock/modules/refinement/openmm/defaults.yaml
```diff
@@ -0,0 +1,206 @@
...
+  long: Simulation timesteps performed during the simulation.
+  group: ''
+  explevel: easy
+
+center_atoms:
+  default: false
+  type: boolean
+  title: Move atoms to the center of the solvation box.
+  short: Move atoms to the center of the solvation box. The pdb is placed around the origin, and water molecules of the solvation box are build up from the origin.
+  long: Move atoms to the center of the solvation box. The pdb is placed around the origin, and water molecules of the solvation box are build up from the origin. Not centering the atoms shouldn't be a problem if periodic boundary conditions are deployed, but employing it also prevents the molecule from jumping around the boundary while visually inspecting the simulation in e.g. PyMOL.
```
joaomcteixeira, line 130: FYI:

It is possible to split into multiple lines as:

```yaml
  long: Move atoms to the center of the solvation box.
    The pdb is placed around the origin, and water molecules of the solvation box are build up from the origin.
    Not centering the atoms shouldn't be a problem if periodic boundary conditions are deployed,
    but employing it also prevents the molecule from jumping around the boundary while visually
    inspecting the simulation in e.g. PyMOL.
```

src/haddock/modules/refinement/openmm/openmm.py
```diff
@@ -0,0 +1,308 @@
...
+    def get_pdb_filepath(self, folder=None):
+        """Get correct path to pdb file."""
+        if folder:
+            pdb_filepath = os.path.join(self.directory_dict["pdbfixer"],
+                                        self.model.file_name)
+        else:
+            if isinstance(self.model, PDBFile):
+                pdb_filepath = f'{self.model.rel_path}'
+            else:
+                pdb_filepath = f'{Path(self.model.path) / self.model.file_name}'
```
joaomcteixeira, line 123:
```suggestion
                pdb_filepath = str(Path(self.model.path,  self.model.file_name))

File: src/haddock/modules/refinement/openmm/openmm.py [hunk 7]
```diff
@@ -0,0 +1,308 @@
...
+                                boxsize_nm,
+                                boxsize_nm) * nanometers
+                modeller.addSolvent(usedForcefield,
+                                    boxSize=box_size,
+                                    neutralize=False
+                                    )
+
+            # Add required extra particles for forcefield,
+            # e.g. Drude particles.
+            if (self.params['add_extra_particles_for_forcefield']):
```
joaomcteixeira, line 179:
```suggestion
            if self.params['add_extra_particles_for_forcefield']:
```

src/haddock/modules/refinement/openmm/openmm.py
```diff
@@ -0,0 +1,308 @@
...
+                log.info("adding extra particles")
+                modeller.addExtraParticles(forcefield)
+            # check centering atoms
+            if (self.params['center_atoms']):
```
joaomcteixeira, line 183:
```suggestion
            if self.params['center_atoms']:
```

src/haddock/modules/refinement/openmm/openmm.py
```diff
@@ -0,0 +1,308 @@
...
+        output_filepath = os.path.join(self.directory_dict["openmm_output"],
+                                       self.model.file_name
+                                       )
+        with open(input_filepath, 'r') as fh:
+            pdbYieldedLines = pdb_delresname.run(fh, 'HOH')
+            with open(output_filepath, 'w') as writefh:
+                writefh.writelines(pdbYieldedLines)
+
+    def run_openmm(self, inputPDBfile, output_directory, solvent_model):
+        """Run openmm simulation of the model pdb."""
```
joaomcteixeira, line 215:
```suggestion
        """
        Run openmm simulation of the model pdb.

        Parameters
        ----------
        inputPDBfile : str
            Path to the PDB file.

        output_folder : str
            Path to the output directory.

        solvent_model : str
            One of the OpenMM's implicit solvent models. For example:
            'implicit/hct.xml', 'implicit/obc1.xml', 'implicit/obc2.xml',
            'implicit/gbn.xml', 'implicit/gbn2.xml'. For more information:
            http://docs.openmm.org/latest/userguide/application/02_running_sims.html#amber-implicit-solvent
        """

File: src/haddock/modules/refinement/openmm/openmm.py [hunk 10]
```diff
@@ -0,0 +1,308 @@
+"""OpenMM module."""
```
joaomcteixeira, line 1:
```suggestion
"""
Refinement module using OpenMM.

This module uses OpenMM to refine the models before exploring docking in-depth.
We have created a generalized refinement protocol using OpenMM for which
you can modify a minimal set of parameters.

To get a list of all possible parameters, run::

     haddock3-cfg -m openmm
     
This module will refine all models coming from the previous workflow step and send them
to the next step in the workflow. If you want to use other modules such as `flexref` or `emref` after
the OpenMM module, you need to recreate the topologies by simply adding a `[topoaa]` step in the workflow.
See examples in `examples/thirdparty/openmm` folder.
"""
```

Please help me further here to add a short paragraph explaining the major concepts of the implementation.

joaomcteixeira (commented):
Great work @mgiulini !
I can't approve the PR because I am its original author.
Here are some minor comments.

src/haddock/modules/refinement/openmm/defaults.yaml
mgiulini: oh nice, will do!

src/haddock/modules/refinement/openmm/__init__.py
mgiulini: nice, thanks!

src/haddock/modules/refinement/openmm/defaults.yaml
```diff
@@ -149,7 +152,8 @@ solvent_boxsize_nm:
   max: 100
   title: Solvent boxsize in nm.
   short: Solvent boxsize in nm used to build the water solvation box around the molecule.
-  long: Solvent boxsize in nm used to build the water solvation box around the molecule. Value is used when the crystallographic cell size is removed in the pipeline.
+  long: Solvent boxsize in nm used to build the water solvation box around the molecule.
+  Value is used when the crystallographic cell size is removed in the pipeline.
```
joaomcteixeira, line 156: I think you need to indent for the `yam` to be valid: two spaces beyond the `long`.

src/haddock/modules/refinement/openmm/__init__.py
joaomcteixeira: Likely this is the last thing missing in this PR :+1: :wink:

mgiulini: I checked. The problem can be solved running openmm in local mode on gpu architectures. The openmm infrastructure is able to detect the available platforms and to choose the faster among them! This is cool and makes the module far more usaful!

rvhonorato (dismissed):

amjjbonvin:
Aren’t the solvation boxes built from a pre-equilibrated water box?

joaomcteixeira:
image

This is so beautiful!

mgiulini:
> Aren’t the solvation boxes built from a pre-equilibrated water box?

Hi Alex, the solvation box in openmm is built by default using the TIP3P solvent model, which should be already equilibrated. But the solute-solvent interface mixture is not equilibrated at all, as the water molecules are simply removed in correspondence of the solute.

amjjbonvin:
That’s standard procedure. 

The first equilibration step should be to minimise the solvent with the solute position restrainted, and then heat up the system slowly releasing the solute so that the solvent adapts to the solute (and not vice-versa)

mgiulini:
> That’s standard procedure. The first equilibration step should be to minimise the solvent with the solute position restrainted, and then heat up the system slowly releasing the solute so that the solvent adapts to the solute (and not vice-versa)

Precisely! Following Ivar's code, I didn't include the full procedure in the current implementation to save resources. The system is built, energy-minimized, and then quickly "equilibrated" at the simulation temperature (300 K) for a few steps. Should I include the full equilibration protocol?

amjjbonvin:
> Precisely! Following Ivar's code, I didn't include the full procedure in the current implementation to save resources. The system is built, energy-minimized, and then quickly "equilibrated" at the simulation temperature (300 K) for a few steps. Should I include the full equilibration protocol?
> 
I guess this should be checked at least. In principle starting directly at 300K is fine as long as there are positions restraints on the solute that are slowly removed.

mgiulini:
> Precisely! Following Ivar's code, I didn't include the full procedure in the current implementation to save resources. The system is built, energy-minimized, and then quickly "equilibrated" at the simulation temperature (300 K) for a few steps. Should I include the full equilibration protocol?
> I guess this should be checked at least. In principle starting directly at 300K is fine as long as there are positions restraints on the solute that are slowly removed.

I will investigate the code more thoroughly then :)

VGPReys:
resolve #776

rvhonorato (dismissed):
lgtm

docs/INSTALL.md
```diff
@@ -183,3 +183,21 @@ export GDOCK_PATH=some-folder
 instructions as only the model generation is used here. Please check the
 [repository page](https://github.com/rvhonorato/gdock) for more
 information.
+
```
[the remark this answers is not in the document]
VGPReys: Not possible, only conda....

examples/thirdparty/openmm/openmm-topoaa-flexref-test.cfg
```diff
@@ -0,0 +1,67 @@
...
+[topoaa]
+
+[caprieval]
+reference_fname = "data/1nx1_refe.pdb"
+
+[flexref]
+tolerance = 20
+ambig_fname = "data/ambig.tbl"
+# Define peptide as fully flexible
+nfle = 2
```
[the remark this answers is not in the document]
VGPReys: indeed ;)

src/haddock/core/exceptions.py
```diff
@@ -62,6 +62,11 @@ class HaddockModuleError(HaddockError):
     pass
 
 
+class ThirdPartyIntallationError(HaddockModuleError):
+    """General error for when third party modules is not properly installed."""
```
mgiulini, line 66:
```suggestion
    """General error when a third party module is not properly installed."""
```

mgiulini (dismissed):

amjjbonvin (dismissed):
What will happen if a protein-ligand complex is passed to the openmm module?
Is there a  check that only "pure" protein and nuclei acids are accepted?
Also I assume (did not check the openmm code) that the system is neutralised by adding counter ions. Is this true?

VGPReys:
> What will happen if a protein-ligand complex is passed to the openmm module? Is there a check that only "pure" protein and nuclei acids are accepted? Also I assume (did not check the openmm code) that the system is neutralised by adding counter ions. Is this true?

- System is neutralized
- No yet checks for ligands... must do it, good point !

VGPReys:
- Added try/except logic to handle cases where topology cannot be generated due to unknown (by the force-field) residues.
- Also fixed module when using implicit solvent !

PR #484, opened 2022-06-21
run analysis modules after topoaa

### LINKED ISSUE #417: problems running non-cns modules after topoaa
Currently it's impossible to run non-cns modules such as `caprieval` and `clustfcc` immediately after `topoaa`. This could be useful, especially when submitting large ensembles of models for validation/analysis purposes. The error is the following:

`
AttributeError: 'tuple' object has no attribute 'rel_path'
` 

for `topoaa->clustfcc` and `topoaa->caprieval` .cfg files. This is due to the fact that `topoaa` produces a tuple instead of a PDBFile object. This should be an easy fix.

mgiulini:

- [ ] Your PR is about CNS

---

Closes #417 : with this PR it is possible to run the analysis modules `caprieval`, `clustfcc`, and `rmsdmatrix` directly after topoaa, without having to go through cns modules. The logic is very simple: when retrieving the models, the `individualize` parameter is set to True, so that each input model is treated as an individual entity. 

This of course makes sense only if the input is an ensemble of structures: should I add some sort of check on that?


joaomcteixeira (approved):
If you run the examples pipelines and they all work well, then I approve the PR. :relaxed:

The only thing I can suggest is to add a small `*-test.cfg` to the examples folder so that these cases also integrate the testing routines. You don't need to add an example to every subfolder, just to the simplest one that reflects the motivation to develop this.

Cheers,

mgiulini:
> If you run the examples pipelines and they all work well, then I approve the PR. ☺️
> 
> The only thing I can suggest is to add a small `*-test.cfg` to the examples folder so that these cases also integrate the testing routines. You don't need to add an example to every subfolder, just to the simplest one that reflects the motivation to develop this.
> 
> Cheers,

I was thinking about that, I deemed it reasonable but I didn't want to overload the examples..but I guess a small cfg file with only `topoaa` followed by `caprieval` shouldn't be a big deal. Will do it asap, thanks for the feedback!

mgiulini:
> There are already some examples that do not concern biological relevant scenarios but that concern testing advanced functionalities of HADDOCK3. We can actually have a folder only for those instead of having them spread around the other cases. You can start with that in this PR, and then we can make another PR moving those other examples to that new folder.

I agree. This could be of particular relevance in the context of the workflow builder, when one may want to set up a quick analysis.

mgiulini:
@joaomcteixeira I created three small test examples in the `examples/analysis` folder, showing how to perform a cns-independent analysis on an ensemble of docked structures. Check them out!

joaomcteixeira:
Looks great, a final thing. Can you add the examples to the `run_tests.py` script?  We need to add them until we make that automatic.

joaomcteixeira (approved):
perfect. thanks! once tests pass green you can merge :rocket:

PR #494, opened 2022-07-18
Added `exit` module

joaomcteixeira:

- [ ] Your PR is about CNS

---

* Added an `[exit]` module.
* Adding this module to the workflow will stop the workflow at that step
* I decided to make a module following the same hierarchy as for other modules. I found that cleaner than implementing an `if` process in a `gear` because doing so would require rewriting some tests.
* works with `--restart` and `--extend-run`

### DIFF
--- a/docs/modules/extras/exit.rst
+++ b/docs/modules/extras/exit.rst
@@ -1,0 +1,12 @@
+Exit module
+===========
+
+.. automodule:: haddock.modules.extras.exit
+   :members:
+   :show-inheritance:
+   :inherited-members:
+
+Default parameters
+------------------
+
+.. include:: params/exit.rst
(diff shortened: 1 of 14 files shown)


joaomcteixeira:
You might be right @mgiulini. But currently, there is no other place to add those examples. We need to create an `integration tests` folder and separate those from examples. Nonetheless, it still is an example. I would make this a new issue and discuss it there.

One nice thing is that, again, an unexpected feature request is implemented with only green lines (red lines here are mostly edits). :wink:

mgiulini:
> You might be right @mgiulini. But currently, there is no other place to add those examples. We need to create an `integration tests` folder and separate those from examples. Nonetheless, it still is an example. I would make this a new issue and discuss it there.
> 
> One nice thing is that, again, an unexpected feature request is implemented with only green lines (red lines here are mostly edits). 😉

sure, it was more a comment than a question. Create a discussion/issue about this if you like, at some point it will become relevant!

Issue #179, opened 2021-12-06
Implement chain-matching/check in `caprieval`


rvhonorato:
Currently there's no check if the chains of the reference matches the model, we should implement a chain-matching subroutine.

rvhonorato:
Closing this for now because there's no simple way of doing the chain-matching without adding a sequence/structural alignment dependency (as far as I know).

joaomcteixeira:
On the contrary, leave it open, otherwise, I forget it exists. :relaxed:

rvhonorato:
Partially addressed in #230

rvhonorato:
has this been addressed @mgiulini ?

mgiulini:
hi @rvhonorato, I somehow missed this message. 
Nope, it's still an open issue. I can take care of this, but maybe it's better to have a chat on the best way to solve the problem.

PR #507, opened 2022-07-25
implemented check_chains

### LINKED ISSUE #506: correct bug in caprieval chains
an input file whose chains are A and E is not accepted by caprieval.

```
Process Worker-13:
  File "/trinity/login/mgiulini/anaconda3/envs/haddock3/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/trinity/login/mgiulini/haddock3/src/haddock/libs/libparallel.py", line 28, in run
    task.run()
  File "/trinity/login/mgiulini/haddock3/src/haddock/modules/analysis/caprieval/capri.py", line 429, in run
    self.calc_lrmsd()
  File "/trinity/login/mgiulini/haddock3/src/haddock/modules/analysis/caprieval/capri.py", line 148, in calc_lrmsd
    l_start, l_end = chain_ranges[self.l_chain]
KeyError: 'B'
```

The root of this error lies in the parameters `self.r_chain`,` self.l_chain` of caprieval, which, by default, are A and B. Given the wide spectrum of haddock3 scenarios that can be run, the solution is not trivial. My suggestion here is to check if the default chain is present and, if not, use the available chain IDs.

@rvhonorato let me know if you have comments on this

mgiulini:

- [ ] Your PR is about CNS

---

Closes #506 by creating a small function, check_chains, that checks if the expected chains are part of the observed chains. In case they are not, it sticks to the observed chains.

### DIFF
--- a/src/haddock/modules/analysis/caprieval/capri.py
+++ b/src/haddock/modules/analysis/caprieval/capri.py
@@ -134,20 +134,23 @@ class CAPRI:
         P = []
         # Note: this MUST be sorted since we will use the indexes to
         #  separate between receptor and ligand coordinates
-        instersection = sorted(ref_coord_dic.keys() & mod_coord_dic.keys())
+        intersection = sorted(ref_coord_dic.keys() & mod_coord_dic.keys())
 
         chain_ranges = {}
-        for i, segment in enumerate(instersection):
+        for i, segment in enumerate(intersection):
             chain, _, _ = segment
             if chain not in chain_ranges:
                 chain_ranges[chain] = []
             chain_ranges[chain].append(i)
 
         chain_ranges = make_range(chain_ranges)
-        r_start, r_end = chain_ranges[self.r_chain]
-        l_start, l_end = chain_ranges[self.l_chain]
 
-        for k in instersection:
+        obs_chains = list(chain_ranges.keys())  # observed chains
+        r_chain, l_chain = self.check_chains(obs_chains)
+        r_start, r_end = chain_ranges[r_chain]
+        l_start, l_end = chain_ranges[l_chain]
+
+        for k in intersection:
             ref_xyz = ref_coord_dic[k]
             mod_xyz = mod_coord_dic[k]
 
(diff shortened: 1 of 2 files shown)


rvhonorato (approved):

joaomcteixeira (approved):
The code seems okay.

Some questions/considerations: I see only `warnings` are output in case of chains differ. What are the implications to the user regarding this? Can this warning hide any "bug" or produce a result not expected by the user? Is this a case where a check at the startup would fit better to avoid errors down the calculation? Some parameter correction or some check that can be implemented in the logic of PDB preprocessing/checking?

Cheers,

mgiulini:
> The code seems okay.
> 
> Some questions/considerations: I see only `warnings` are output in case of chains differ. What are the implications to the user regarding this? Can this warning hide any "bug" or produce a result not expected by the user? Is this a case where a check at the startup would fit better to avoid errors down the calculation? Some parameter correction or some check that can be implemented in the logic of PDB preprocessing/checking?
> 
> Cheers,

Hey! thanks for checking! The warnings arise when the chain IDs specified by the user (or chosen by default) do not match the observed ones. The user that has a chain A and C in their pdb but selects chains A and B in the cfg file is notified with a warning and caprieval uses the observed chain IDs. This will be corrected when chain matching #179 will be implemented (check also #506), so this modification is temporary.

The preprocessing might be helpful in the context of #179, as it could detect at the beginning of the execution if the observed and expected chains coincide..what's your opinion?

joaomcteixeira:
Thanks for explaining. I agree with you, this PR is useful to overcome bugs and provide a temporary solution. However, as you said, I suggest performing as many checks as possible at startup to avoid problems after hours or days of computation. Generally, the most common users will just ignore warning messages and only look for them in case of problems in the results. So, checks during the `prepare_run` phase are preferable; in this case, such checks are also executed with the `--setup` option, which is a way to confirm the proper setup of the HADDOCK3 run.

You can merge this and yet mention this discussion in an issue.

Cheers,

PR #515, opened 2022-07-28
implement `optint` module

rvhonorato:

- [X] You have stick to Python. Talk with us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [X] Your code is well documented: proper docstrings and explanatory comments for those tricky parts

---

This PR adds a new `optint` module. That does interface optimization via genetic algorithm using the `haddock-score` as fitness function. I've added an option to run the optimization in a specific `region`, if nothing is provided then the interface is automatically detected and used instead. The mutations of the first generation can be either `alanine`, `random` or `keep`, which uses the WT as starting point.

```toml
[optint]
  number_of_generations = 15
  population_size = 100
  # cxpb = 0.6
  # mutpb = 0.8
  # eta = 0.8
  # indpb = 0.2
  output_models = 5
  target_chain = "A"
  region = [101,102,103,104,105,106,107,108]
  # the initial population will be based on the wild-type
  initial_mutation = "keep"
```

Examples are added in `examples/thirdparty/design-antibody-antigen`, 
* Run interface optimization in an Ab complex, targeting the HV3 loop
* Run a docking simulation and subsequently do multiple rounds of interface optimizations targeting the HV3 loop.

For this module to work it needs [deap](https://github.com/DEAP/deap), so you need to install it with `pip install deap`. 

I've also left a jupyter notebook with code to generate these plots;

drawing


examples/thirdparty/design-antibody-antigen/design-ab-docking-w-optimization.cfg
```diff
@@ -0,0 +1,37 @@
...
+#   of interface optimizations targeting the HV3 loop.
+#
+# ====================================================================
+run_dir = "ab-docking-w-optimization"
+ncores = 64
+molecules =  ["data/4G6K_fv.pdb", "data/4I1B-matched.pdb"]
+# ====================================================================
+[topoaa]
+[rigidbody]
+  ambig_fname = "data/ambig.tbl"
```
amjjbonvin, line 13: A more relevant example would be to target a specific epitope, i.e. using instead the ambiguous-CDR-NMR-CSP.tbl restraint file instead of the ambig.tbl one with targets the entire surface

amjjbonvin: Actually you have it in another config file. But this ab initio example does not seem very relevant to me

src/haddock/modules/design/optint/defaults.yaml
```diff
@@ -0,0 +1,128 @@
...
+  type: integer
+  min: 1
+  max: 150
+  title: Output models
+  short: The number of models to output
+  long: The number of models to output
+  group: ""
+  explevel: guru
+
+target_chain:
```
amjjbonvin, line 93: Can you design more than one chain at the same time?  
Is there an interface option to automatically select all residues at the interface?

src/haddock/modules/design/optint/defaults.yaml
```diff
@@ -0,0 +1,128 @@
...
+  short: The probability of independent mutation
+  long: The probability of independent mutation
+  group: ""
+  explevel: guru
+
+output_models:
+  default: 1
+  type: integer
+  min: 1
+  max: 150
```
amjjbonvin, line 86: Why 150? I guess the max must be related to the population of the genetic algorithm, correct?

src/haddock/modules/design/optint/defaults.yaml
```diff
@@ -0,0 +1,128 @@
...
+  minchars: 1
+  maxchars: 1
+  title: Target chain
+  short: The target chain
+  long: The target chain
+  group: ""
+  explevel: guru
+
+region:
+  default: []
```
amjjbonvin, line 105: Shouldn't target and region be merged?  Or rather region belong to the target section?
A region must contain a chain ID to be unambiguously defined

amjjbonvin (commented):
Made various comments.

Also I would suggest to put the example directory directly under examples.
It does use deap, a third party software for the GA, but combines also HADDOCK modules.

amjjbonvin:
PS: Is the animated figure above the result of starting from the WT sequence?  If this is the case we will have to do some optimisation as we move away from it and the score goes up...

amjjbonvin:
Also if we put it in the main branch, it should come with big warnings that while the machinery is working, it is very much untested and work in progress.

src/haddock/modules/__init__.py
```diff
@@ -43,6 +43,7 @@
     "scoring",
     "analysis",
     "extras",
+    "design"
```
joaomcteixeira, line 46: This list is ordered by workflow progression. You may wish to add `design` in a different position.

src/haddock/modules/design/optint/__init__.py
```diff
@@ -0,0 +1,59 @@
+"""interface optimization ."""
```
joaomcteixeira, line 1: I suggest some more details in the documentation. It is just a suggestion.

src/haddock/modules/design/optint/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+from haddock.libs.libontology import PDBFile, TopologyFile
+from haddock.modules import BaseHaddockModule
+
+
+RECIPE_PATH = Path(__file__).resolve().parent
+DEFAULT_CONFIG = Path(RECIPE_PATH, "defaults.yaml")
+
+
+class HaddockModule(BaseHaddockModule):
+    """HADDOCK3 module to perform energy minimization scoring."""
```
joaomcteixeira, line 16: Suggest updating this sentence.

src/haddock/modules/design/optint/defaults.yaml
```diff
@@ -0,0 +1,128 @@
...
+  maxchars: 7
+  choices:
+    - alanine
+    - random
+    - keep
+  title: Initial mutation
+  short: How to handle the initial mutation
+  long: How to handle the initial mutation
+  group: ""
+  explevel: guru
```
joaomcteixeira, line 128: I see all parameters are `guru` level. You may wish to review that.

docs/INSTALL.md
```diff
@@ -135,3 +135,9 @@ export GDOCK_PATH=some-folder
 instructions as here only the model generation is used. Please check the
 [repository page](https://github.com/rvhonorato/gdock) for more
 information.
+
+## `optint` module
+
```
joaomcteixeira, line 140: You may wish to add a short user-friendly sentence here.

joaomcteixeira (approved):
I left minor comments that you can opt to address. The module respects the required API for module classes. I have not checked all other implementations underneath, as those are free to define to the module's creator.

If all `tox` tests pass and the `run_tests.py -b` also passes, to ensure the integrity of all other workflows, everything is good to go.

15k lines plus and 0 edited. Demonstrates the quality of the code base.

Cheers

examples/thirdparty/design-antibody-antigen/design-ab-docking-w-optimization.cfg
rvhonorato: I took these restraints from `examples/docking-antibody-antigen/docking-antibody-antigen-CDR-accessible-clt-full-mpi.cfg` should it be another one instead?

src/haddock/modules/design/optint/defaults.yaml
rvhonorato: > Can you design more than one chain at the same time?

Theoretically yes but that might increase the complexity of the optimization, should be benchmarked first

> Is there an interface option to automatically select all residues at the interface?

If no region is defined then the interface residues are used

src/haddock/modules/design/optint/defaults.yaml
rvhonorato: It is yes, could be any number actually, but also something we should test - will increase for now

examples/thirdparty/design-antibody-antigen/design-ab-docking-w-optimization.cfg
amjjbonvin: I would say we remove this example, there is another one doing docking with the CDR-NMR restraints, correct? That's a targeted docking to the epitope and then design. Would be the most likely scenario

src/haddock/modules/design/optint/defaults.yaml
amjjbonvin: Then this explanation should be part of the documentation in the default.yaml file

src/haddock/modules/design/optint/defaults.yaml
rvhonorato: It would be best, but I don't know which type of parameter allows for that dictionary-like definition - ping @joaomcteixeira

src/haddock/modules/design/optint/defaults.yaml
amjjbonvin: I guess in principle only the best models will be written - so probably a small number anyway

rvhonorato:
> Also if we put it in the main branch, it should come with big warnings that while the machinery is working, it is very much untested and work in progress.

Done

> PS: Is the animated figure above the result of starting from the WT sequence? If this is the case we will have to do some optimisation as we move away from it and the score goes up...

Indeed, this module here serves as a demonstration more then a feature - much more research to be done.

src/haddock/modules/design/optint/defaults.yaml
amjjbonvin: Could be done as for the symmetry restraints where we can define multiple segments, each having a segID and in this case a list of residues. Check e.g. the default.yaml file of the rigidbody module

rvhonorato:
> If all `tox` tests pass and the `run_tests.py -b` also passes, to ensure the integrity of all other workflows, everything is good to go.
> 
> 15k lines plus and 0 edited. Demonstrates the quality of the code base.

But if 0 lines were edited why is `run_tests.py -b` still needed?

amjjbonvin:
> > Also if we put it in the main branch, it should come with big warnings that while the machinery is working, it is very much untested and work in progress.
> 
> Done
> 
> > PS: Is the animated figure above the result of starting from the WT sequence? If this is the case we will have to do some optimisation as we move away from it and the score goes up...
> 
> Indeed, this module here serves as a demonstration more then a feature - much more research to be done.

Should we then have may-be a `modules-devel` directory for modules in development that need proper testing / validation?  Just to avoid that users might waste time on something not yet fully working?

rvhonorato:
> Should we then have may-be a modules-devel directory for modules in development that need proper testing / validation? Just to avoid that users might waste time on something not yet fully working?

Good suggestion! Done

src/haddock/modules-devel/design/optint/modules/pdb.py
```diff
@@ -0,0 +1,70 @@
...
+                if chain not in coord_dic:
+                    coord_dic[chain] = {}
+                resnum = int(line[22:26])
+                resname_three_letter = line[17:20]
+                if resnum not in coord_dic[chain]:
+                    coord_dic[chain][resnum] = AA_TO_CODE[resname_three_letter]
+    return coord_dic
+
+
+def get_interface_dict(pdb_f, cutoff=5.0):
```
mgiulini, line 24: does it make sense to re-use (adapt) the `identify_interface` method of the CAPRI class here?

src/haddock/modules-devel/design/optint/modules/fitness.py
```diff
@@ -0,0 +1,117 @@
...
+
+    return emscoring_pdb
+
+
+def calc_haddockscore(pdb):
+    """Calculate the haddock-score."""
+    topoaa_pdb_obj = run_topoaa(pdb)
+    emscoring_pdb_f = run_emscoring(topoaa_pdb_obj)
+    haddock_obj = HaddockModel(emscoring_pdb_f)
+    weights = {"w_vdw": 1.0, "w_elec": 0.2, "w_desolv": 1.0, "w_air": 0.1}
```
mgiulini, line 115: is it OK to consider the AIR component here?

mgiulini (approved):
hi! I left some minor comments.

PS: I guess that the file `examples/design-antibody-antigen/data/ambig.tbl` has to be removed now, correct?

I only have a glimpse on the genetic algorithm, I would love to try another class of optimizers to see if they can do better..that's why I am interested in the computational cost of the single mutation/rescoring process. I am going to do some tests now!

examples/thirdparty/design-antibody-antigen/design-ab-docking-w-optimization.cfg
rvhonorato: Done

src/haddock/modules-devel/design/optint/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+        output_model_list = run(input_pdb_l, self.params)
+
+        self.output_models = []
+        for element in output_model_list:
+            log.warning("#" * 80)
+            log.warning("#" * 80)
+            log.warning("#" * 80)
+            log.warning(
+                "THIS MODULE HAS NOT BEEN TESTED/VALIDATED DO NOT USE IT"
```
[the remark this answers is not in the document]
rvhonorato: Think its best to tell people not to use it altogether unless we (or someone) does some proper benchmark on it.

src/haddock/modules-devel/design/optint/defaults.yaml
```diff
@@ -0,0 +1,128 @@
...
+region:
+  default: []
+  type: list
+  minitems: 0
+  maxitems: 1000
+  title: Region
+  short: The region to be optimized
+  long: The region to be optimized
+  group: ""
+  explevel: guru
```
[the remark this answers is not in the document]
rvhonorato: It does yes, will add that

src/haddock/modules-devel/design/optint/modules/ga.py
```diff
@@ -0,0 +1,399 @@
...
+            for element in offspring:
+                # mutant = element[2]
+                if random.random() < self.mutpb:  # nosec
+                    self.toolbox.mutate_aa(element)
+                    del element[2].fitness.values
+
+            # evaluate only the mutated and crossed
+            invalid_ind = [ind for ind in offspring if not ind[2].fitness.valid]
+            try:
+                fitnesses = self.toolbox.map(self.toolbox.evaluate, invalid_ind)
```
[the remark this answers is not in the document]
rvhonorato: As long as `topoaa+emscoring`, interesting to know the timing actually, will measure it

src/haddock/modules-devel/design/optint/modules/pdb.py
rvhonorato: It does, but since `identify_interface` is in another module i thought its best to keep the modularity, but we could have a a new `libgeometry` with these functions

src/haddock/modules-devel/design/optint/modules/fitness.py
rvhonorato: It should always be 0, I left in there in case we eventually add some sort of restraints to the calculations

src/haddock/modules-devel/design/optint/modules/pdb.py
mgiulini: yep, we could port `identify_interface` to the existing library `libalign`, which does all the geometric stuff

joaomcteixeira:
> But if 0 lines were edited why is run_tests.py -b still needed?

In principle not. is just to play on the safe side.

src/haddock/modules/design/optint/defaults.yaml
joaomcteixeira: The module template now explains how to add a dictionary parameter. The `topoaa` also uses dictionary like parameters. However, try to use such parameter only if really needed as they are very hard to maintain in tests and everything else. Try to keep the config as flat as possible. :+1:

rvhonorato:
before merging this the module needs to be put in a `modules-devel`, `modulesdev`, `devel` category or something like that.

the parameter `region` needs to be converted into `region_` (similar to `resdic_`) and the code needs to be refactored to handle multiple target regions instead of only one.

marking this back as a draft until these are sorted out

rvhonorato:
Actually I was just checking some preliminary data and the results are quite strange. I'll close it and we may re-implement in another time once we've run some proper benchmarking.

PR #516, opened 2022-07-28
add `toml` support

rvhonorato:

- [X] You have stick to Python. Talk with us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [X] PR does not add any *install dependencies* unless permission granted by the HADDOCK team

---

This PR adds support for `.toml` file as configuration, if the `toml` python package is installed. This is needed to integrate HADDOCK3 in other tools such as the [@Ivre](https://github.com/i-VRESSE/workflow-builder).

Ping @sverhoeven


joaomcteixeira:
Giving users two different ways to provide configuration files might create confusion. I think it is better IVRESSE to write toml configs with the `.cfg` extension, and we could make the `config_reader` sort out the titles with the integer index, for example: `[rigidbody.1]`.

On the other hand, if at IVRESSE you can write the toml STRING to memory and remove those integer suffixes from the headings at the level of the string and only then save the text to a `cfg` file, that would also solve this problem. Is that possible?

rvhonorato:
This is not only about IVRESSE, there are also other software in development that will integrate haddock3. The `toml` format that can be generated by many languages and can greatly help integration.

joaomcteixeira:
Okay. Currently, our `config_reader` solves the following issues with `toml` related to haddock3:

1. repeated header names. The `config_reader` don't need them. I think this is the most important one.
2. the `config_reader` can read `None` or `none` values, and `nulls`. `toml` can't
3. We have already implemented many internal value conversion formats in the config reader. For example, we have a list of special keys that are automatically converted to `Path`, and we also have the `EmptyPath` object. 
4. With our `config_reader` we can easily add any specificities we may need in the future.

First, how to address the above issues with `toml`? We can try to think of a wrapper function somehow. For example, reading the `toml` plain string, replace the values that differ from the HADDOCK3 specifications, and then reading with the config reader. It is a bit of double work, but we facilitate the life of other projects.

Second, if the other projects write the configs to `toml`, we *can* read the `toml` files without problem, considering the types of values used in HADDOCK3. So, we don't need actually to use the `toml` lib, maybe only upgrading the `config_reader` to cope with the extra suffixes.

Projects that are written in python can use our config_reader and writer.

I am happy to help solve this issue. It is indeed a big one. The `config_reader` was developed because there was a need for it. And that need still exists today.

How can we solve this?

@amjjbonvin and @mgiulini join this conversation, please.

rvhonorato:
I've opened a separate issue for this alternative implementation and assigned you

rvhonorato:
Continue the discussion here: https://github.com/haddocking/haddock3/discussions/518

Issue #487, opened 2022-06-22
check values in cfg files against .yaml files   


mgiulini:
currently, values of parameters defined in modules of cfg files are not checked. For example, if there are have many `choices` for a parameter in the yaml file, the code never checks that the value in the cfg file is part of the list present in the corresponding `defaults.yaml` file.

example from clustrmsd:
```
criterion:
  default: "maxclust"
  type: "string"
  ...
  choices:
  - distance
  - maxclust
```
Discussed with @joaomcteixeira.

Related to #333

mgiulini:
The same is also true for the value type..This must be addressed soon, as currently the user can input whatever parameter type and no error/weird errors is raised. As an example, if the user sets irmsd = "false" in caprieval, the code proceeds and the irmsd is effectively calculated, while an error should be raised at the beginning.

mgiulini:
In the postprocessing issue #564 such type check is fundamental for being able to handle non-string types from the command line. I don't know whether/how the haddock-score CLI #510 handles this

PR #520, opened 2022-07-29
Add template to write new modules

joaomcteixeira:
You are about to submit a new Pull Request. Before continuing, make sure you read the [contributing guidelines](CONTRIBUTING.md), and you comply with the following criteria:

---

* Add a template explaining how to develop a new module.

Please let me know if it is clear or if any information is missing according to your recent experience. Thanks

### DIFF
--- a/src/haddock/modules/_template_cat/README.md
+++ b/src/haddock/modules/_template_cat/README.md
@@ -1,0 +1,30 @@
+Welcome!
+
+To develop your own HADDOCK3 module follow the patterns used in other modules.
+
+1. If your module belongs to a new category, create a folder for that category
+   under the `modules` folder.
+1. Else, create a folder for your module inside its relevant category.
+   The name of that folder is the name of the module, i.e., the name used to call
+   the module in the haddock3 configuration files and used throughout the code base.
+1. Copy the `__init__.py` file here to the new module's folder and edit it accordingly
+   to the instructions there.
+1. Do the same for the `defaults.yaml` file.
+1. You can then add any extra files needed inside your module's folder in order to
+   develop your module fully.
+1. If your module requires any extra libraries, describe how to install those libraries
+   in the `docs/INSTALL.md` file. Unless approved by the Haddock Team, do not add
+   those dependencies to the `requirements.*` files.
+1. HADDOCK3 has already many features related to IO, subprocess run, sending jobs,
+   etc. Please, look around the `libs` folder for pertinent functions, but, above all,
+   feel welcomed to reach out to us with any doubts.
+1. Please write also tests for your module. Our testing machinery already
+   tests for the common patterns, for example, inspecting the `defaults.yaml` file.
+   But you should write any additional tests to ensure your module work appropriately.
+   See other examples in the `tests/` folder.
+1. Finally, add an example of how to use your module in the `examples/` folder.
+   The example should have a short sampling scheme. Name the config file ending with
+   `-test.cfg`.
+
+Thanks, and we are happy to help you! :-)
+@ The Haddock Team
--- a/src/haddock/modules/_template_cat/README.md
+++ b/src/haddock/modules/_template_cat/README.md
@@ -20,7 +20,7 @@ To develop your own HADDOCK3 module follow the patterns used in other modules.
    feel welcomed to reach out to us with any doubts.
 1. Please write also tests for your module. Our testing machinery already
    tests for the common patterns, for example, inspecting the `defaults.yaml` file.
-   But you should write any additional tests to ensure your module work appropriately.
+   But you should write any additional tests to ensure that your module works properly.
    See other examples in the `tests/` folder.
 1. Finally, add an example of how to use your module in the `examples/` folder.
    The example should have a short sampling scheme. Name the config file ending with
--- a/src/haddock/modules/_template_cat/_template_mod/__init__.py
+++ b/src/haddock/modules/_template_cat/_template_mod/__init__.py
@@ -59,7 +59,8 @@ class HaddockModule(BaseHaddockModule):
 
         # if your module uses CNS you might need to define where the main CNS
         # script is localted. See examples in `topoaa`, `emref`.
-        cns_script = Path(RECIPE_PATH, "cns", "main.cns")
+        #  else leave it out.
+        # cns_script = Path(RECIPE_PATH, "cns", "main.cns")
 
         # use one of the following:
         # if the module does not use CNS:
--- a/src/haddock/modules/_template_cat/_template_mod/__init__.py
+++ b/src/haddock/modules/_template_cat/_template_mod/__init__.py
@@ -1,0 +1,103 @@
+"""
+Template module
+===============
+
+Use this template to implement your own HADDOCK3 module.
+
+In this docstring, please write the documentation explaining what the module
+does technically, scientifically, and its purpose.
+
+Anything you want to write to describe the module, also do it here.
+You should use restructureText syntax:
+
+    https://docutils.sourceforge.io/docs/user/rst/quickstart.html
+"""
+# Import here what you need
+from pathlib import Path
+
+# If your module does not use CNS, import the following
+from haddock.modules import BaseHaddockModule
+
+# If your module uses CNS import
+from haddock.modules.base_cns_module import BaseCNSModule
+
+
+# this is mandatory, don't erase nor edit these lines
+RECIPE_PATH = Path(__file__).resolve().parent
+DEFAULT_CONFIG = Path(RECIPE_PATH, "defaults.yaml")
+
+
+# this is the main class of the module. It should be named exactly as this.
+class HaddockModule(BaseHaddockModule):
+    # inherit from BaseCNSModule in case the module uses CNS
+    """
+    Here you can write any extra documentation you wish about the class.
+
+    Use numpydoc syntax:
+
+    https://numpydoc.readthedocs.io/en/latest/format.html
+    """
+
+    # this is mandatory
+    name = RECIPE_PATH.name
+
+    # this __init__ structure is mandatory, but you can extend it to the
+    # module's needs as long as you keep the main structure. Surely,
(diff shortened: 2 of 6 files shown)


src/haddock/modules/_template_cat/_template_mod/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        # Add here all the magic that your module does. You can split this part
+        # into many functions, classes, or add any extra files or subfolders to
+        # the module's folder as you may need. You can even import other modules
+        # if needed.
+
+        # If you module exports models, the line below is almost mandatory.
+        # see other modules, such as topoaa or mdref, emref to see how they
+        # handle `output_models`.
+        self.output_models = list_of_created_models
+        self.export_output_models(faulty_tolerance=self.params["tolerance"])
```
rvhonorato, line 97: I found out that this will fail if there's no`tolerance` in the the yaml file

src/haddock/modules/_template_cat/_template_mod/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # Add here all the magic that your module does. You can split this part
+        # into many functions, classes, or add any extra files or subfolders to
+        # the module's folder as you may need. You can even import other modules
+        # if needed.
+
+        # If you module exports models, the line below is almost mandatory.
+        # see other modules, such as topoaa or mdref, emref to see how they
+        # handle `output_models`.
+        self.output_models = list_of_created_models
```
rvhonorato, line 96: this `list_of_created_models` needs to be all of PDBObjects

rvhonorato, line 96:
```suggestion
        output_model_list = your_function(models_to_use, self.params)
        # output_model_list = [(pdb, psf, score), ...]
        list_of_created_models = []
        for element in output_model_list:
            pdb, psf, score = element
            # IMPORTANT: pass `file_name=Path.name`
            pdb_object = PDBFile(
                Path(pdb).name,
                topology=TopologyFile(
                    Path(psf).name, path="."
                    ),
                path=".")
            pdb_object.score = score
            list_of_created_models.append(pdb_object)
            
        self.output_models = list_of_created_models
```

src/haddock/modules/_template_cat/_template_mod/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        super().__init__(order, path, initial_params, cns_script=cns_script)
+
+    @classmethod
+    def confirm_installation(cls):
+        """Confirm if the module is ready to use."""
+        # here, you should write any code needed to confirm that all the
+        # dependencies required by your module are installed.
+        # this class method will be executed when HADDOCK3 starts.
+
+        # if you module does not import any run-time dependency, just leave
```
rvhonorato, line 74: what happens if you put an `import` here that tries to import a dependency that is not installed, will it handle the error?

src/haddock/modules/_template_cat/_template_mod/__init__.py
joaomcteixeira: Done. Thanks

src/haddock/modules/_template_cat/_template_mod/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+            order,
+            path,
+            *ignore,
+            init_params=DEFAULT_CONFIG,
+            **everything,
+            ):
+
+        # if your module uses CNS you might need to define where the main CNS
+        # script is localted. See examples in `topoaa`, `emref`.
+        cns_script = Path(RECIPE_PATH, "cns", "main.cns")
```
rvhonorato, line 58:
```suggestion
        # if your module uses CNS you might need to define where the main CNS
        # script is localted. See examples in `topoaa`, `emref`.
        #  else leave it out.
        # cns_script = Path(RECIPE_PATH, "cns", "main.cns")
```

src/haddock/modules/_template_cat/_template_mod/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+
+    # here is where the module magic will happen
+    def _run(self):
+        # you can refer to other modules as examples to see how they perform.
+        # Likely, the new module will need to load the output from the previous
+        # module. For example:
+        models_to_use = self.previous_io.retrieve_models()
+
+        # once the module loads, you can access its parameters in
+        # `self.params`. See more in the BaseHaddockModule class.
```
rvhonorato, line 86:
```suggestion
        # once the module loads, you can access its parameters in
        # `self.params`. See more in the BaseHaddockModule class.
        #  `self.params` is a dictionary that contains both the general
        #  parameters such as the number of processors 
        #  `self.params['ncores']` parameters as well as those 
        #  defined in your `defaults.yml`
```

rvhonorato:
Thank you **very much** for this (:

src/haddock/modules/_template_cat/_template_mod/__init__.py
joaomcteixeira: I am not sure if this is exactly mandatory. For example, `emref` uses the `libs.libcns.prepare_expected_pdb` function. I think that is more general and can be used here. What do you think?

src/haddock/modules/_template_cat/_template_mod/__init__.py
joaomcteixeira: As well as those defined by the user in case the module runs in the workflow. :+1:

src/haddock/modules/_template_cat/_template_mod/__init__.py
rvhonorato: `prepare_expected_pdb` generates a new `PDBFile` with a different name based on a current pdb object

src/haddock/modules/_template_cat/_template_mod/__init__.py
joaomcteixeira: Addressed, thanks.

src/haddock/modules/_template_cat/_template_mod/__init__.py
joaomcteixeira: Hmm. But I think it is possible to replace the this block:

```python
            pdb, psf, score = element
            # IMPORTANT: pass `file_name=Path.name`
            pdb_object = PDBFile(
                Path(pdb).name,
                topology=TopologyFile(
                    Path(psf).name, path="."
                    ),
                path=".")
```

for the `prepare_expected_pdb` function. But okay, we can add that, and also refer to the `prepare_expected_pdb` function.

src/haddock/modules/_template_cat/_template_mod/__init__.py
joaomcteixeira: Done.

joaomcteixeira:
thanks for the review. I have addressed all the issues.

rvhonorato (approved):
This template should greatly help new developers to add new modules - most likely for the majority of modules nothing will needs to be changed in the core python shell and this template clarifies where things should be added in the module context.

👍🏽

src/haddock/modules/_template_cat/_template_mod/__init__.py
mgiulini: I like the chek @rvhonorato implemented in the optint PR #515 . Maybe we can copy it here? commented out of course

src/haddock/modules/_template_cat/README.md
```diff
@@ -0,0 +1,30 @@
...
+   develop your module fully.
+1. If your module requires any extra libraries, describe how to install those libraries
+   in the `docs/INSTALL.md` file. Unless approved by the Haddock Team, do not add
+   those dependencies to the `requirements.*` files.
+1. HADDOCK3 has already many features related to IO, subprocess run, sending jobs,
+   etc. Please, look around the `libs` folder for pertinent functions, but, above all,
+   feel welcomed to reach out to us with any doubts.
+1. Please write also tests for your module. Our testing machinery already
+   tests for the common patterns, for example, inspecting the `defaults.yaml` file.
+   But you should write any additional tests to ensure your module work appropriately.
```
mgiulini, line 23:
```suggestion
   But you should write any additional tests to ensure that your module works properly.
```

mgiulini (approved):

mgiulini:
Good addition indeed, it's going to be useful!

A curiosity: does the check on the `choices` in the yaml file work?

joaomcteixeira:
> A curiosity: does the check on the choices in the yaml file work?

not yet: https://github.com/haddocking/haddock3/issues/487

> I like the chek @rvhonorato implemented in the optint PR https://github.com/haddocking/haddock3/pull/515 . Maybe we can copy it here? commented out of course

what is this?

sverhoeven:
Works for me, the variables `from haddock.modules import modules_category, category_hierarchy` dont list the template module and category so my [script](https://github.com/i-VRESSE/workflow-builder/blob/main/packages/haddock3_catalog/generate_haddock3_catalog.py) also does not list them.

joaomcteixeira:
> Works for me, the variables from haddock.modules import modules_category, category_hierarchy dont list the template module and category so my [script](https://github.com/i-VRESSE/workflow-builder/blob/main/packages/haddock3_catalog/generate_haddock3_catalog.py) also does not list them.

Yes, that was my intention. Do you think the template should be listed? I think not because otherwise, users could use it in the workflow, and that makes no sense. Let me know. Cheers.

sverhoeven:
> > Works for me, the variables from haddock.modules import modules_category, category_hierarchy dont list the template module and category so my [script](https://github.com/i-VRESSE/workflow-builder/blob/main/packages/haddock3_catalog/generate_haddock3_catalog.py) also does not list them.
> 
> Yes, that was my intention. Do you think the template should be listed? I think not because otherwise, users could use it in the workflow, and that makes no sense. Let me know. Cheers.

No it should not be listed, the template is for module contributors, not for module consumers.

joaomcteixeira:
Perfect :+1: Thanks!

PR #524, opened 2022-08-02
Updates docs theme to Karma theme

joaomcteixeira:


amjjbonvin (approved):
Actually quite nice - and better than readable which I just tested locally

PR #527, opened 2022-08-02
Uses action to deploy pages

joaomcteixeira:
Finally, found this to deploy the pages in a clean way: https://github.com/s0/git-publish-subdir-action#when-pushed-to-master-push-publicsite-to-the-gh-pages-branch-on-the-same-repo

Once @amjjbonvin updates the token, it should work. :crossed_fingers:


amjjbonvin (approved):
Anything needed from my side? What about this token update?

PR #539, opened 2022-08-07
Add full support to `toml`

### LINKED ISSUE #538: Add full support to toml (and dependency)
Finally,

We can use toml to read/write the configs if we create an intermediate step where the user config files are edited (internally) to meet the toml specifications. For example:

1. reading: haddock3 reads the user config, edits headers and other stuff to meet toml and the toml reads that string.
1. writing: toml writes the config string from the dictionary, and haddock does the final edits to meet the user config specifications.

I think this way is the more general one that could be applied in every language. I am pasting below again the specifications the user config is now differing from toml:

1. repeated header names. The config_reader don't need them. I think this is the most important one.
1. the config_reader can read None or none values, and nulls. toml can't
1. We have already implemented many internal value conversion formats in the config reader. For example, we have a list of special keys that are automatically converted to Path, and we also have the EmptyPath object.

_Originally posted by @joaomcteixeira in https://github.com/haddocking/haddock3/discussions/518#discussioncomment-3295686_

I am addressing this issue, it will be done shortly. When it is done, HADDOCK3 will fully support toml **while** keeping compatibility **also** with the current config files. Inevitably, `toml` will become a dependency, but that's ceased to be a problem as not depending on `toml` creates higher problems with other projects integrating haddoc3.

joaomcteixeira:

- [x] PR adds `toml` as dependency

---

* Add full support to `toml`
* Keep supporting the `cfg` files (hence, both formats are valid)
* Add feature to write `cfg` **and** pure `toml` files
* Uses `toml` as a mandatory dependency
* Closes #538
* Addresses #518
* supersedes #473

**Relevant notes:**

The following points were possible with the original haddock3 config reader but aren't possible with `toml`. However, they are also not mandatory for HADDOCK3.

* `toml` doesn't accept `None` values
* `toml` needs lists of the same type: `[10, 12.1]` is not accepted
* `toml` needs floats to be fully specified: `.1` is not accepted
* `toml` requires only lower case booleans

@rvhonorato @sverhoeven @amjjbonvin


rvhonorato (approved):

requirements.txt
```diff
@@ -5,4 +5,5 @@ jsonpickle==2.1.0
 numpy==1.22.2
 pyyaml==6.0
 scipy==1.8.0
+toml
```
mgiulini, line 8: shouldn't we have a version here?

mgiulini (approved):
looks OK! one remark: maybe let's add toml version to the requirements txt and yml?

requirements.txt
joaomcteixeira: Yes. It is definitively safer to increase the version of the dependencies manually. Thanks

joaomcteixeira:
All tests and pipelines pass. Ready to be merged

PR #543, opened 2022-08-29
changed retrieve_models logic

### LINKED ISSUE #542: change logic in retrieve_models
In order to run cns modules (such as `flexref` and `emref`) after `topoaa` in the middle of the workflow it is necessary to change a bit the logic of `retrieve_models` method of `ModuleIO` : https://github.com/haddocking/haddock3/blob/7164a41d5acc82fb9aaee3643927943fce6a7b2f/src/haddock/libs/libontology.py#L133

In particular, when flexref or emref are executed after a topoaa module placed in the middle of the workflow (for example after the openmm module #482 ) the program enters this clause
https://github.com/haddocking/haddock3/blob/7164a41d5acc82fb9aaee3643927943fce6a7b2f/src/haddock/libs/libontology.py#L148
and the list of models is created according to
https://github.com/haddocking/haddock3/blob/7164a41d5acc82fb9aaee3643927943fce6a7b2f/src/haddock/libs/libontology.py#L158
which is not correct in this scenario, as it creates a tuple instead of the usual list of PDBFile objects.

mgiulini:

- [ ] Your PR is about CNS

---

Closes #542. Tests and examples run as usual, but I don't know the reason why the method was developed like that in the first place, please tell me if I missed a point.

PS: a build test failed for no apparent reason. I don't think it's critical.

### DIFF
--- a/src/haddock/libs/libontology.py
+++ b/src/haddock/libs/libontology.py
@@ -155,7 +155,7 @@ class ModuleIO:
                 raise Exception(_msg)
 
             # prepare pairwise combinations
-            model_list = [values for values in zip(*input_dic.values())]
+            model_list = [pdb[0] for pdb in input_dic.values()]
         elif input_dic and crossdock and not individualize:
             model_list = [
                 values for values in itertools.product(*input_dic.values())


joaomcteixeira (approved):
Nice catch. From what I understood, there was an `if-` clause that was executed only when `flexref` (and others) were placed right after `topoaa`. However, in this case the logic to retrieve the models was wrong because it created a list of tuples instead of a list of PDB objects. So, this was actually a hidden bug that becomes visible now when developing the `openmm` module because this feature was actually never used in the regular pipelines. Is that it?

If all the pipelines work and the results are not affected (capri tables, for example), good to go! :+1:

mgiulini:
That's precisely what happened! The list of tuples was not supposed to be there, right?
Everything seems to run smoothly now, and also the openmm #482 PR should benefit from this as you correctly pointed out.

PR #551, opened 2022-09-09
Add SAXS scoring module

cvnoort:

- [ ] Your PR is about CNS

---

Introducing a new scoring module to calculate a HADDOCK-SAXS score as described by Karaca & Bonvin (2013) in Acta Crystallographica.


cvnoort:
TO DO: Use HADDOCK score from models (calculated by the preceding module), rather than calling calc_haddock_score to calculate from scratch. We then no longer need the individual weights for the scoring terms, replace these with one weight for the pre-calculated HADDOCK score (default = 1) and of course keep w_saxs for the SAXS term (default = 50).

cvnoort:
As discussed, the `saxsscore` module is a `BaseHaddockModule`, rather than a `ScoringModule`. 

This is because the latter is a child class of `BaseCNSModule` that is designed to use CNS, which we do not need for this module. Instead of calling the `output` function from the `BaseCNSModule` class, some of it is copied in the module itself to produce a similar output table...

PR #562, opened 2022-10-05
removed absolute paths in topoaa

### LINKED ISSUE #561: reduce filename length in docking-protein-peptide
one of the molecules in the example `docking-protein-peptide` (`DAIDALSSDFT_3conformations.pdb`) has a quite long filename, which may cause troubles to cns (exceeding maximum filename length) if the haddock3 installation is a bit nested.

mgiulini:

- [ ] Your PR is about CNS

---

Closes #561 by reading the list of pdb files here https://github.com/haddocking/haddock3/blob/aa68fd3abf3ebe9e3c7c34812639761c7be9d69c/src/haddock/libs/libpdb.py#L256 always from the current directory, thus eliminating the path dependence and the potentially problematic usage of absolute paths.

### DIFF
--- a/src/haddock/libs/libpdb.py
+++ b/src/haddock/libs/libpdb.py
@@ -226,9 +226,9 @@ def get_new_models(pdb_file_path):
     return new_models
 
 
-def get_pdb_file_suffix_variations(file_name, path=None, sep="_"):
+def get_pdb_file_suffix_variations(file_name, sep="_"):
     """
-    List suffix variations of a PDB file.
+    List suffix variations of a PDB file in the current path.
 
     If `file.pdb` is given, and files `file_1.pdb`, `file_2.pdb`, exist
     in the folder, those will be listed.
@@ -237,9 +237,6 @@ def get_pdb_file_suffix_variations(file_name, path=None, sep="_"):
     ----------
     file_name : str or Path
         The name of the file with extension.
-
-    path : str or pathlib.Path
-        Path pointing to a directory where to perform the search.
 
     sep : str
         The separation between the file base name and the suffix.
@@ -252,8 +249,7 @@ def get_pdb_file_suffix_variations(file_name, path=None, sep="_"):
         If no files are found return an empty list.
     """
     basename = Path(file_name)
-    path = path or Path.cwd()
-    return list(path.glob(f"{basename.stem}{sep}*{basename.suffix}"))
+    return list(Path(".").glob(f"{basename.stem}{sep}*{basename.suffix}"))
 
 
 def read_RECORD_section(lines, section_slice, func=set):


amjjbonvin (approved):
All tests runs passing?

joaomcteixeira:
Hi,

I need some time to have a look at this carefully, even if they are just a few lines of code. Be mindful with tests here because changing from relative to absolute, and *vice-versa*, is a very tricky change and can bring hidden bugs. HADDOCK3 have some modes that interchange absolute to relative, for example, the `self_contained`flag.
I am not saying the code is wrong. Simply, give extra care to this PR.

Big cheers,

mgiulini:
> 

Yes!

mgiulini:
> All tests runs passing?

Yes!

mgiulini:
> Hi,
> 
> I need some time to have a look at this carefully, even if they are just a few lines of code. Be mindful with tests here because changing from relative to absolute, and _vice-versa_, is a very tricky change and can bring hidden bugs. HADDOCK3 have some modes that interchange absolute to relative, for example, the `self_contained`flag. I am not saying the code is wrong. Simply, give extra care to this PR.
> 
> Big cheers,

Sure, take your time! I agree that these changes are quite tricky, and this is why I would make this part of the code a bit less nested to facilitate the comprehension, as it took me a while to figure out where the problem was. For example, I would remove the function `get_result_or_same_in_list`, as it is used just once. Let me know what you think after you go through this change, cheers!

rvhonorato (approved):

src/haddock/libs/libpdb.py
```diff
@@ -252,8 +249,7 @@ def get_pdb_file_suffix_variations(file_name, path=None, sep="_"):
         If no files are found return an empty list.
     """
     basename = Path(file_name)
-    path = path or Path.cwd()
-    return list(path.glob(f"{basename.stem}{sep}*{basename.suffix}"))
+    return list(Path(".").glob(f"{basename.stem}{sep}*{basename.suffix}"))
```
joaomcteixeira, line 252: Sorry, I am struggling to get free time these weeks. Some comments follow:

1. If we need to change this function specifically for `topoaa`, could it be safer to make a new function only for `topoaa`? In order to have one that returns absolute paths and another that returns relative paths'. I remember editing this function some time ago to get some functionalities; that `path=None` is not innocent. I believe.
2. functions like `get_result_or_same_in_list` and `transform_to_list` are likely impossible to remove. Because, in many cases, we can have arguments that are strings or lists. For example, `molecules=path` or molecules=[path1, path2]`; yet the functionality needs to be maintained. Yes, we can use the unpacking in python functions (`*`), but it is troublesome to handle that all around the code. I can say with almost 100% sure that if the `get_result_or_same_in_list` is because there are some situations where it was needed.

joaomcteixeira: Yes, I see. Perfect! I believe that part of the code can be improved and cleaned. Especially after `pdb-tools` lib-like functionalities were developed afterwards. Go ahead with this change. I see there's confidence; I just wanted to call attention when changing from relative to absolute paths. Because HADDOCK3 is meant to run locally and on grid controlling paths becomes tricky also because of the need to be moving the CWD across steps as they run.

:+1: merge! :wink:

joaomcteixeira (approved):
I left a comment in the thread above :+1:

mgiulini:
> I left a comment in the thread above 👍

Thanks for the feedback @joaomcteixeira :)

PR #585, opened 2022-11-21
Capri memory error

### LINKED ISSUE #584: correct memory error in caprieval
Currently, a tedious error can occur in the caprieval module when multiple capri jobs are executed in parallel.

mgiulini:

- [ ] Your PR is about CNS

Closes #584 

The memory error was due to the parallel execution of multiple capri jobs. Apparently, calling the `get_intermolecular_contacts` function can lead to crashes in CAPRI jobs. Additionally, it was found that the same function is sometimes not 100% precise in retrieving the correct contacts.

This PR addresses this problems by:

1. creating a haddock3-specific, memory-friendly `load_contacts` function https://github.com/haddocking/haddock3/blob/f58ccb1d1238ba8c88ee3ad308be463a150b27a2/src/haddock/modules/analysis/caprieval/capri.py#L27-L75 : this function iterates over all the atoms of a reference chain, retrieving the contacts that the atom forms with the target chain. It is slower than the original implementation, but more accurate and less memory-demanding (a contact matrix is not needed anymore)
2. giving the possibility to retrieve the full set of heavy atoms of a protein or DNA molecule (useful for calculating contacts) https://github.com/haddocking/haddock3/blob/f58ccb1d1238ba8c88ee3ad308be463a150b27a2/src/haddock/libs/libalign.py#L65-L106
3. introducing an additional test for caprieval, where two heterogeneous complexes are compared. The values of irmsd and fnat obtained with the previous routines were slightly wrong, while the current ones have been compared to those of another software


src/haddock/libs/libalign.py
```diff
@@ -61,6 +61,30 @@
 DNA_RES = ["DA", "DC", "DT", "DG"]
```
amjjbonvin, line 61: Do you also have RNA defined?

amjjbonvin (approved):

src/haddock/libs/libalign.py
mgiulini: > Do you also have RNA defined?

not explicitly, but the function should use by default all the heavy atoms whenever the structure is not dna or protein

mgiulini: worth to test it by the way!

amjjbonvin:
DNA or RNA is almost the same - I would make it then nucleix acid and add the rna bases to it

=================================
 Prof. A.M.J.J. Bonvin | Faculty of Science
 Scientific Director Bijvoet Centre
 Utrecht University | bonvinlab.org 
 ***@***.*** | T: +31 30 2533859 
=================================

rvhonorato (approved):
looks good to me

PR #594, opened 2022-12-01
Compress `.inp`, `.out`, and `.seed` in CNS Jobs on the fly

### LINKED ISSUE #587: The clean option should only keep one out file and not package all into one archive

We can save significant space by not archiving all out files of a module into a single `out.gz` file, but rather only compressing the first one and deleting the others.

Instead of compressing and packaging all out files into one archive, delete all but one (e.g. keep the 1st model) and gzip that file.

### LINKED ISSUE #593: haddock3-unpack should have an option to only  unpack the relevant files
It does not make much sense to unpack the `.out` and `.inp` files... So the default behaviour should be to unpack only the PDB files (and any file required for the workflow to work).  

An option `-a` could be added to unpack everything if requested.

joaomcteixeira:

- [ ] Your PR is about CNS

---

Closes #587
Closes #593

@amjjbonvin can you test this? In case you don't yet have my fork on your clone:

```
git remote add joaot https://github.com/joaomcteixeira/haddock3
git fetch joaot
git checkout unpackmore
```

And test some example.  Let me know if you with more to be done regarding **on the fly packing of output files**.

Cheers!

### DIFF
--- a/src/haddock/libs/libworkflow.py
+++ b/src/haddock/libs/libworkflow.py
@@ -26,7 +26,7 @@ class WorkflowManager:
         # terminate is used to synchronize the `clean` option with the
         # `exit` module. If the `exit` module is removed in the future,
         # you can also remove and clean the `terminate` part here.
-        self._terminated = 0
+        self._terminated = None
 
     def run(self):
         """High level workflow composer."""
--- a/docs/clients/cliclean.rst
+++ b/docs/clients/cliclean.rst
@@ -1,4 +1,4 @@
-Clean output Client
+Clean HADDOCK3 runs
 ===================
 
 .. argparse::
--- a/docs/clients/cliunpack.rst
+++ b/docs/clients/cliunpack.rst
@@ -1,5 +1,5 @@
-Unpack output
-=============
+Unpack HADDOCK3 run
+===================
 
 .. argparse::
    :module: haddock.clis.cli_unpack
--- a/src/haddock/clis/cli.py
+++ b/src/haddock/clis/cli.py
@@ -163,6 +163,7 @@ def main(
         restart_step = restart
         WorkflowManager_ = WorkflowManager
 
+    print(params)
     with (
             working_directory(other_params[RUNDIR]),
             log_error_and_exit(),
--- a/src/haddock/clis/cli.py
+++ b/src/haddock/clis/cli.py
@@ -163,7 +163,6 @@ def main(
         restart_step = restart
         WorkflowManager_ = WorkflowManager
 
-    print(params)
     with (
             working_directory(other_params[RUNDIR]),
             log_error_and_exit(),
--- a/src/haddock/clis/cli_clean.py
+++ b/src/haddock/clis/cli_clean.py
@@ -3,14 +3,16 @@
 Clean the output of an HADDOCK3 run directory.
 
 The clean process performs file archiving and file compressing
-operations. File with extension `seed`, `inp`, `out`, and `con` are
-compressed and archived into `.tgz` files. While files with `.pdb` and
-`.psf` extension are compressed to `.gz` files. The original files are
-deleted.
+operations.
+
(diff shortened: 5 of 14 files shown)


amjjbonvin (changes requested):
No need to compress on the fly the `.inp` and `.seed` files - those a peanuts in terms of size. 
Best to only archive those with the clean command.

The `.out` files are on contrary very large and compressing them will save a lot of space.

joaomcteixeira:
It also takes peanuts to compress those. I can make that an internal option so that each module can operate as desired.

amjjbonvin:
> It also takes peanuts to compress those. I can make that an internal option so that each module can operate as desired.

But it is not worth it... Ok to archive those all into one gzipped archive. Not worth compressing them individually.

joaomcteixeira:
Okay. Done. Now it's an option with default to `False`. **Only** compresses `.out` files. But I need to see now how the `clean` option would work. Let's make that another PR?

amjjbonvin:
I guess the clean option should archive any `.out` and `.out.gz` files

mgiulini:
Nice work! I tested it on the protein-protein-docking-test scenario and the folder size dropped from 30M to 18M.
I would set the `compress_inp` and `compress_seed` options to `True` by default, thus compressing inp and seed files into two archives by default. I think this can be useful for a better readability and faster listing of the important files (thinking about high sampling).

amjjbonvin:
> I would set the compress_inp and compress_seed options to True by default, thus compressing inp and seed files into two archives by default.
> 
The seed files are needed by the next step, i.e. not a good idea.

ing files could be archived by default, but it does not make sense to gzip each individual file for those.

The out files are a different issue. It is best to gzip those on the fly for each completed model.
And at the end if clean is true archive them into one archive

mgiulini:
> I would set the compress_inp and compress_seed options to True by default, thus compressing inp and seed files into two archives by default.
> The seed files are needed by the next step, i.e. not a good idea. ing files could be archived by default, but it does not make sense to gzip each individual file for those. The out files are a different issue. It is best to gzip those on the fly for each completed model. And at the end if clean is true archive them into one archive

makes sense!

joaomcteixeira:
Okay, will do

joaomcteixeira:
Okay. I am in the making of it. Need to review some of the tests and some parts.

joaomcteixeira:
Hi.

Implemented all requirements in #587 and #593. Finishing some tests on the `*-test.cfg`. I will let you know when they are done. :+1:

amjjbonvin (approved):

amjjbonvin:
Just to confirm: are the out files gzipped first on the fly for each completed model? Or is the cleaning only occurring at the end of a module / run?

To save space while running, compression on the fly of the out files is a must (actually irrespective of the clean options)

joaomcteixeira:
> Just to confirm: are the out files gzipped first on the fly for each completed model? 

Yes. Always.

Plus, the `clean`, if `True`, will further compress the other files and delete all but the first for `inp` and `out`. I updated the docs also.

But I am having  some error in the `test` I will try to sort it out soon. Don't merge yet.

joaomcteixeira:
The tests ended, and I tried also the `--restart` and `haddock3-copy` commands, and were working. I think it's safe to merge. You can test it further locally. Let me know.
Cheers!

src/haddock/libs/libworkflow.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, workflow_params, start=0, **other_params):
         # terminate is used to synchronize the `clean` option with the
         # `exit` module. If the `exit` module is removed in the future,
         # you can also remove and clean the `terminate` part here.
-        self._terminated = 0
+        self._terminated = None
```
joaomcteixeira, line 29: Note for the future. This was one of the most tricky parts of this implementation.

amjjbonvin (approved):

PR #596, opened 2022-12-08
Validate ambigs

### LINKED ISSUE #595: check for the existence of restraints defined in tbl files
it is possible that one or more distance restraints defined in .tbl files do not exist, namely at least one of the two residues does not exist in the reference structure.

It would be nice to implement a python-based check such that the execution

- **stops** if all the restraints do not exist
- **raises a warning** if at least one restraint cannot be defined

Where to place it? at the beginning of the workflow? before each CNS execution?

mgiulini:

- [ ] Your PR is about CNS

---

Closes #595

### DIFF
--- a/src/haddock/libs/librestraints.py
+++ b/src/haddock/libs/librestraints.py
@@ -36,7 +36,7 @@ def validate_ambig_fname(ambig_fname, models):
         not_found += 1
     print(f"found {found} vs not_found {not_found} restraints")
     if found == 0:
-        raise Exception("No valid restraints are available. Aborting")
+        raise Exception(f"No valid restraints are available for models {models} in {ambig_fname}. Aborting")
     print(f"time for validating{time.time() - st_time}")
     return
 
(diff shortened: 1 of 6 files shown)


joaomcteixeira (commented):
Quick comment:
I suggest this operation is performed during the `prepare run` phase, such that errors are identified before starting any calculations. Having these tests at the level of the modules as they are now is also important. I think both types of validations are needed.

mgiulini:
Hi @joaomcteixeira , thanks for the feedback! It would be indeed nice, but there are a few caveats that make this not viable to me:

- with PR #563 a set of ambig files can be provided in an archive. Not easy to inspect those at the `prepare_run` level
- the possible diversity of the input models implies that we have to inspect all pdb files to evaluate the consistency of restraints. again, the `prepare_run` level does not seem to be the right place for this
Wrapping up: it would be nice, but it does not seem possible to me
PS: note this PR is not yet ready as one of the examples is failing

PR #606, opened 2023-01-11
added png option

### LINKED ISSUE #601: implement png option in postprocessing
#567 implements the postprocessing. This analysis should be extended so that if the user provides a `--png ` option, the current plots are rendered in png format with a higher quality.

mgiulini:

- [ ] Your PR is about CNS

---

Closes #601 

Now the user can get static png scatter plots and box plots by adding the `--png` flag while running `haddock3-analyse`

Example:
`haddock3-analyse -r examples/docking-antibody-antigen/run1-ranairCDR-cltsel-test/ -m 5 --png=True
`

### DIFF
--- a/src/haddock/libs/libplots.py
+++ b/src/haddock/libs/libplots.py
@@ -162,6 +162,8 @@ def box_plots(capri_filename, cl_ranking, png):
             colors = plt.cm.tab20.colors
             for patch, color in zip(bplot['boxes'], colors):
                 patch.set_facecolor(color)
+            for median, color in zip(bplot['medians'], colors):
+                median.set_color(color)
             # plot details
             ax.legend(bplot["boxes"], legend_labels, bbox_to_anchor=(1.0, 1.0))
             plt.ylabel(AXIS_NAMES[x_ax])
@@ -364,7 +366,7 @@ def scatter_plots_png(capri_filename, cl_ranking):
         plt_fname = f"{x_ax}_{y_ax}.png"
         plt.tight_layout()
         plt.savefig(plt_fname, dpi=200)
-        plt.close()
+
         # creating full picture
         if not gb_other.empty:
             plt.scatter(
@@ -381,4 +383,5 @@ def scatter_plots_png(capri_filename, cl_ranking):
             plt_fname = f"{x_ax}_{y_ax}_full.png"
             plt.tight_layout()
             plt.savefig(plt_fname, dpi=200)
-            plt.close()
+        plt.close()
+        
--- a/src/haddock/libs/libplots.py
+++ b/src/haddock/libs/libplots.py
@@ -83,7 +83,188 @@ def read_capri_table(capri_filename, comment="#"):
     return capri_df
 
 
-def box_plots(capri_filename, cl_ranking, png, dpi):
+def in_capri(column, df_columns):
+    """
+    Check if the selected column is in the set of available columns.
+
+    Parameters
+    ----------
+    column : str
+        column name
+    df_columns : pandas.DataFrame.columns
+        columns of a pandas.DataFrame
+    Returns
+    -------
+    resp : bool
+        if True, the column is present
+    """
+    resp = True
+    if column not in df_columns:
+        log.warning(f"quantity {column} not present in capri table")
+        resp = False
+    return resp
+
+
+def update_layout_plotly(fig, x_label, y_label, title=None):
+    """
+    Update layout of plotly plot.
+
+    Parameters
+    ----------
+    fig : plotly Figure
+        figure
+    x_label : str
+        x axis name
+    y_label : str
+        y axis name
+    title : str or None
+        plot title
+    """
+    px_dict = {
+        "title": title,
+        "xaxis": dict(
+            title=x_label,
+            tickfont_size=14,
+            titlefont_size=40,
+            ),
+        "yaxis": dict(
+            title=y_label,
+            tickfont_size=14,
+            titlefont_size=40,
+            ),
+        "legend": dict(x=1.01, y=1.0, font_family="Helvetica", font_size=16),
+        "hoverlabel": dict(font_size=16, font_family="Helvetica")
+        }
+    fig.update_layout(px_dict)
+    return
+
+
+def update_layout_matplotlib(x_label, y_label):
+    """
+    Update layout of matplotlib plot.
+
+    Parameters
+    ----------
+    x_label : str
+        x axis name
+    y_label : str
+        y axis name
+    """
+    plt.title(f"{x_label} vs {y_label}", fontsize=40)
+    plt.xlabel(x_label, fontsize=40)
+    plt.ylabel(y_label, fontsize=40)
+    plt.xticks(fontsize=20)
+    plt.yticks(fontsize=20)
+    legend = plt.legend(bbox_to_anchor=(1.0, 1.0), fontsize=30)
+    for handle in legend.legendHandles:
+        handle.set_sizes([150])
+    plt.tight_layout()
+    return
+
+
+def box_plot_plotly(gb_full, y_ax):
+    """
+    Create a scatter plot in plotly.
+    
+    Parameters
+    ----------
+    gb_full : pandas DataFrame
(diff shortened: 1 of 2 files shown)


amjjbonvin (approved):
Looking good - resolution at 200 dpi - we can change it if needed.  For publication quality we would need 300 dpi (or an editable eps)

mgiulini:
> Looking good - resolution at 200 dpi - we can change it if needed. For publication quality we would need 300 dpi (or an editable eps)

thanks! would you make the resolution an input parameter?

something like this could work:
`haddock3-analyse -r examples/docking-antibody-antigen/run1-ranairCDR-cltsel-test/ -m 5 --png=300 `

amjjbonvin:
A good idea indeed

mgiulini:
done! at the end I defined it as another input option to avoid confusion 

Now the usage is
`haddock3-analyse -r examples/docking-antibody-antigen/run1-ranairCDR-cltsel-test/ -m 5 --png=True --dpi=300 `

if dpi is not specified, 200 is used by default

src/haddock/libs/libplots.py
```diff
@@ -138,8 +152,29 @@ def box_plots(capri_filename, cl_ranking):
             legend=dict(x=1.01, y=1.0, font_family="Helvetica", font_size=16),
             hoverlabel=dict(font_size=16, font_family="Helvetica"),
             )
-
         fig.write_html(px_fname, full_html=False, include_plotlyjs='cdn')
+        # create png boxplot if necessary
+        if png:
```
rvhonorato, line 157: also why this check here?

src/haddock/libs/libplots.py
```diff
@@ -281,3 +316,78 @@ def scatter_plots(capri_filename, cl_ranking):
             )
 
         fig.write_html(px_fname, full_html=False, include_plotlyjs='cdn')
+        
+
+def scatter_plots_png(capri_filename, cl_ranking, dpi):
+    """
+    Create scatter plots in png format.
+
+    analogous to scatter_plots, but here the plots are saved in png format.
```
rvhonorato, line 325: By looking at the functions `scatter_plots` and `scatter_plots_png` they indeed look very similar but it's unclear to me what's the actual difference between them; aka: why do you need a separate function for it?

src/haddock/libs/libplots.py
```diff
@@ -107,10 +112,19 @@ def box_plots(capri_filename, cl_ranking):
         else:
             cl_df["capri_rank"] = cl_ranking[cl_id]
             gb_good = pd.concat([gb_good, cl_df])
+            
     gb_other["cluster-id"] = "Other"
     gb_other["capri_rank"] = len(cl_ranking.keys()) + 1
     gb_cluster = pd.concat([gb_good, gb_other])
+    # when png is true we already order everything according to capri_rank
+    if png:
```
rvhonorato, line 120: Why do you need this check here?

src/haddock/libs/libplots.py
rvhonorato, line 157: Also why this check here?

rvhonorato (changes requested):
Having the option to save as `.png` is nice!

I have some comments because, in general, code to make plots tends to get very confusing very quickly. I see you added another function to handle the `.png` format and also some logic checks in other plotting functions, why is that? 

I'd suggest that you break down these functions even further, separating the ones that treat the data and adding some granularity into the ones that handle the plot (`make_labels`, `edit_ticks`, etc) - probably out of scope of this PR but definitively something we should look into sooner rather than later.

mgiulini:
> Having the option to save as `.png` is nice!
> 
> I have some comments because, in general, code to make plots tends to get very confusing very quickly. I see you added another function to handle the `.png` format and also some logic checks in other plotting functions, why is that?
> 
> I'd suggest that you break down these functions even further, separating the ones that treat the data and adding some granularity into the ones that handle the plot (`make_labels`, `edit_ticks`, etc) - probably out of scope of this PR but definitively something we should look into sooner rather than later.

Hi Rodrigo, thanks for the feedback! the functions are indeed quite long already. The way I thought about it was to have a pair of plotting functions that do the job (as they are not part of the full haddock3 pipeline), but since the postprocessing/analysis will probably grow with time it makes sense to break down the code as much as possible.

src/haddock/libs/libplots.py
mgiulini: unfortunately the specifics of plotly and matplotlib make an effective integration very hard. will try to come out with something prettier!

src/haddock/libs/libplots.py
mgiulini: it's a technicality: plotly can act directly on pandas dataframe, sorting them according to whatever order (column). For matplotlib that's not true as far as I know

src/haddock/libs/libplots.py
mgiulini: this checks if we have to create a matplotlib boxplot!

rvhonorato: I see... but can't you do the png with plotly? Check herer [`plotly.io.write_image`](https://plotly.github.io/plotly.py-docs/generated/plotly.io.write_image.html#plotly.io.write_image)

```python
if png:
    fig.write_image(file=px_fname, frormat="png")
else:
    fig.write_html(file=px_fname, full_html=False, include_plotlyjs='cdn')
```

mgiulini: yep, I know, but the output is quite ugly..not a viable option if you want to create almost-publication-ready figures

not to mention the fact that the figure itself must be tweaked in order to take into account for the static nature of the png (downsizing of scatter points, etc..)..matplotlib is much better for png images overall

mgiulini:
hi @rvhonorato check out the new library, code has been broken down in smaller functions. Now for box plots and scatter plots there's a handler function that takes care of the preprocessing. Then, plotly and matplotlib-specific functions take care of the plots.

Let me know your thoughts:)

src/haddock/libs/libplots.py
rvhonorato: Ah! If the point is to provide a figure that people can copy and paste it makes sense to make one that its actually useful for that purpose, all clear!

src/haddock/libs/libplots.py
```diff
@@ -83,7 +83,188 @@ def read_capri_table(capri_filename, comment="#"):
...
+    Update layout of matplotlib plot.
+
+    Parameters
+    ----------
+    x_label : str
+        x axis name
+    y_label : str
+        y axis name
+    """
+    plt.title(f"{x_label} vs {y_label}", fontsize=40)
```
rvhonorato, line 153: this `plt` is defined outside the function scope, you could pass it as an argument and also return it

```python
def update_layout_matplotlib(_plt, x_label, y_label):
    # do things
    return _plt

# somewhere else in the code
updated_plt = update_layout_matplotlib(plt, TITLE_NAMES[x_ax], TITLE_NAMES[y_ax])
```

something like this

src/haddock/libs/libplots.py
```diff
@@ -83,7 +83,188 @@ def read_capri_table(capri_filename, comment="#"):
...
+    Create a scatter plot in plotly.
+    
+    Parameters
+    ----------
+    gb_full : pandas DataFrame
+        data to box plot
+    y_ax : str
+        variable to plot
+    """
+    fig = px.box(gb_full,
```
rvhonorato, line 176: `px` is also outside the scope

rvhonorato (approved):
Some nitpicky things about the scope of the functions that could be improved, but its fine as it is now! great addition

src/haddock/libs/libplots.py
mgiulini: I don't know..since both `plt` and `px` are aliases to imported libraries, does it make sense to pass them to the functions?
From what I've seen they're always called directly from within functions to modify the current plot..

mgiulini:
> Some nitpicky things about the scope of the functions that could be improved, but its fine as it is now! great addition

thanks for the feedback!

PR #611, opened 2023-02-08
haddock3-analyse: create report html page based on individual plots

SarahAlidoost:

- [ ] ~Your PR is about CNS~
- [ ] ~You wrote tests for the new code~
- [ ] ~`tox` tests pass. *Run `tox` command inside the repository folder*~
- [ ] ~`-test.cfg` examples execute without errors. *Inside `examples/` run `python run_tests.py -b`*~
- [ ] ~Your PR is about writing documentation for already existing code :fire:~
- [ ] ~Your PR is about writing tests for already existing code :godmode:~

---

This PR extends the functionality of the command line `haddock3-analyse`. After running the analyse, a HTML file is generated that contains all of the plots (box and scatter plots). The HTML file name starts with `report` and can be found next to other HTML files. 

Examples:
[report_Box_plots_of_9_caprieval.html](https://github.com/i-VRESSE/haddock3/blob/create_report/examples/analysis/report_Box_plots_of_9_caprieval.html)
[report_Scatter_plots_of_9_caprieval.html](https://github.com/i-VRESSE/haddock3/blob/create_report/examples/analysis/report_Scatter_plots_of_9_caprieval.html)


SarahAlidoost:
@mgiulini can you please review this pull request? I couldn't add a reviewer to the **`Reviewers`** list.

SarahAlidoost:
> Hi @SarahAlidoost , thanks for the contribution, the report looks very nice!
> 
> I think though that this PR must be checked and tested a bit more thoroughly: I tried to run haddock3-analyse on one of the examples folders (`examples/docking-antibody-antigen/run1-ranairCDR-cltsel-test`), and it failed. In this case I used the modules 1, 5 and 13.
> 
> The reason why it failed on module 1 is clear (there's a typo in the column `model-cluster-ranking` in `capri_ss.tsv`) and will be corrected (see #626), while for the other two modules the situation is a bit more obscure (error message reads: `Length of values (80) does not match length of index (20)`)
> 
> can you try to test the addition on some examples to see if there are other problems?

Thanks for checking this. It fails on finding the best structures from `capri_ss` file when creating tables. The number of the best structures is **4** and hard coded in the module, see [here](https://github.com/i-VRESSE/haddock3/blob/e3d655436e96055e3bdc8a2e6a7a3d53d24a1d60/src/haddock/libs/libplots.py#L703). For `-m 5 ` looking at the `examples/docking-antibody-antigen/run1-ranairCDR-cltsel-test/05_caprieval/capri_ss.tsv`, it seems that the number of best structures should be 1. The same issue for `-m 13`.  Is the number of best structures defined by the user or the code? In the former case, we can introduce an argument to the cli and create a better error message. 

For `-m 15`, there are two more problems: the columns `cluster-ranking` and `model-cluster-ranking` are filled with `-`. Is it correct? Another problem is a typo in the column name, `self.model-cluster-ranking` should be `model-cluster-ranking`.

SarahAlidoost:
@mgiulini in my last commit, I fixed the issue for the number of best structures. Now, the `analyse` command should work for `-m 5` and` -m 13`. I also added a todo for `-m 2 and 15` i.e. `Unclustred` cases.

amjjbonvin:
Be careful with the number of best structures. This depends on the clustering settings and other parameters. 

Also the tests do not have the full sampling, which might explain some of the issue you are having in the analysis

src/haddock/libs/libplots.py
```diff
@@ -523,29 +561,317 @@ def scatter_plot_handler(capri_filename, cl_rank, png, dpi):
...
+    table_df = best_struct_df.copy()
+    for col_name in table_df.columns[1:]:
+        file_name = best_struct_df[col_name].apply(lambda row: Path(row).name)
+        # href empty for local files
+        # TODO fix links to Download and Visibility
+        # we can use html href syntax but html should be responsive!
+        dl_text = "Download"
+        vis_text = "Visibility"
+        table_df[col_name] = file_name + ", " + dl_text + ", " + vis_text
+    return table_df
```
mgiulini, line 757: I guess that this is the function that should create all the links to the structures and to their visualization..the problem here is that in a standard run we might have hundreds (if not thousands) of clusters, and including all the links here would probably end up in a huge html report, what do you think?

src/haddock/libs/libplots.py
```diff
@@ -523,29 +561,317 @@ def scatter_plot_handler(capri_filename, cl_rank, png, dpi):
...
+    ----------
+    dfcl : pandas DataFrame
+        dataframe of capri values
+
+    Returns
+    -------
+    dfcl : pandas DataFrame
+        DataFrame of capri table with new column names
+    """
+    dfcl = dfcl.sort_values(by=["cluster_id"])
```
mgiulini, line 777: related to the previous comment: it might happen that the run gives rise to hundreds/thousands of clusters..in that case I am not sure it makes sense to list all of them and for sure they should not be ordered by cluster-id (the best cluster (lower `model-cluster-ranking`) might have `cluster_id = 2000`)

src/haddock/libs/libplots.py
```diff
@@ -523,29 +561,317 @@ def scatter_plot_handler(capri_filename, cl_rank, png, dpi):
...
+        )
+    return dfcl[table_col]
+
+
+def clt_table_handler(clt_file, ss_file):
+    """
+    Create a figure include tables.
+
+    The idea is to create tidy tables that report statistics recored in
+    capri_clt.tsv file and best structures recored in capri_ss.tsv file. By
```
mgiulini, line 800:
```suggestion
    The idea is to create tidy tables that report statistics available in
    capri_clt.tsv and capri_ss.tsv files. By
```

src/haddock/libs/libplots.py
```diff
@@ -523,29 +561,317 @@ def scatter_plot_handler(capri_filename, cl_rank, png, dpi):
...
+    return fig
+
+
+def find_best_struct(ss_file, number_of_struct=4):
+    """
+    Find best structures.
+
+    It inspects model-cluster-ranking recorded in capri_ss.tsv file and finds
+    the best models (models with lower ranks). By default, it selects the 4 best
+    models.
```
mgiulini, line 709: I agree with @amjjbonvin here: the number of structures to choose depend on many parameters, and the data we show here can disagree with those reported in previous steps of the workflow (for example the `clustfcc` step).

I see three potential solutions here:
1. we report mean and standard deviation over the full set of structures belonging to a cluster
2. we choose 4 as the default number (in this case we should state it clearly, e.g. "data refer to the first 4 structures")
3. (not so immediate) we make the analysis read this number from the `capri_clt.tsv` file, where the value of `clt_threshold` is reported

src/haddock/libs/libplots.py
```diff
@@ -523,29 +561,317 @@ def scatter_plot_handler(capri_filename, cl_rank, png, dpi):
...
+        number of models with lower model-cluster-ranking
+
+    Returns
+    -------
+    best_struct_df : pandas DataFrame
+        DataFrame of best structures
+    """
+    dfss = read_capri_table(ss_file)
+    dfss = dfss.sort_values(by=["cluster-id", "model-cluster-ranking"])
+    # TODO need a check for "Unclustered"
```
mgiulini, line 725: I wonder how to implement this check..I would maybe report the top `number_of_struct` unclustered structures in the table as well

src/haddock/libs/libplots.py
```diff
@@ -523,29 +561,317 @@ def scatter_plot_handler(capri_filename, cl_rank, png, dpi):
...
+        ] * number_of_cluster
+    best_struct_df = best_struct_df.assign(Structure=col_names)
+    best_struct_df = best_struct_df.pivot_table(
+        index=["cluster-id"],
+        columns=["Structure"],
+        values="model",
+        aggfunc=lambda x: x,
+        )
+    best_struct_df.reset_index(inplace=True)
+    best_struct_df.rename(columns={"cluster-id": "Cluster ID"}, inplace=True)
```
mgiulini, line 742: could you add some comments here and there to help clarifying the procedure?

src/haddock/libs/libplots.py
SarahAlidoost: good point. I am wondering how this issue was handled in haddock 2.4. Here, the table creates vertical scrolling. I committed the `report.html` generated for `-m 5`, see [my commit](https://github.com/haddocking/haddock3/pull/611/commits/996d16dae256bab94537dce146c037af12247efe). If the file size of the report is a concern, we need to think about exporting it to another format perhaps a csv file!

src/haddock/libs/libplots.py
SarahAlidoost: sure, I added comments. Let me know if something is still unclear.

src/haddock/libs/libplots.py
SarahAlidoost: see my comment above.

amjjbonvin:
On the HADDOCK server we report by default the top 10 clusters. For the visualisation all should be plotted, but may-be only the top10 color-coded and labelled

And the order is based on the scoring function and not the cluster number as Marco explained. I.e. the `score` column of `capri_clt.tsv`

sverhoeven:
The number of best structures has been set to 10. 

Replaced examples/analysis/report.html with [script](https://github.com/i-VRESSE/haddock3/blob/create_report/examples/analysis/report.py) to generate report.html from [tests/golden_data](https://github.com/haddocking/haddock3/tree/main/tests/golden_data)
The report.html is 831Kb. Should report.html be added to the repo or GH pages?

mgiulini (approved):
looks good now, time to merge it into the main branch!
as for the lint, yes, no worries, it will be fixed at some time

PR #617, opened 2023-02-20
Refactor installation

rvhonorato:
You are about to submit a new Pull Request. Before continuing make sure you read the [contributing guidelines](CONTRIBUTING.md) and you comply with the following criteria:

- [ ] You have stick to Python. Talk with us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [ ] Your code is well documented: proper docstrings and explanatory comments for those tricky parts
- [ ] You structured the code into small functions as much as possible. You can use classes if there's a (state) purpose
- [ ] code follows our coding style
- [ ] You wrote tests for the new code
- [ ] `tox` tests pass. *Run `tox` command inside the repository folder*
- [ ] `-test.cfg` examples execute without errors. *Inside `examples/` run `python run_tests.py -b`*
- [ ] PR does not add any *install dependencies* unless permission granted by the HADDOCK team
- [ ] PR does not break licensing
- [ ] Your PR is about writing documentation for already existing code :fire:
- [ ] Your PR is about writing tests for already existing code :godmode:

---

*detele this line and explain your PR here*

PR #619, opened 2023-02-20
added postprocess option to workflow

### LINKED ISSUE #614: add a postprocessing key to execute haddock3-analyse at the end of the workflow
For the workflow builder it would be nice to have `haddock3-analyse` run always at the end of the workflow. With a new option in the .cfg workflow (`postprocessing = true` ? ), haddock3 should run the postprocessing on a certain subset of folders (I would say only on the `caprieval` folders). In the workflow builder this option will be always active, while standard haddock command line users will have it set to false by default.

Discussed with @sverhoeven

mgiulini:

- [ ] Your PR is about CNS

---

Closes #614 by adding a new workflow parameter, `postprocess` (set by default to false). When this parameter is set to true, `haddock3-analyse` is run at the end of the workflow (over all the available `caprieval` folders).

In this way the workflow can create plots from the precalculated capri tables.


rvhonorato (approved):

amjjbonvin (approved):
Nice addition!

mgiulini:
Actually there's a problem with the tests..since the user does not have to have the plotting libraries (plotly, matplotlib) installed, the call to haddock3-analyse will fail..I guess this could be a good reason for unifying the requirements (incorporating the plotting libraries into the standard haddock3 environment)

rvhonorato:
> I guess this could be a good reason for unifying the requirements

I had done this in #617 as part of the refactoring but deleted that PR as it would not compatible with our current execution and not worth the trouble to test/debug.

We can tackle this specificaly migrating to `pyproject.toml` instead of `setup.py`

mgiulini:
> > I guess this could be a good reason for unifying the requirements
> 
> I had done this in #617 as part of the refactoring but deleted that PR as it would not compatible with our current execution and not worth the trouble to test/debug.
> 
> We can tackle this specificaly migrating to `pyproject.toml` instead of `setup.py`

I see...but in this case we would only need to drop the difference between `requirements.yml` and `requirements_full.yml`, incorporating the plotting libraries in `requirements.yml` and `requirements.txt` (so that they become part of the standard installation)

amjjbonvin:
I guess for now we can indeed merge requirements.yml and requirements_full.yml

mgiulini:
> I guess for now we can indeed merge requirements.yml and requirements_full.yml

perfect, will do in a separate PR!

PR #645, opened 2023-04-25
add function and variable annotations

alchemistcai:
Add most function and variable annotations for haddock.

Add `src/haddock/core/typing.py` to contain relative definations.

Adjust some functions to accept both `Path` and `str` when docstrings  say they accept `Path` and `str`while the code only accept `str`, and other situations like this.

Alter some lists to generators when they are used only once.

Known issues:I can't infer types of `haddock.libs.libontology.PDBFile.topology`, `ModuleIO.input`,`ModuleIO.output` and relative functions.They seem to have many types.

I have to say that adding annotations for haddock is not an esay job,maybe even not a good idea.Feel free to join this discussion to show your opinion about type annotations #637 :eyes: .


rvhonorato:
Thanks a lot @alchemistcai for putting the effort on this! 

I not only think that adding the types is a good idea but I also think we should add `mypy` to the CI actions, but this is not the most popular opinion in the haddock team.

I'll go over your changes in the coming days, for now I've merged the main into this PR since there were some conflicts after #649

rvhonorato:
Sorry for the delay in handling this one.

It has become a huge atomic PR, so I have rebased it to [haddocking:typing](https://github.com/haddocking/haddock3/tree/typing) and will continue in a separate PR to check for issues.

Thanks again @alchemistcai!

PR #648, opened 2023-05-01
make `Path()` accept `Persistent` and it's subclass.

alchemistcai:
Add a `__fspath__` protocal in `Persistent` defination to make `Path()` accept `Persisent` and it' subclass.Type-checking statements related with `Persisent` are simplified.

Functions accepting `Path` or `Persistent` now don't have to use `isinstance()` to check variable types.Use `Path(path)` is enough.

In order to get `Path` or `Persistent`'s string representation, I recommend to use `os.fspath(path)` instead of `str(Path(path))`.

If some variable is `Persistent` only,use `Persistent.rel_path` is better because `Path(pdbfile)` converts `Persistent` to str first then converts to Path,while `Persistent.rel_path` is Path already.

### DIFF
--- a/src/haddock/libs/libalign.py
+++ b/src/haddock/libs/libalign.py
@@ -232,8 +232,7 @@ def load_coords(pdb_f, atoms, filter_resdic=None, numbering_dic=None):
     coord_dic = {}
     chain_dic = {}
     idx = 0
-    if isinstance(pdb_f, PDBFile):
-        pdb_f = pdb_f.rel_path
+    pdb_f = Path(pdb_f)
     with open(pdb_f, "r") as fh:
         for line in fh.readlines():
             if line.startswith("ATOM"):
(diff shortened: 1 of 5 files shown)


alchemistcai:
According to [PEP 519](https://peps.python.org/pep-0519/).

joaomcteixeira:
Quick comment. While the implementation is correct, think about how obvious it is that `Path()` on `Persistent` returns a relative path. I agree with the explicit use of `os.fspath`, but the `.rel_path` attribute is more explicit. Instead of `isinstance`, `Persistent(path).rel_path` is an option. Cheers,

rvhonorato:
I see this one has gone stale and diverged quite a bit @alchemistcai do you still have capacity to solve the conflicts?

alchemistcai:
I think what @joaomcteixeira said makes sense.

I'd like to close this PR as it is not a nesessary improvement.

PR #658, opened 2023-06-05
Issuefixes

### LINKED ISSUE #487: check values in cfg files against .yaml files

### LINKED ISSUE #575: Booleans with capitals
**Describe the bug**

Booleans write with a capital letter cause an error when running haddock3.
It seems to come from the toml support that requires only small letters.
This is however not consistent with what we have in the default yaml files.

**To Reproduce**
Steps to reproduce the behavior, if relevant, provide the input that triggered the error:

Define a boolean starting with a capital letter, e.g.

randomremoval = False

**Expected behavior**

It is a rather trivial thing - I would think that raising an error for this is not necessary and the booleans can be transformed internally to lower case.

### LINKED ISSUE #578: Adding the config file to the data dir of a run

It would be nice and useful to copy the original config file of a run to the data dir of the run.
Currently we don't store it in the run which is not optimal for reproducibility.

Simply copy the original cfg file to the run/data directory

VGPReys:

- [X] You have sticked to Python. Please talk to us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [X] Your code is well documented: proper docstrings and explanatory comments for those tricky parts
- [X] You structured the code into small functions as much as possible. You can use classes if there is a (state) purpose

---
This PR aims at closing three issues:
- With the addition of a regular expression that checks the presence of upper case True and False in the .toml input file and lowering them on-the-fly, this closes #575 
- Also by checking values types and ranges/choices, by comparing values present in the defaults.yaml and parameter =  value present in the user provided .cfg file, this closes #487 
- Finally, a bunch of copies of the workflow haddock3 input files are now written in the `data` directory of the run. This includes the `original inputed file`, the `regular expression cleaned file` and the `enhanced full parameter file`, which includes all default parameters in addition to the ones provided by the user. This closes #578 

Accordingly, additionnal tests were added, while some other test were modified to fit the new implementation.

Finally, `run_tests.py` ran without errors and the comparisons by `compare-runs.py` returned only `No errors found - OKAY!`.

### DIFF
--- a/src/haddock/gear/prepare_run.py
+++ b/src/haddock/gear/prepare_run.py
@@ -78,6 +78,7 @@ ALL_POSSIBLE_GENERAL_PARAMETERS = set.union(
     config_optional_general_parameters,
     )
 
+# Dict mapping string types (in default.yaml) into python3 objects types
 TYPES_MAPPER = {
     'boolean': (bool,),
     'bool': (bool,),
@@ -116,7 +117,24 @@ def with_config_error(func):
 
 @lru_cache
 def _read_defaults(module_name, default_only=True):
-    """Read the defaults.yaml given a module name."""
+    """Read the defaults.yaml given a module name.
+    
+    Parameters
+    ----------
+    module_name : str
+        Name of the HADDOCK3 module
+    default_only : bool
+        If True, only the default value of each parameters found in the
+        default.yaml will be return, else all information are retrieved.
+
+    Return
+    ------
+    mod_default_config : dict
+        A dict holding default value for each parameters of the module if
+        default_only == True,
+        Else a dict of dict, contraining all information present in the
+        default.yaml file of the module.
+    """
     module_name_ = get_module_name(module_name)
     pdef = Path(
         haddock3_source_path,
@@ -125,8 +143,8 @@ def _read_defaults(module_name, default_only=True):
         module_name_,
         'defaults.yaml',
         ).resolve()
-
-    return read_from_yaml_config(pdef, default_only=default_only)
+    mod_default_config = read_from_yaml_config(pdef, default_only=default_only)
+    return mod_default_config
 
 
 def setup_run(
@@ -307,7 +325,7 @@ def setup_run(
     enhanced_haddock_params = deepcopy(general_params)
     enhanced_haddock_params.update(modules_params)
     config_files['enhanced_haddock_params'] = enhanced_haddock_params
-    save_configuration_files(config_files, data_dir)
+    config_saves = save_configuration_files(config_files, data_dir)
 
     if scratch_rest0:
         copy_molecules_to_data_dir(
@@ -340,7 +358,7 @@ def setup_run(
     return modules_params, general_params
 
 
-def save_configuration_files(configs: dict, datadir: [str, Path]):
+def save_configuration_files(configs: dict, datadir: [str, Path]) -> dict:
     """Write a copy of configuration files (GitHub issue #578).
 
     Parameters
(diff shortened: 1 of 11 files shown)


src/haddock/gear/prepare_run.py
```diff
@@ -75,6 +78,22 @@
...
+    'float': (float, int,),
+    'double': (float,),
+    'string': (str,),
+    'str': (str,),
+    'list': (list,),
+    'array': (list,),
+    'path': (str, Path, PosixPath, EmptyPath,),
+    'file': (str, Path, PosixPath, EmptyPath,),
+    }
+
```
mgiulini, line 96: could you add a comment explaining this?

src/haddock/gear/prepare_run.py
```diff
@@ -96,7 +115,7 @@ def wrapper(*args, **kwargs):
 
 
 @lru_cache
-def _read_defaults(module_name):
+def _read_defaults(module_name, default_only=True):
```
mgiulini, line 118: what is default_only? could you write its meaning in the docstring?

src/haddock/gear/prepare_run.py
```diff
@@ -315,6 +340,74 @@ def setup_run(
...
+
+    # Add README to help user
+    with open(Path(confpaths, 'README.txt'), 'w') as f:
+        f.write(f"{'#'*80}\n# Information about configuration "
+                f"files present in the same directory #\n{'#'*80}\n")
+        for confname in list_save_conf:
+            if confname not in added_files.keys():
+                continue
+            f.write(f'"{added_files[confname]}": {infofile[confname]}\n')
+
```
mgiulini, line 409: I would return something here, like the paths to the saved file

src/haddock/gear/prepare_run.py
```diff
@@ -404,6 +497,144 @@ def pretty_print(match):
                 )
             raise ConfigurationError(_msg)
 
+        # Now that existance of the parameter was checked,
```
mgiulini, line 500:
```suggestion
        # Now that existence of the parameter was checked,
```

src/haddock/gear/yaml2cfg.py
```diff
@@ -119,13 +119,32 @@ def _yaml2cfg_text(ycfg, module, explevel):
...
+        Set the return value of this function; if True (default value), only
+        returns default values, else return the fully loaded configuration file
+
+    Return
+    ------
+    ycfg : dict
+        The full default configuration file as a dict
+    OR
+    cfg : dict
+        A dictionnary containing only the default parameters values
```
mgiulini, line 139:
```suggestion
        A dictionary containing only the default parameters values
```

mgiulini, line 139: is there any particular reason to have two different dict objects in output?

mgiulini (commented):
added a few comments, I will test it asap!

src/haddock/gear/yaml2cfg.py
VGPReys: There is the original one `cfg` that only contains the ['default'] value and used in other parts of the code.
While the `ycfg` one contrains verything present in the default.yaml file loaded as  a dict, used for the analysis of the agreement (type, range, choices, sizes) between user provided configuration and authorized one.

src/haddock/gear/prepare_run.py
VGPReys: Good idea !

rvhonorato:
Sorry I have absolutely no idea about this part of the code, will trust your judgment 🤞🏽

mgiulini (approved):
Runs nicely on a couple of examples, great addition @VGPReys!

PR #675, opened 2023-08-08
Chain matching and multibody capri

### LINKED ISSUE #164: Add multibody support to `caprieval`
We can align on chain A and calculate the metrics or simply produce all combinations of interfaces; `A_BC, B_AC, C_AB`, ligands; `A_BC, B_AC, C_AB, AB_C, AC_B, BC_A` etc. 

Which should I implement @amjjbonvin ?

### LINKED ISSUE #179: Implement chain-matching/check in `caprieval`

mgiulini:

- [ ] Your PR is about CNS

---

Closes #164 and Closes #179 

Sorry for the PR solving two issues but I wanted to dig into the caprieval routines only once :)

this PR adds the multibody support to caprieval by simply converting the ligand_chain string into a list, such that more chains are considered during L-RMSD and IL-RMSD calculations. Fnat calculation is not touched by this modification

Chain matching is implemented as follows:

1. if the reference and model chains are not equivalent, we calculate all possible combinations of chains
2. we perform sequence alignment for all these pairs of chains
3. we select the pairs that give the highest identity iteratively, removing the paired chains from the available combinations

PS: if the chains in reference and model structures are the same (for ex. [A,B,C] and [A,B,C]), chain matching is not executed to avoid potential problems while analysing homomers (and to save computing time).

PS2: if the chains differ from the standard [A, B] we still raise a warning


mgiulini:
> I will trust you on this. Did you run the examples to make sure it does not affect the standard CAPRI stats for systems with only two molecules?

Yes, almost all the data are the same with the obvious exception of the homotrimer example, where I checked the new data against an external software. There are small differences in the protein-protein and protein-ligand examples, but I think the new data are correct..in the protein-protein FNAT calculation now HIS90 and NEP90 are not matched anymore in the alignment, while previously they were considered as the same...I know it would be optimal to have them matched (so as to consider their contact in the FNAT calculation), but there's no way we can easily handle this in capri, as NEP90 is labelled as "X" in the alignment.
We could modify the BLOSUM alignment parameters to penalize the gap opening, but that does not seem to be within HADDOCK purpose..alternatively, we can add an option to have a sequential, alignment-free matching..what do you think?

amjjbonvin:
> in the protein-protein FNAT calculation now HIS90 and NEP90 are not matched anymore in the alignment, while previously they were considered as the same...I know it would be optimal to have them matched (so as to consider their contact in the FNAT calculation), but there's no way we can easily handle this in capri, as NEP90 is labelled as "X" in the alignment. 

Any way to replace on the fly NEP by HIS in such a case?

mgiulini:
> > in the protein-protein FNAT calculation now HIS90 and NEP90 are not matched anymore in the alignment, while previously they were considered as the same...I know it would be optimal to have them matched (so as to consider their contact in the FNAT calculation), but there's no way we can easily handle this in capri, as NEP90 is labelled as "X" in the alignment.
> 
> Any way to replace on the fly NEP by HIS in such a case?

Wonderful idea! we can couple non-standard amino acids to standard ones here https://github.com/haddocking/haddock3/blob/2fa67402647714196a365b0580383fb724ca1969/src/haddock/libs/libalign.py#L365-L391
this should solve everything!

mgiulini:
if you agree I'll modify the list to account for these residues https://wenmr.science.uu.nl/haddock2.4/library

amjjbonvin:
Great - and we can account in that way for all modified amino acids currently supported

amjjbonvin (approved):
Great!

amjjbonvin:
PS: Could this give issues if you are using all heavy atoms for RMSD calculations?

mgiulini:
> PS: Could this give issues if you are using all heavy atoms for RMSD calculations?

No, as

- this modification concerns only the sequence alignment
- in RMSD calculations we always consider atoms that are present in both structures (the intersection)

But since we are mapping modified amino-acids to standard ones in the sequence alignment, probably it makes sense to consider them as amino acids in the RMSD calculations (here: https://github.com/haddocking/haddock3/blob/dd04ad31be01ec658eef53faa8e484c2949e02d7/src/haddock/libs/libalign.py#L39-L60), so that we don't treat them as "ligands" (considering all heavy atoms) but as proper amino acids, considering only the backbone in RMSD calculations..what do you think?

amjjbonvin:
Indeed - these are clearly not ligands

> But since we are mapping modified amino-acids to standard ones in the sequence alignment, probably it makes sense to consider them as amino acids in the RMSD calculations (here: https://github.com/haddocking/haddock3/blob/dd04ad31be01ec658eef53faa8e484c2949e02d7/src/haddock/libs/libalign.py#L39-L60 <https://github.com/haddocking/haddock3/blob/dd04ad31be01ec658eef53faa8e484c2949e02d7/src/haddock/libs/libalign.py#L39-L60>), so that we don't treat them as "ligands" (considering all heavy atoms) but as proper amino acids, considering only the backbone in RMSD calculations..what do you think?

src/haddock/libs/libalign.py
```diff
@@ -387,7 +396,29 @@ def pdb2fastadic(pdb_f):
             ("DA", "A"),
             ("DG", "G"),
             ("DC", "C"),
-            ("DT", "T"),
+            ("DT", "T"),  # 9/8/2023: adding non-standard amino-acids (https://wenmr.science.uu.nl/haddock2.4/library) # noqa: E501
```
rvhonorato, line 399: This `wenmr` link is already derived and for a different version of the code, please use the source: `src/haddock/cns/toppar/protein-allhdg5-4.top` related to this project

src/haddock/libs/libalign.py
```diff
@@ -596,113 +627,272 @@ def align_strct(reference, model, output_path, lovoalign_exec=None):
...
+    aligner.substitution_matrix = substitution_matrices.load("BLOSUM62")
+    alns = aligner.align(seq_ref, seq_model)
+    top_aln = alns[0]
+
+    aln_denom = min(len(seq_ref), len(seq_model))
+    identity = (str(top_aln).count("|") / float(aln_denom)) * 100
+    
+    aln_ref_seg, aln_mod_seg = top_aln.aligned
+    # this should always be true
+    assert len(aln_ref_seg) == len(aln_mod_seg)
```
rvhonorato, line 685: This will raise a  non-specific `AssertionError`, it's best to replace this with a custom exception and do the proper error handling

src/haddock/libs/libalign.py
```diff
@@ -596,113 +627,272 @@ def align_strct(reference, model, output_path, lovoalign_exec=None):
...
             log.warning(
-                f"No alignment for chain {ref_chain} is it protein/dna-rna? "
+                f"No alignment for chain {ref_ch} is it protein/dna-rna? "
                 "Matching sequentially"
                 )
             if all(
-                    "X" in s for s in seq_ref) and all(
-                    "X" in s for s in seq_model):
+                    "X" in s for s in self.seqs_ref[ref_ch]) and all(
+                    "X" in s for s in self.seqs_model[mod_ch]):
```
rvhonorato, line 738: Any chance these "X" endup being lowercase? If so best to use `.upper()` in the conditional

src/haddock/libs/libalign.py
```diff
@@ -596,113 +627,272 @@ def align_strct(reference, model, output_path, lovoalign_exec=None):
...
             if identity <= 40.0:
                 # Identity is very low
                 log.warning(
-                    f"Sequence identity of chain {ref_chain} is "
+                    f"Sequence identity of chain {ref_ch} is "
                     f"{identity:.2f}%, please check the results carefully")
-                log.warning(
-                    "Please use alignment_method = \"structure\" instead")
+                # log.warning(
+                #    "Please use alignment_method = \"structure\" instead")
```
rvhonorato, line 756: dead code?

src/haddock/modules/analysis/caprieval/capri.py
```diff
@@ -203,10 +213,11 @@ def calc_lrmsd(self):
         if len(obs_chains) < 2:
             log.warning("Not enough chains for calculating lrmsd")
         else:
-            r_chain, l_chain = self.check_chains(obs_chains)
+            r_chain, l_chains = self.check_chains(obs_chains)
             r_start, r_end = chain_ranges[r_chain]
-            l_start, l_end = chain_ranges[l_chain]
-
+            l_starts = [chain_ranges[l_chain][0] for l_chain in l_chains]
+            l_ends = [chain_ranges[l_chain][1] for l_chain in l_chains]
```
rvhonorato, line 219: `l_chain` in `l_chains` is a bit confusing, how about `for _l in l_chains`

rvhonorato (changes requested):
would be also good to remove the commented code sections

src/haddock/libs/libalign.py
mgiulini: I was not aware of this, thanks! in the new topology there's residue CHX and I have no idea what it maps to (maybe cysteine?)..do you know it?

src/haddock/libs/libalign.py
mgiulini: absolutely no chance :)

amjjbonvin:
I will not map this one to any standard amino acid.


> I was not aware of this, thanks! in the new topology there's residue CHX and I have no idea what it maps to (maybe cysteine?)..do you know it?

src/haddock/modules/analysis/caprieval/capri.py
mgiulini: good point!

src/haddock/libs/libalign.py
mgiulini: corrected!

rvhonorato (approved):

PR #685, opened 2023-08-28
ilrmsdmatrix module

### LINKED ISSUE #684: develop ilRMSD-based clustering
HADDOCK2.X had a way to implement RMSD clustering that used the interface ligand RMSD as a measure of similarity between models. The current `rmsdmatrix` module in HADDOCK uses a different logic to calculate the RMSD.

It should be pretty easy to add a `ilrmsdmatrix` module to calculate the ilrmsd matrix. The already available `clustrmsd` module can be then easily applied to cluster this matrix, with all the advantages coming from using an agglomerative clustering approach over the greedy clustering used in HADDOCK2.X

mgiulini:

- [ ] Your PR is about CNS

---

Closes #684


amjjbonvin (commented):
One comment - further I leave it to the experts

examples/docking-protein-glycan/docking-protein-glycan-ilrmsd-test.cfg
```diff
@@ -0,0 +1,62 @@
...
+reference_fname = "data/target.pdb" 
+
+# rigidbody models containing glycans can be very similar to each other
+# especially when the glycans are short and linear. ilRMSD clustering after
+# rigidbody is useful to remove redundant models
+[ilrmsdmatrix]
+
+[clustrmsd]
+criterion = 'maxclust'
+tolerance = 4 # the number of clusters to be formed
```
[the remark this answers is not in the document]
mgiulini: Actually it is, but I agree it's super confusing.
I'll change things according to this logic:
```
tolerance -> clust_cutoff (if numerical value)
tolerance -> n_clusters (if we want a certain number of clusters)
threshold -> min_population
```
I'll do this in another PR!

src/haddock/modules/analysis/ilrmsdmatrix/ilrmsd.py
```diff
@@ -0,0 +1,324 @@
...
+        self.path = path
+        self.atoms = {}
+        self.receptor_chain = params["params"]["receptor_chain"]
+        self.ligand_chains = params["params"]["ligand_chains"]
+        self.unique_rec_res = []
+        self.unique_lig_res = []
+        for m in model_list:
+            self.atoms.update(get_atoms(m))
+        
+    def run(self):
```
rvhonorato, line 66: question; you need to have this method for the schedulers to pick it up, right?

src/haddock/modules/analysis/ilrmsdmatrix/ilrmsd.py
```diff
@@ -0,0 +1,324 @@
...
+        """Write down the RMSD matrix."""
+        output_fname = Path(self.path, self.output_name)
+        # check if there are very low values in the RMSD vector
+        check_low_values = np.isclose(
+            self.data[:, 2],
+            np.zeros(self.npairs),
+            atol=0.1
+            ).any()
+        if check_low_values:
+            log.warning(f"core {self.core}: low values of RMSD detected.")
```
rvhonorato, line 306: Could you add some comment here why this warning is raised - why is it important that people know that there are low values of rmsd.

src/haddock/modules/analysis/ilrmsdmatrix/ilrmsd.py
```diff
@@ -0,0 +1,324 @@
...
+                data_str += os.linesep
+                out_fh.write(data_str)
+
+
+def get_pair(nmodels, idx):
+    """Get the pair of structures given the 1D matrix index."""
+    if (nmodels < 0 or idx < 0):
+        err = "get_pair cannot accept negative numbers"
+        err += f"Input is {nmodels} , {idx}"
+        raise ValueError(err)
```
rvhonorato, line 319: Strange to have this validation here, what would trigger a negative input?

rvhonorato (changes requested):

integration_tests/test_ilrmsdmatrix.py
```diff
@@ -0,0 +1,58 @@
...
+    exp_contacts_file = Path(ilrmsdmatrix_module.path, "receptor_contacts.con")
+    assert exp_ilrmsd_matrix.exists(), "ilrmsd.matrix does not exist"
+    assert exp_contacts_file.exists(), "receptor_contacts.con does not exist"
+    # open files and check content
+    with open(exp_ilrmsd_matrix) as f:
+        assert f.readline() == "1 2 20.020\n"
+    with open(exp_contacts_file) as f:
+        lines = f.readlines()
+        assert lines[0] == "R 12 13 15 17 21 25 30 31 32 33 34 35 37 38 39 40 41 54 55 56 57 58 59 60 61 62 63 64 65 66 67 68 69 70 71 73 95 102 103 105\n"
+        assert lines[1] == "S 809 810 814 822 824 825 826 828 829 832 833 836 869 872 875 876 879 880 881 882 884 908 909 910 911 912 913 929 930 931 932 934 935 936 938 939 940 942 943 944 945 963 1002 1003 1006 1007 1010 1019\n"
```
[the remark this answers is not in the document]
mgiulini: definitely, nice catch!

src/haddock/modules/analysis/ilrmsdmatrix/defaults.yaml
```diff
@@ -0,0 +1,36 @@
...
+  type: integer
+  min: 1
+  max: 20000
+  title: Maximum number of models to calculate RMSD matrix
+  short: If the number of models exceeds max_models the execution is blocked
+  long: If the number of models exceeds the few thousands, the calculation of
+    the RMSD matrix is computationally demanding, especially in terms of CPU
+    and disk space.
+  group: ''
+  explevel: easy
```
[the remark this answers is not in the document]
mgiulini: mmm we don't want users to run matrix calculation that take forever..but probably it makes sense to set the value to 10K without allowing to modify it

src/haddock/modules/analysis/ilrmsdmatrix/ilrmsd.py
mgiulini: exactly!

src/haddock/modules/analysis/ilrmsdmatrix/ilrmsd.py
mgiulini: certainly!

src/haddock/modules/analysis/ilrmsdmatrix/ilrmsd.py
mgiulini: it should be impossible, but since a negative number could mess up things badly I thought to put an extra validation step there

integration_tests/test_ilrmsdmatrix.py
```diff
@@ -0,0 +1,58 @@
...
+from pathlib import Path
+import shutil
+import pytest
+import pytest_mock
+
+from haddock.libs.libontology import PDBFile
+
+from haddock.modules.analysis.ilrmsdmatrix import DEFAULT_CONFIG as DEFAULT_ILRMSD_CONFIG
+from haddock.modules.analysis.ilrmsdmatrix import HaddockModule as IlrmsdmatrixModule
+DATA_DIR = Path(Path(__file__).parent.parent / "tests" / "golden_data")
```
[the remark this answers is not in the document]
rvhonorato: It is important, yes

src/haddock/libs/libparallel.py
```diff
@@ -41,6 +41,8 @@ def get_index_list(nmodels, ncores):
     index_list : list
         List of model indexes to be used for the parallel scanning.
     """
+    if nmodels < 1 or ncores < 1:
+        raise ValueError("nmodels and ncores must be greater than 0")
```
VGPReys, line 45: what about something more explicit ?

```python3
    if nmodels < 1:
        raise ValueError("nmodels must be greater than 0")
    elif ncores < 1:
        raise ValueError("ncores must be greater than 0")
```

VGPReys (dismissed):

src/haddock/libs/libparallel.py
mgiulini: good suggestion, will implement it!

src/haddock/modules/analysis/ilrmsdmatrix/__init__.py
```diff
@@ -39,23 +39,7 @@
...
-    rem = nmodels % ncores
-    # now the list of indexes to be used for the SCAN calculation
-    index_list = [0]
-    for core in range(ncores):
-        if core < rem:
-            index_list.append(index_list[-1] + spc + 1)
-        else:
-            index_list.append(index_list[-1] + spc)
-    return index_list
+MAX_MODELS = 10000
```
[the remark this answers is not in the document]
mgiulini: it doesn't make much sense to let the users modify this parameter..10k is already a lot, if there're more input models you should never use RMSD-based clustering (at least in docking related contexts)

mgiulini:
@VGPReys @rvhonorato I should have implemented your suggestions in the code, thanks for the review!

amjjbonvin:
> it doesn't make much sense to let the users modify this parameter..10k is already a lot, if there're more input models you should never use RMSD-based clustering (at least in docking related contexts)
> 
Any idea of the timing for clustering 10K models with RMSD?

mgiulini:
> it doesn't make much sense to let the users modify this parameter..10k is already a lot, if there're more input models you should never use RMSD-based clustering (at least in docking related contexts)
> Any idea of the timing for clustering 10K models with RMSD?

clustering is never a problem, the matrix calculation is..with the native python implementation in HADDOCK3 the ilrmsdmatrix in the glycan example took ~10 minutes for 1k models on 10 CPU cores..since it scales quadratically probably we should limit the number of models to 4-5k instead of 10k

rvhonorato (approved):

VGPReys (approved):

amjjbonvin:
I would make that a parameter - up to the user to decide how much time they want to spend on it.

If you select less than the number of available models, will that be done automatically on the top ranked ones?
Or would you need a seletop step before that, e.g. to reduce from 20k to 5k before clustering

mgiulini:
> I would make that a parameter - up to the user to decide how much time they want to spend on it. If you select less than the number of available models, will that be done automatically on the top ranked ones? Or would you need a seletop step before that, e.g. to reduce from 20k to 5k before clustering

OK about making it a parameter, but the max value should not exceed 20k imo, so as to avoid memory problems and super long executions
currently if MAX_MODELS is less than the number of models the code raises an error

amjjbonvin:
Ok - the max param can be 10K (expert users can modify this)

And this means a seletop step would be needed first to cluster less than the total number of models

rvhonorato:
> I would make that a parameter - up to the user to decide how much time they want to spend on it.

Shouldn't we simply improve the code to make this faster?

mgiulini:
> > I would make that a parameter - up to the user to decide how much time they want to spend on it.
> 
> Shouldn't we simply improve the code to make this faster? Adding this as a parameter and all this contour conditions just to avoid programimg?

not sure code improvements can make much of a difference here: for sure the computation takes longer than it should, but the matrix calculation scales quadratically with the number of models..even if we made the code faster by a factor of 10, the execution will take a lot of time for 10k input models, because 50 million distances have to be computed.

as for the code improvements, it's a big choice, as it means adding extra dependencies and spending a lot of time on it. And there aren't so many cases in which we really need that efficiency. I am not saying I am against that, just that it must be discussed with the other developers

PR #688, opened 2023-08-29
added description to analysis modules

### LINKED ISSUE #687: parameters description in some analysis modules
some analysis module have no parameter description in their defaults.yaml file

mgiulini:

- [ ] Your PR is about CNS

---

Closes #687 by adding the parameter description when missing. Can you check that the description of `fraction_cutoff` 
https://github.com/haddocking/haddock3/blob/2cf706d15cba353f895049ca3489ae2fba421290/src/haddock/modules/analysis/clustfcc/defaults.yaml#L26C3-L26C3

and `strictness` https://github.com/haddocking/haddock3/blob/2cf706d15cba353f895049ca3489ae2fba421290/src/haddock/modules/analysis/clustfcc/defaults.yaml#L47 parameters are correct?


src/haddock/modules/analysis/seletop/defaults.yaml
```diff
@@ -3,8 +3,8 @@ select:
   type: integer
   min: -9999
```
VGPReys, line 4: Why -9999 is an allowed value ?
Maybe information in the `long` description to explain how to use negative values could be nice ? or should it be limitted to `min = 1` ?

src/haddock/modules/analysis/seletopclusts/defaults.yaml
```diff
@@ -14,8 +14,8 @@ top_models:
   min: -9999
```
VGPReys, line 14: Why -9999 is an allowed value ?
Maybe information in the `long` description to explain how to use negative values could be nice ? or should it be limitted to `min = 1` ?

VGPReys (dismissed):
Nice addition.

src/haddock/modules/analysis/seletop/defaults.yaml
mgiulini: I have no idea, I didn't modify it..probably makes sense to put `min=1`

src/haddock/modules/analysis/clustfcc/defaults.yaml
```diff
@@ -1,51 +1,51 @@
...
+  short: Relative path to contact_fcc executable.
+  long: As contact_fcc is an external tool, its path must be provided. Basically, this parameter should not be changed, as the tool is downloaded at the same time as haddock3. Nevertheless, one could try to have the contact_fcc tool written in another programming language or whatever.
+  group: executable
+  explevel: expert
 contact_distance_cutoff:
   default: 5.0
   type: float
-  min: -9999
-  max: 9999
+  min: 0.000
```
amjjbonvin, line 12: I would put the min to 2 and not 0

src/haddock/modules/analysis/clustfcc/defaults.yaml
```diff
@@ -1,51 +1,51 @@
...
+  max: 1.000
   precision: 3
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Fraction of common contacts to not be considered a singleton model.
```
amjjbonvin, line 26: singleton model is not a very common term and will be confusing to users. Why not a simpler description, e.g. minimum fraction of common contacts to be considered in a cluster

amjjbonvin: And for the other levels, I would write it as minimum fraction to be considered in a cluster (rather than not be considered - we want to cluster at the end :-) )

src/haddock/modules/analysis/clustfcc/defaults.yaml
```diff
@@ -1,51 +1,51 @@
...
-  min: -9999
+  min: 1
   max: 9999
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  title: Clustering population threshold.
+  short: Threshold employed to exclude clusters with less than this number of members. By default 4.
+  long: Minimum set of grouped complexes to define a cluster. If a potential cluster is composed of less than this number of members, it will be considered as `singloton` and discarded from final clusters. If set to 1, all complexes will be part of a cluster. By default 4.
```
amjjbonvin, line 38: singleton

src/haddock/modules/analysis/clustfcc/defaults.yaml
```diff
@@ -1,51 +1,51 @@
...
+  group: analysis
   explevel: easy
 strictness:
```
amjjbonvin, line 41: Isn't this parameter redundant with the fraction_cutoff ? I.e. can be removed (should make sure it is no longer used) Now we have inconsistent values (0.6 vs 0.75)

amjjbonvin (commented):
Please see/address my comments. Then should be merged as important default settings and explanations

src/haddock/modules/analysis/clustfcc/defaults.yaml
mgiulini: I think those two parameters address two different tasks..the fraction_cutoff tells you if a model will belong to any cluster (if it has fcc < 0.6 wrt all the other models then it's a "singleton"), while the strictness controls when you should put complexes in the same cluster..I am not 100% sure though

amjjbonvin: The haddock2.4 default for FCC clustering is 0.6.  Which is the default for the fraction_cutoff.

I don't see why we should have two different parameters for those two tasks. The same parameter can be used for both tasks, with > to make it cluster and < to make it a singleton

src/haddock/modules/analysis/seletopclusts/defaults.yaml
```diff
@@ -6,16 +6,15 @@ top_cluster:
...
-  title: No title yet
-  short: No short description yet
-  long: No long description yet
-  group: ''
+  default: 10
+  type: integer
+  min: 1
+  max: 99999
+  title: Number of best-ranked models to select per cluster
+  short: Number of best-ranked models to select per cluster. By default, all models are selected.
```
amjjbonvin, line 17: By default the top 10 models are selected

src/haddock/modules/analysis/seletopclusts/defaults.yaml
```diff
@@ -6,16 +6,15 @@ top_cluster:
...
+  long: Number of best-ranked models to select per cluster. By default, all models are selected. If one expects to find many clusters (e.g., in the case of vague restraints), this number can be reduced to speed up the protocol. Instead, if few clusters are expected (very specific and localized restraints) this number can be increased.
```
amjjbonvin, line 18: By default the top 10 models are selected

amjjbonvin (commented):
Minor changes since we now have the default max number of selected models set to 10

VGPReys (approved):

PR #690, opened 2023-08-30
alascan module

### LINKED ISSUE #290: `alascan` for analysis module.
It is a nice analysis module!

_Originally posted by @amjjbonvin in https://github.com/haddocking/haddock3/issues/272#issuecomment-1021993275_

mgiulini:

- [ ] Your PR is about CNS

---

Closes #290 by adding a new alascan module, that performs an Alanine (by default, but any residue is possible) scanning of the interface of the input models using the haddock3-score CLI (topology + short emscoring). This will create a scan_{model_name}.csv file (for example `scan_emref_1.csv`) with the following structure:
```
##########################################################
# `alascan` results for emref_1.pdb
#
# z_score is calculated with respect to the other residues
##########################################################
chain   res     ori_resname     end_resname     score   vdw     elec    desolv  bsa     delta_ori_score delta_score     delta_vdw       delta_elec      delta_desolv delta_bsa        z_score
A       32      GLN     ALA     -110.950        -37.055 -454.059        16.917  1297.070        -15.220 1.629   2.544   4.309   -1.776  -11.030 -0.251
A       42      PRO     ALA     -116.217        -41.453 -466.581        18.552  1312.090        -20.487 -3.637  -1.854  -8.213  -0.141  3.990   -1.073
A       38      VAL     ALA     -111.742        -39.273 -456.969        18.925  1292.850        -16.012 0.838   0.326   1.399   0.232   -15.250 -0.375
```
This file has one row per mutated residue, containing all the differences in the score and int its components. The Z score column is calculated by comparing the `delta_score` value to those of the other residues.

If cluster information is available, alanine scanning data are grouped in a cluster-based manner in the `scan_clt_{clt_id}.csv` file (one per cluster):
```
#######################################################################
# `alascan` cluster results for cluster 1
#
# z_score is calculated with respect to the mean values of all residues
#######################################################################
chain   resnum  resname full_resname    delta_score     delta_vdw       delta_elec      delta_desolv    delta_bsa       frac_pres       z_score
A       10      ARG     A-10-ARG        8.575   -1.296  56.270  -1.383  -35.645 0.800   1.140
A       16      THR     A-16-THR        -2.734  -0.775  -6.355  -0.688  -2.783  0.600   -1.136
A       18      THR     A-18-THR        3.393   0.749   11.262  0.392   -12.760 0.400   0.098
A       27      LYS     A-27-LYS        7.572   -1.695  65.472  -3.827  -66.829 0.800   0.938
A       28      GLN     A-28-GLN        6.883   2.498   29.169  -1.449  -15.900 0.200   0.800  
```
where the difference in the components are averaged over the different models. The Z score here is calculated against the average `delta_score` values of the other residues.

The module has five parameters (https://github.com/haddocking/haddock3/blob/alascan2.0/src/haddock/modules/analysis/alascan/defaults.yaml).

Future (possible) development: give the possibility to choose `mdscoring` to rescore the mutated models


rvhonorato:
Don't forget to add the integration test (https://github.com/haddocking/haddock3/pull/716/commits/de7d7a388ee62b4e87bfb4e0ca747b6a54bf2eef) 🤓

integration_tests/test_alascan.py
```diff
@@ -0,0 +1,46 @@
...
+            order=0, path=tmpdir, initial_params=DEFAULT_ALASCAN_CONFIG
+        )
+        alascan.params["int_cutoff"] = 3.5
+        alascan.params["output"] = True
+        yield alascan
+    
+class MockPreviousIO:
+    def retrieve_models(self, individualize: bool = False):
+        shutil.copy(Path("..", "tests", "golden_data", "protprot_complex_1.pdb"), Path(".", "protprot_complex_1.pdb"))
+        shutil.copy(Path("..", "tests", "golden_data", "protprot_complex_2.pdb"), Path(".", "protprot_complex_2.pdb"))
```
rvhonorato, line 25: Instead of `..`, add a `GOLDEN_DATA` variable to `integration_tests/__init__.py` and use it here

integration_tests/test_alascan.py
```diff
@@ -0,0 +1,46 @@
...
+    mocker.patch("haddock.modules.BaseHaddockModule.export_io_models", return_value = None)
+    alascan_module.run()
+
+    expected_csv1 = Path(alascan_module.path, "scan_protprot_complex_1.csv")
+    expected_csv2 = Path(alascan_module.path, "scan_protprot_complex_2.csv")
+    expected_clt_csv = Path(alascan_module.path, "scan_clt_-.csv")
+
+    assert expected_csv1.exists(), f"{expected_csv1} does not exist"
+    assert expected_csv2.exists(), f"{expected_csv2} does not exist"
+    assert expected_clt_csv.exists(), f"{expected_clt_csv} does not exist"
```
rvhonorato, line 45: Maybe its good to also check if the `csv` files are not empty and if they contain the actual deltas, ideally you'd assert that the observed are the same as the reference values or within a (small) range of variation

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -485,7 +485,7 @@ def run(self):
             try:
                 ori_score = float(ori_score)
             except ValueError:
-                ori_score = "-"
+                ori_score = np.nan
```
rvhonorato, line 488: 👍🏽

tests/test_module_alascan.py
```diff
@@ -318,8 +316,13 @@ def test_no_oriscore(mocker, scan_obj):
     scan_obj.model_list[0].ori_score = None
     mocker.patch("haddock.modules.analysis.alascan.scan.calc_score", return_value = (-106.7, -29, -316, -13, 1494))
     scan_obj.run()
-
-
+    df = pd.read_csv("scan_protprot_complex_1.csv", sep="\t", comment="#")
+    # column delta_ori_score should be nan
+    assert np.isnan(df["delta_ori_score"][0])
+    print(os.listdir("."))
```
rvhonorato, line 322: This print is an artefact of the development?

tests/test_module_alascan.py
```diff
@@ -407,7 +410,10 @@ def test_alascan_cluster_analysis(protprot_input_list, scan_file):
 
     # clean up
```
rvhonorato, line 411: Just in general, if your assert comes before the cleaning, if the test fail the files will not be deleted and this might interfere with the next round of tests (causing some artifact)

src/haddock/modules/analysis/alascan/defaults.yaml
```diff
@@ -0,0 +1,52 @@
...
+  maxitems: 100
+  title: List of residues to be mutated
+  short: List of residues to be mutated. By default all the interface residues are
+    mutated.
+  long: List of residues to be mutated. By default all the interface residues are
+    mutated. resdic_* is an expandable parameter. You can provide resdic_A,
+    resdic_B, resdic_C, etc, where the last capital letter is the chain
+    identifier.
+  group: analysis
+  explevel: easy
```
rvhonorato, line 14: Should this (and the other) be easy?

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -0,0 +1,591 @@
...
+        Desolvation energy.
+    bsa : float
+        Buried surface area.
+    """
+    cmd = f"haddock3-score {pdb_f} --full --run_dir {run_dir}"
+    p = subprocess.run(
+        shlex.split(cmd),
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+        )
```
rvhonorato, line 216: Not sure about this part, you are making a (external) subprocess call to run another CLI which is inside the same code. Here instead you should get the function instead, something like `import haddock.clis.cli_score.main as calc_score`.

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -0,0 +1,591 @@
...
+                    }
+            else:
+                clt_scan[cl_id][ident]['delta_score'] += delta_score
+                clt_scan[cl_id][ident]['delta_vdw'] += delta_vdw
+                clt_scan[cl_id][ident]['delta_elec'] += delta_elec
+                clt_scan[cl_id][ident]['delta_desolv'] += delta_desolv
+                clt_scan[cl_id][ident]['delta_bsa'] += delta_bsa
+                clt_scan[cl_id][ident]['frac_pr'] += 1
+    # now average the data
+    for cl_id in clt_scan:
```
rvhonorato, line 311: I'd recomend you split this loop into another function

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -0,0 +1,591 @@
...
+        with open(scan_clt_filename, 'w') as f:
+                f.write(f"#######################################################################{os.linesep}")  # noqa E501
+                f.write(f"# `alascan` cluster results for cluster {cl_id}{os.linesep}")  # noqa E501
+                f.write(f"#{os.linesep}")
+                f.write(f"# z_score is calculated with respect to the mean values of all residues{os.linesep}")  # noqa E501
+                f.write(f"#######################################################################{os.linesep}")  # noqa E501
+                f.write(fl_content)
+    return clt_scan
+
+def check_alascan_jobs(alascan_jobs):
```
rvhonorato, line 354: Shouldn't this handled by the ontology lib?

rvhonorato, line 354: Shouldn't this handled by the ontology lib?

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -0,0 +1,591 @@
...
+            make_alascan_plot(
+                df_scan_clt,
+                clt_id,
+                scan_residue
+                )
+        except Exception as e:
+            log.warning(
+                "Could not create interactive plot. The following error"
+                f" occurred {e}"
+                )
```
rvhonorato, line 427: This should `raise` an exception and you handle it upstream

rvhonorato (changes requested):
Nice! See the comments below. Great job on the tests 🤖

integration_tests/test_alascan.py
mgiulini: good point!

integration_tests/test_alascan.py
mgiulini: makes sense

tests/test_module_alascan.py
mgiulini: ooof indeed

tests/test_module_alascan.py
mgiulini: yep, that's a problem..should use the tempfile module

src/haddock/modules/analysis/alascan/defaults.yaml
mgiulini: it's an analysis module..I would give users the freedom to experiment. maybe only this resdic_ should be `expert`

src/haddock/modules/analysis/alascan/scan.py
mgiulini: I'll check if that's possible, if it is then it's definitely cleaner!

src/haddock/modules/analysis/alascan/scan.py
mgiulini: is it handled there? then this function can be removed!

src/haddock/modules/analysis/alascan/scan.py
mgiulini: not sure about this..maybe the code fails in creating a single plot (for some crazy reason) but all the others can be generated correctly. I think it makes sense to handle this scenario at this level

mgiulini:
@rvhonorato should be ready now, check it out!

integration_tests/test_alascan.py
```diff
@@ -0,0 +1,73 @@
...
+        atol=10)
+
+    # check clt csv
+    df_clt = pd.read_csv(expected_clt_csv, sep="\t", comment="#")
+    assert df_clt.shape == (18, 11), f"{expected_clt_csv} has wrong shape"
+    # average delta score of A-38-ASP should be around 8.18
+    assert np.isclose(
+        df_clt.loc[df_clt["full_resname"] == "A-38-ASP"]["delta_score"],
+        8.18,
+        atol=2)
```
rvhonorato, line 72: 👏🏽

src/haddock/libs/libalign.py
```diff
@@ -460,8 +468,10 @@ def load_coords(
                         #     " was not matched!"
                         #     )
                         continue
-                # identifier = f"{chain}.{resnum}.{atom_name}"
-                identifier = (chain, resnum, atom_name)
+                if add_resname is True:
+                    identifier = (chain, resnum, atom_name, resname)
+                else:
+                    identifier = (chain, resnum, atom_name)
```
rvhonorato, line 474: Why this `add_resname`? isn't it simpler to just change the identifer to always contain the resname?

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -0,0 +1,581 @@
...
+    ("PNS", "S"),
+    ("PTR", "Y"),
+    ("SEP", "S"),
+    ("TOP", "T"),
+    ("TYP", "Y"),
+    ("TYS", "Y"),
+    ])
+
+
+def get_index_list(nmodels, ncores):
```
rvhonorato, line 66: Looks like this function should be somewhere more general, some lib maybe?

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -0,0 +1,581 @@
...
+        if core < rem:
+            index_list.append(index_list[-1] + spc + 1)
+        else:
+            index_list.append(index_list[-1] + spc)
+    return index_list
+
+
+def mutate(pdb_f, target_chain, target_resnum, mut_resname):
+    """
+    Mutate a resnum to a resname.
```
rvhonorato, line 98: Could you improve this? A bit confusing

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -0,0 +1,581 @@
...
+    Returns
+    -------
+    out : list
+        List of strings with the score output.
+    """
+    f = io.StringIO()
+    with redirect_stdout(f):
+        cli_score.main(pdb_f, run_dir, full=True)
+    out = f.getvalue().split(os.linesep)
+    return out
```
rvhonorato, line 209: Nice

rvhonorato (approved):

rvhonorato:
With +1k lines new lines, 50 indirect changes and considering how esoteric the core of the code is, I'd say we eat the negative delta in code coverage.

src/haddock/modules/analysis/alascan/scan.py
mgiulini: makes sense. same function exists in #685 so I will have to call it also there

src/haddock/libs/libalign.py
mgiulini: that would not be optimal for the CAPRI module, where we may want to calculate CAPRI quantities irrespectively of the resname (if the alignment is satisfied of course)

amjjbonvin:
But we do have all those modified residues supported by haddock. So we should allow mutations to those, but this will only work when using a three letter code

mgiulini:
> But we do have all those modified residues supported by haddock. So we should allow mutations to those, but this will only work when using a three letter code

should we? yes, there's a list of accepted three-letter codes here https://github.com/haddocking/haddock3/blob/dbf142f3e4592f93f909148b41133ee41edaa6d2/src/haddock/modules/analysis/alascan/defaults.yaml#L34-L44

I just have to update it with the accepted residues

rvhonorato (approved):

PR #691, opened 2023-08-31
handled max sampling_factor

### LINKED ISSUE #689: increase max sampling factor in refinement modules
PR #658 added checks on parameters' values. Therefore, the max value of `sampling_factor` should be higher than one, otherwise it does not make sense to have this parameter. I propose to use `100` as maximum value (high enough to have variability when needed, for example in single structure refinement) and to modify the python code so as to handle possible mistakes when the user has, say, 1k models and sampling_factor = 100 (this should not be allowed).

Default and min values are correct.

mgiulini:

- [ ] Your PR is about CNS

---

Closes #689 by changing the max value of `sampling_factor` to 100 and adding a check at the python level such that if the product between the number of models and the `sampling_factor` is > 50 000 the code exits with an error.

I have absolutely no idea about why tests are failing..

### DIFF
--- a/src/haddock/modules/refinement/emref/__init__.py
+++ b/src/haddock/modules/refinement/emref/__init__.py
@@ -47,8 +47,9 @@ class HaddockModule(BaseCNSModule):
             sampling_factor = 1
         if sampling_factor > 100:
             self.log("[Warning] sampling_factor is larger than 100")
-
-        if len(models_to_refine) * sampling_factor > 50000:
+        
+        max_nmodels = self.params["max_nmodels"]
+        if len(models_to_refine) * sampling_factor > max_nmodels:
             self.finish_with_error(
                 "Too many models to refine. Please reduce the number of "
                 "models or decrease the sampling_factor."
--- a/src/haddock/modules/refinement/emref/__init__.py
+++ b/src/haddock/modules/refinement/emref/__init__.py
@@ -47,6 +47,12 @@ class HaddockModule(BaseCNSModule):
             sampling_factor = 1
         if sampling_factor > 100:
             self.log("[Warning] sampling_factor is larger than 100")
+
+        if len(models_to_refine) * sampling_factor > 50000:
+            self.finish_with_error(
+                "Too many models to refine. Please reduce the number of "
+                "models or decrease the sampling_factor."
+                )
         
         # checking the ambig_fname:
         try:
--- a/src/haddock/modules/refinement/emref/__init__.py
+++ b/src/haddock/modules/refinement/emref/__init__.py
@@ -49,10 +49,12 @@ class HaddockModule(BaseCNSModule):
             self.log("[Warning] sampling_factor is larger than 100")
         
         max_nmodels = self.params["max_nmodels"]
-        if len(models_to_refine) * sampling_factor > max_nmodels:
+        nmodels = len(models_to_refine) * sampling_factor
+        if nmodels > max_nmodels:
             self.finish_with_error(
-                "Too many models to refine. Please reduce the number of "
-                "models or decrease the sampling_factor."
+                f"Too many models ({nmodels}) to refine, max_nmodels ="
+                f" {max_nmodels}. Please reduce the number of models or"
+                " decrease the sampling_factor."
                 )
         
         # checking the ambig_fname:
--- a/src/haddock/modules/refinement/emref/defaults.yaml
+++ b/src/haddock/modules/refinement/emref/defaults.yaml
@@ -1346,7 +1346,7 @@ sampling_factor:
   default: 1
   type: integer
   min: 1
-  max: 1
+  max: 100
   title: Sampling factor for each starting model
   short: This paramater control how many times a model will be refined.
   long: This paramater control how many times a model will be refined. For EM refinement only it does not make sense to increase it unless random removal of restraints is turned on. If not then the energy minimisation will lead to the same final conformation. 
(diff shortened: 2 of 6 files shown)


rvhonorato:
Tests were failing because of tox, see https://github.com/haddocking/haddock3/pull/693, please merge from main before merging the pr

src/haddock/modules/refinement/emref/__init__.py
```diff
@@ -47,6 +47,12 @@ def _run(self):
             sampling_factor = 1
         if sampling_factor > 100:
             self.log("[Warning] sampling_factor is larger than 100")
+
+        if len(models_to_refine) * sampling_factor > 50000:
```
rvhonorato, line 51: out of scope here, but these numbers should be configurable constants instead of hardcoded

rvhonorato (approved):

src/haddock/modules/refinement/emref/__init__.py
VGPReys: should this parameter be in the `general` ones ?

VGPReys (changes requested):
I must agree with rodrigo, should we place a kind of `overall_max_models` parameter somewhere ? and setting its default value to 50000, which is a lot already...

mgiulini:
> I must agree with rodrigo, should we place a kind of `overall_max_models` parameter somewhere ? and setting its default value to 50000, which is a lot already...

I actually agree with this...there should be a `max_nmodels` parameter to be placed in the `defaults.yaml`..I propose the following default values for the three refinement modules:

- `max_nmodels`=10000 for flexref and mdref
- `max_nmodels`=50000 for emref

What do you say? Then probably the maximum `sampling_factor` can even be raised to 200 or 400

amjjbonvin:
Fine to have max values, but it is at the end up to the users.

Should be properly documented.

Also what if you hit the max, e.g. after a clustering step and increased sampling factor at flexref? Do you stop or only compute to the max number of models. And in that case how do you select the models?

VGPReys:
> > I must agree with rodrigo, should we place a kind of `overall_max_models` parameter somewhere ? and setting its default value to 50000, which is a lot already...
> 
> I actually agree with this...there should be a `max_nmodels` parameter to be placed in the `defaults.yaml`..I propose the following default values for the three refinement modules:
> 
>     * `max_nmodels`=10000 for flexref and mdref
> 
>     * `max_nmodels`=50000 for emref
> 
> 
> What do you say? Then probably the maximum `sampling_factor` can even be raised to 200 or 400

So any modules within the `modules/refinement/*/default.yaml` should have a `max_nmodels` parameter from now ?

I guess the values are OK.
Still can be risky to have a 10000 `mdref` accessible to any user.
Increasing the `explevel` could help not to have crazy `easy` users doing 10000 `mdref`.

Of note: It is now the second time that I think upper/lower boundaries of some parameters should be related to `explevel`.
But I do not know if with the current implementation this is feasable.
- It could be a kind of a new `guru parameters` with default setting to 0 that sum on top of `easy` one ?
- Should some parameters that require high computation be somehow restricted only using the web interface ?

mgiulini:
> Fine to have max values, but it is at the end up to the users. Should be properly documented. Also what if you hit the max, e.g. after a clustering step and increased sampling factor at flexref? Do you stop or only compute to the max number of models. And in that case how do you select the models?

In the proposed implementation haddock3 is stopped right at the beginning of the refinement module (message is "`Too many models to refine. Please reduce the number of models or decrease the sampling_factor`", will add the current and max number of allowed models to it).

I would not do a model selection here, since 99% of the times this limit is hit will be a user mistake.

mgiulini:
> > > I must agree with rodrigo, should we place a kind of `overall_max_models` parameter somewhere ? and setting its default value to 50000, which is a lot already...
> > 
> > 
> > I actually agree with this...there should be a `max_nmodels` parameter to be placed in the `defaults.yaml`..I propose the following default values for the three refinement modules:
> > ```
> > * `max_nmodels`=10000 for flexref and mdref
> > 
> > * `max_nmodels`=50000 for emref
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What do you say? Then probably the maximum `sampling_factor` can even be raised to 200 or 400
> 
> So any modules within the `modules/refinement/*/default.yaml` should have a `max_nmodels` parameter from now ?

I guess so:)
> 
> I guess the values are OK. Still can be risky to have a 10000 `mdref` accessible to any user. Increasing the `explevel` could help not to have crazy `easy` users doing 10000 `mdref`.

Yes of course, the `sampling_factor` is only accessible to expert users!

> Of note: It is now the second time that I think upper/lower boundaries of some parameters should be related to `explevel`. But I do not know if with the current implementation this is feasable.
> 
> * It could be a kind of a new `guru parameters` with default setting to 0 that sum on top of `easy` one ?
> * Should some parameters that require high computation be somehow restricted only using the web interface ?

I think that the overall HADDOCK idea is to give no flexibility to users below a certain expertise level, and full flexibility to the others..In my opinion implementing expertise-specific thresholds is a bit of an overcomplication..

rvhonorato:
I was suggesting a configuration value, not a module parameter; something in `src/haddock/modules/defaults.yaml`. 

what I mean is that imo validation is not in the scope of a module and should be handled in the workflow composition step instead - anyway out of scope of this PR

mgiulini:
> I was suggesting a configuration value, not a module parameter; something in `src/haddock/modules/defaults.yaml`.
> 
> what I mean is that imo validation is not in the scope of a module and should be handled in the workflow composition step instead - anyway out of scope of this PR

certainly out of the scope of this PR, but good point..the problem is that a priori you cannot know how many models will reach the refinement (thinking of clustering here), so there's no way to check this at the workflow composition step

VGPReys (dismissed):

VGPReys (approved):

PR #724, opened 2023-10-16
Implement type annotations

rvhonorato:
This PR continues #645;

**Many** type annotations were added, and where the code was too esoteric, I added `type: ignore`. I have also updated the actions to run `mypy` before running the tests. 

Since it support [gradual typing](https://mypy.readthedocs.io/en/stable/existing_code.html) and typing is not the norm, I've set it so that only the (new) typed functions should be checked - new untyped code should not be affected.


mgiulini (approved):

sverhoeven:
It would be nice to add py.typed file see https://peps.python.org/pep-0561/#packaging-type-information

rvhonorato:
That looks interesting indeed @sverhoeven - I'll do it in another PR

PR #732, opened 2023-10-24
Haddock3 restraints

### LINKED ISSUE #734: incorporate old restraints scripts in haddock3
old haddock scripts for restraints generation should be embedded in haddock3 so as to test them automatically before tutorials.

This will add the freesasa dependency to the code.

mgiulini:

- [ ] Your PR is about CNS

---

Closes #734 by adding a new CLI, haddock3-restraints, with four subcommands (`validate_tbl`, `restrain_bodies`, `passive_from_active`, `active_passive_to_ambig`) that can be used to generate, validate, and handle restraints.

this adds the freesasa dependency (necessary for `passive_from_active`).

For the sake of time I didn't rewrite these scripts completely, but at least now they are properly tested.


requirements.txt
```diff
@@ -10,3 +10,4 @@ tox==3.24.5
 pandas==1.5.1
 plotly==5.11.0
 kaleido==0.2.1
+freesasa==2.2.0.post3
```
rvhonorato, line 13: I don't see any information in their documentations about this `post` version, can you use `2.2.0` instead?

src/haddock/libs/librestraints.py
```diff
@@ -0,0 +1,627 @@
...
+    segid2 : string
+        Segid to use for the second model
+
+    """
+
+    all1 = active1 + passive1
+    all2 = active2 + passive2
+
+    for resi1 in active1:
+        print('assign (resi {:d} and segid {:s})'.format(resi1, segid1))
```
rvhonorato, line 192: Maybe a personal preference, but I'm not a big fan of these prints and would use `sys.stdout.write` instead but its a python nitpick, it's good as it is.

rvhonorato (dismissed):
I see quite a lot of minor issues but those are related to the legacy script rather than this port. I'd suggest you add a disclaimer/warning/big comment somewhere that this was ported from old code and there is room for improvment.

requirements.txt
mgiulini: probably yes!

mgiulini:
> I see quite a lot of minor issues but those are related to the legacy script rather than this port. I'd suggest you add a disclaimer/warning/big comment somewhere that this was ported from old code and there is room for improvment.

makes sense, I will add it in the main CLI. I won't print the disclaimer though, as the output of these scripts is typically redirected

requirements.txt
mgiulini: tried with 2.2.0 and 2.1.0, both failed for some hyper-weird reason..I would stick to this version for the time being

rvhonorato (approved):
Nicely ported!

PR #764, opened 2023-12-21
modif float comparisons in contact map tests

VGPReys:
Solve the digit comparison issues while changing hardware by using `np.isclose()`


rvhonorato (approved):
I'll trust you on the number of digits 🚀

PR #780, opened 2024-01-09
Fix prot segid

### LINKED ISSUE #741: topoaa does not check for similar chainIDs (and does not allow to define those in the config file)
**Describe the bug**

Buy default haddock3 takes the chainID/segID from the input PDB files. When two files are given with the same chainID, there is not check done and this will cause the next steps to fail (problem reported by an user). Basically the topology (psf) of the complex will only contain one chain (or some weird combination, but only with chain A). 

In principle, in the parameters (yaml file) we allow to define manually the segIDs which should allow to correct that:

```
(haddock3) bash-3.2$ haddock3-cfg -m topoaa

...
[topoaa.mol1]
prot_segid = 'A'  # $minchars 0 / $maxchars 4 / $title Segment ID / $group molecule

```

Specifying different `port_segid` values for the two chains in the config file is however not resolving the problem as those are ignored.

**To Reproduce**
Steps to reproduce the behavior, if relevant, provide the input that triggered the error:

Give twice the same molecule (but different filenames) for the docking in the config file and try some simple docking scenario.

Try to specify the prot_segid for each molecule in the config file, e.g.:

```
[topoaa]
[topoaa.mol1]
prot_segid = 'A'
[topoaa.mol2]
prot_segid = 'B'
```

**Expected behavior**

Ideally, haddock3 should check that the various molecules given as input don't have the same chainIDs.

Or, manually specifying the chainIDs should overwrite the chainIDs/segIDs from the input proteins.

Simple dataset to reproduce the problem:

[test-haddock3.tgz](https://github.com/haddocking/haddock3/files/13319732/test-haddock3.tgz)

mgiulini:

- [ ] Your PR is about CNS

---

Closes #741 by adding a check in libcns : for non-scoring modules, if the same chain ID is present in different PDBs then an error is raised and the execution is terminated.

The useless `prot_segid` parameter is also removed.

### DIFF
--- a/examples/run_tests.py
+++ b/examples/run_tests.py
@@ -153,14 +153,6 @@ def main(examples, break_on_errors=True):
                     stderr=sys.stderr,
                     )
 
-                subprocess.run(
-                    "haddock3 docking-protein-protein-test-start-from-cp.cfg --extend-run run2",  # noqa: E501
-                    shell=True,
-                    check=break_on_errors,
-                    stdout=sys.stdout,
-                    stderr=sys.stderr,
-                    )
-
                 # test exit with extend-run
                 rmtree("run2", ignore_errors=True)
                 subprocess.run(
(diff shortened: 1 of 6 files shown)


amjjbonvin (approved):
Are we checking somewhere that the input PDB files to have a proper chainID in them? I.e. not empty chainID

PR #825, opened 2024-03-15
Offline mode

### LINKED ISSUE #790: Offline mode
Would be nice to have an offline mode of Haddock3, not fetching anything from internet.

Add a new global parameter `offline` that should be used by libraries (especially `plotly`) when deciding how to embed their javascript functions

**Additional context**
Discussed with @sverhoeven

VGPReys:

- [X] You have sticked to Python. Please talk to us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [X] You wrote tests for the new code

---

Closes #790 by adding an additional global parameter `offline`, mainly used by plotly.
If set to `True`, the javascript plotly functions will be written in the generated html files.
Else, an internet call is made to obtain the corresponding script.

By default this parameter is set to False, allowing to reduce the size of the generated plots html files.
This implies the user to have an internet connection `ON` while using haddock3.


src/haddock/libs/libplots.py
```diff
@@ -8,6 +8,7 @@
 import plotly.express as px
 import plotly.graph_objects as go
 from plotly.io._utils import plotly_cdn_url
+from plotly.offline.offline import get_plotlyjs
```
rvhonorato, line 11: Please double check if this will affect the renderes - it could be that by using this the plots would only be visible in the browser and maybe not in jupyterlab, cocalc, etc.

src/haddock/libs/libplots.py
```diff
@@ -1120,15 +1206,15 @@ def report_generator(boxes, scatters, tables, step):
...
 def heatmap_plotly(
         matrix: NDFloat,
         labels: Optional[dict] = None,
         xlabels: Optional[list] = None,
         ylabels: Optional[list] = None,
         color_scale: str = 'Greys_r',  # Greys_r, gray
         title: Optional[str] = None,
-        output_fname: Path = Path('contacts.html'),
+        output_fname: Path = HEATMAP_DEFAULT_PATH,
+        offline: bool = False,
```
rvhonorato, line 1217: Why are even the labels optionals but this is not? here and in many other functions

src/haddock/modules/analysis/seletopclusts/defaults.yaml
```diff
@@ -18,3 +18,15 @@ top_models:
   long: Number of best-ranked models to select per cluster. By default, 10 models are selected. If one expects to find many clusters (e.g., in the case of vague restraints), this number can be reduced to speed up the protocol. Instead, if few clusters are expected (very specific and localized restraints) this number can be increased.
   group: analysis
   explevel: easy
+offline:
+  default: false
+  type: boolean
+  title: Remove any web calls.
+  short: Completely isolate the haddock3 run & results from internet.
```
rvhonorato, line 25: "Remove any web calls."  does not make sense.

Also how does this influence the usage of the haddock3 webapp cdn? ping @sverhoeven

rvhonorato (changes requested):

src/haddock/modules/analysis/seletopclusts/defaults.yaml
VGPReys: was already discussed with @sverhoeven, there is no issue.

rvhonorato (approved):

PR #862, opened 2024-04-16
sasascore module

### LINKED ISSUE #861: accessibility based scoring
Develop an accessibility-based scoring module that allows to do a posteriori filtering of docking models based on the accessibility of some input residues 

the user can add lists of supposedly buried/accessible amino acids. Every time a supposedly buried/accessible entity is found to be accessible (resp. buried), the score increases by one. The lower the final score the better the model (i.e. more in agreement with the input data)

**Additional context**
asked from different members of the community (see for example https://ask.bioexcel.eu/t/post-filtering-for-residue-restraints/4923/2)

mgiulini:

- [ ] Your PR is about CNS

---

Closes #861 by creating a `sasascore` module, which allows to score PDB files against existing accessibility information.

As an example, if some glycosylation sites on chain A (say residues 40 and 50) are known to be preserved upon complex formation, a penalty can be added if they are buried in the resulting model. At the same time, if some residues are known to be buried in the complex, we can impose a penalty if they are accessible.

An example application of the module:
```
[sasascore]
resdic_accessible_A = [40,50]
residic_buried_B = [22,23,24]
```

This will create a sasascore.tsv file, analogous to the other scoring tsv files.

```
structure       original_name   md5     score
cluster_1_model_1.pdb   emref_1.pdb     None    2
cluster_1_model_2.pdb   emref_2.pdb     None    2
cluster_2_model_1.pdb   emref_4.pdb     None    5
cluster_3_model_1.pdb   emref_3.pdb     None    5
cluster_2_model_2.pdb   emref_5.pdb     None    5
```

Here the score is the number of times the input information has not be satisfied (the lower the better). A file named violations.tsv is also produced, with a detailed picture of the violations:

```
structure             bur_A acc_B
cluster_1_model_1.pdb 40 22
cluster_1_model_2.pdb 50 23 
cluster_2_model_1.pdb 40,50 22,23,24
cluster_3_model_1.pdb 40,50 22,23,24  
cluster_2_model_2.pdb 40,50 22,23,24
```


amjjbonvin (commented):
I would suggest to rename this module to `sasascore`

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+    def _rearrange_output(self, output_name: FilePath, path: FilePath,
+                          ncores: int) -> None:
+        """Combine different tsv outputs in a single file."""
+        output_fname = Path(path, output_name)
+        self.log(f"rearranging output files into {output_fname}")
+        key = output_fname.stem.split(".")[0]
+        # Combine files
+        with open(output_fname, 'w') as out_file:
+            for core in range(ncores):
+                tmp_file = Path(path, f"{key}_" + str(core) + ".tsv")
```
VGPReys, line 56:
```suggestion
                tmp_file = Path(path, f"{key}_{core}.tsv")
```

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+                 path: Path,
+                 initial_params: FilePath = DEFAULT_CONFIG) -> None:
+        super().__init__(order, path, initial_params)
+
+    @classmethod
+    def confirm_installation(cls) -> None:
+        """Confirm module is installed."""
+        return
+
+    def _rearrange_output(self, output_name: FilePath, path: FilePath,
```
VGPReys, line 47:
```suggestion

    @staticmethod
    def _rearrange_output(output_name: FilePath, path: FilePath,
```

VGPReys, line 47: I would put the reading/rearranging for the tsv files in the `sasascore.py` script.
Maybe just keep this method here for the `self.log()` wrapping the thing ?

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+        output_fname = Path(path, output_name)
+        self.log(f"rearranging output files into {output_fname}")
+        key = output_fname.stem.split(".")[0]
+        # Combine files
+        with open(output_fname, 'w') as out_file:
+            for core in range(ncores):
+                tmp_file = Path(path, f"{key}_" + str(core) + ".tsv")
+                with open(tmp_file) as infile:
+                    out_file.write(infile.read())
+                tmp_file.unlink()
```
VGPReys, line 59: can't we keep the data in memory and have only one IO operation once everything is computed ?

src/haddock/modules/scoring/sasascore/defaults.yaml
```diff
@@ -0,0 +1,49 @@
...
+probe_radius:
+  default: 1.4
+  type: float
+  min: 0.0
+  max: 10.0
+  precision: 3
+  title: probe radius
+  short: probe radius
+  long: probe radius
+  group: 'scoring'
```
VGPReys, line 48:
```suggestion
  title: Probe radius
  short: Probe radius
  long: Sets the probe radius (in Angstrom) used to compute solvent accessible surface area.
  group: analysis
```

src/haddock/modules/scoring/sasascore/defaults.yaml
```diff
@@ -0,0 +1,49 @@
...
+  explevel: easy
```
VGPReys, line 49: I belive we do not want to expose this on the web for `easy` expertise level
```suggestion
  explevel: expert
```

src/haddock/modules/scoring/sasascore/sasascore.py
```diff
@@ -0,0 +1,164 @@
+"""Accessibility scoring calculations."""
+from haddock import log
+from pathlib import Path
+import pandas as pd
+from haddock.core.typing import FilePath
+from haddock.clis.restraints.calc_accessibility import (
```
VGPReys, line 6: ...
Should work...
Until removed from the code ...
...

src/haddock/modules/scoring/sasascore/defaults.yaml
```diff
@@ -0,0 +1,49 @@
...
+#
+cutoff:
+  default: 0.4
+  type: float
+  min: 0.0
+  max: 1.0
+  precision: 3
+  title: relative accessibility cutoff 
+  short: relative accessibility cutoff 
+  long: relative accessibility cutoff
```
VGPReys, line 12: A title short in my opinion

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,143 @@
+"""
+Surface accessibility scoring module.
+
+This module performs a solvent accessibility analysis based on 
+some user-defined residues that should be buried or accessible.
+
+If a supposedly buried (resp. accessible) residue is accessible (resp. buried),
+the score should increase by one. The lower the final score the more consistent
+the model with the user data.
```
VGPReys, line 9: Do we want to add here how to use the expendable parameters ?
resdic_accessible_X = []
resdic_buried_X = []

src/haddock/modules/scoring/sasascore/sasascore.py
```diff
@@ -0,0 +1,164 @@
...
+    apply_cutoff,
+    get_accessibility,
+    )
+
+
+def prettify_df(output_fname, columns, sortby=None):
```
VGPReys, line 12: typing of the function
```suggestion
def prettify_df(output_fname: str, columns: list[str], sortby: Optional[str] = None) -> None:
```

src/haddock/modules/scoring/sasascore/sasascore.py
```diff
@@ -0,0 +1,164 @@
+"""Accessibility scoring calculations."""
+from haddock import log
+from pathlib import Path
+import pandas as pd
+from haddock.core.typing import FilePath
```
VGPReys, line 5:
```suggestion
from haddock.core.typing import FilePath, Optional
```

src/haddock/modules/scoring/sasascore/sasascore.py
```diff
@@ -0,0 +1,164 @@
...
+    """A Job dedicated to the parallel running of Accscore jobs."""
+
+    def __init__(
+            self,
+            accscore_obj):
+        """Initialise AccScoreJob."""
+        self.accscore_obj = accscore_obj
+        self.output = accscore_obj.output_name
+
+    def run(self):
```
VGPReys, line 86:
```suggestion
    def run(self) -> None:
```

src/haddock/modules/scoring/sasascore/sasascore.py
```diff
@@ -0,0 +1,164 @@
...
+            for ch in self.acc_resdic:
+                if ch in a_viols:
+                    acc_str = ",".join([str(res) for res in sorted(a_viols[ch])])
+                    violations_data.append(acc_str)
+                else:
+                    violations_data.append("None")
+            self.violations.append(violations_data)
+        return
+
+    def output(self):
```
VGPReys, line 156:
```suggestion
    def output(self) -> None:
```

src/haddock/modules/scoring/sasascore/sasascore.py
```diff
@@ -0,0 +1,164 @@
...
+        self.path = path
+        self.buried_resdic = buried_resdic
+        self.acc_resdic = acc_resdic
+        self.cutoff = cutoff
+        self.data = []
+        self.violations = []
+        self.viol_output_name = viol_output_name
+        self.probe_radius = probe_radius
+
+    def run(self):
```
VGPReys, line 122:
```suggestion
    def run(self) -> None:
```

src/haddock/modules/scoring/sasascore/sasascore.py
```diff
@@ -0,0 +1,164 @@
...
+    
+    Returns
+    -------
+    acc_score : int
+        The accessibility score.
+    """
+    # filter the residues
+    acc_score = 0
+    b_viols = {}
+    a_viols = {}
```
VGPReys, line 60:
```suggestion
    acc_score: int = 0
    b_viols: dict[str, set] = {}
    a_viols: dict[str, set] = {}
```

VGPReys (dismissed):
Code base: OK

still have to test it

src/haddock/modules/scoring/sasascore/__init__.py
mgiulini: fstring will take care of the conversion!

src/haddock/modules/scoring/sasascore/__init__.py
mgiulini: good point!

VGPReys (approved):

src/haddock/clis/restraints/calc_accessibility.py
```diff
@@ -214,10 +223,12 @@ def add_calc_accessibility_arguments(calc_accessibility_subcommand):
             'PNS': 78.11,
             }
     }
+DEFAULT_PROBE_RADIUS = 1.4
 
 
 def get_accessibility(
```
rvhonorato, line 229: best to move this inside the `sasascore` module and call it here instead of the other way around since these CLIs are marked for deletion

src/haddock/modules/scoring/sasascore/__init__.py
rvhonorato: Exactly, this method's only self attribute is the logging, best to move it to the `sasascore.py` module and log it from there

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+        output_fname = Path(path, output_name)
+        self.log(f"rearranging output files into {output_fname}")
+        key = output_fname.stem.split(".")[0]
+        # Combine files
+        with open(output_fname, 'w') as out_file:
+            for core in range(ncores):
+                tmp_file = Path(path, f"{key}_" + str(core) + ".tsv")
+                with open(tmp_file) as infile:
+                    out_file.write(infile.read())
+                tmp_file.unlink()
```
rvhonorato, line 69: This pattern here looks very similar to what you had before in the `caprieval` module and that we changed to produce less files. See the implementation I did there: https://github.com/haddocking/haddock3/blob/main/src/haddock/modules/analysis/caprieval/capri.py#L905

You can extract the values directly from the scheduler object, no need to have these intermediate files

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+        buried_resdic = {
+            key[-1]: value for key, value
+            in self.params.items()
+            if key.startswith("resdic_buried")
+            }
+        acc_resdic = {
+            key[-1]: value for key, value
+            in self.params.items()
+            if key.startswith("resdic_accessible")
+            }
```
rvhonorato, line 97: imo this is very hard to read, but I know it works

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+        buried_resdic.pop("_")
+        acc_resdic.pop("_")
+        # finding the chains
+        buried_violations_chains = [f"acc_{ch}" for ch in acc_resdic.keys()]
+        acc_violations_chains = [f"bur_{ch}" for ch in buried_resdic.keys()]
+        violations_chains = acc_violations_chains + buried_violations_chains
+        # initialize jobs
+        sasascore_jobs: list[AccScoreJob] = []
+
+        for core in range(ncores):
```
rvhonorato, line 108: I'm a bit confused here, why are you doing this per core and why not per model?

src/haddock/modules/scoring/sasascore/sasascore.py
rvhonorato: Exactly, these functions should be moved here - but we can do that then deleting the restraints CLIs

src/haddock/modules/scoring/sasascore/sasascore.py
```diff
@@ -0,0 +1,165 @@
...
+        """Initialise AccScoreJob."""
+        self.accscore_obj = accscore_obj
+        self.output = accscore_obj.output_name
+
+    def run(self) -> None:
+        """Run this AccScoreJob."""
+        log.info(f"core {self.accscore_obj.core}, running AccScore...")
+        self.accscore_obj.run()
+        self.accscore_obj.output()
+        return
```
rvhonorato, line 92: Why this?

tests/test_module_sasascore.py
```diff
@@ -0,0 +1,126 @@
...
+        }
+    acc_score, b_viols, a_viols = calc_acc_score(result_dict,
+                                                 buried_resdic,
+                                                 acc_resdic)
+    assert acc_score == 0
+    assert b_viols == {"A": set()}
+    assert a_viols == {"A": set(), "B": set()}
+
+
+def test_prettiy_df():
```
rvhonorato, line 112: typo here?

rvhonorato (changes requested):

src/haddock/clis/restraints/calc_accessibility.py
mgiulini: I can move it to the restraints library, imo a better place than the module file

src/haddock/modules/scoring/sasascore/__init__.py
mgiulini: we don't want to launch hundreds/thousands of extremely short jobs, but rather concatenate them so that each core is dealing with a decent number of jobs

src/haddock/modules/scoring/sasascore/sasascore.py
mgiulini: ?

src/haddock/modules/scoring/sasascore/__init__.py
rvhonorato: This is already handled by the scheduler:

https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/libs/libparallel.py#L148

src/haddock/modules/scoring/sasascore/sasascore.py
rvhonorato: Why do you have two classes with two run methods; `AccScoreJob` and `AccScore`?

src/haddock/modules/scoring/sasascore/__init__.py
rvhonorato: Something is producing several files, and they are being merge here, right? This thing that is producing the several files is the one that should go trough the scheduler and then here instead of merging the file you just retrieve the values

src/haddock/modules/scoring/sasascore/__init__.py
mgiulini: @rvhonorato I have been thinking about this. I don't think this way of handling the tasks is a good idea (related also to the caprieval code linked above). If the running mode is not local and less_io is not active (the if clause in the caprieval output), this still amounts at creating a huge amount of files (for example 10k in the higher sampling scenarios), which can be avoided by simply parallelising internally on the number of cores (much lower in general). Additionally, for super-short tasks like this one, this probably makes things faster. what do you think?

src/haddock/modules/scoring/sasascore/__init__.py
rvhonorato: I just found out that pattern is found in many other places in the code, see here: https://github.com/haddocking/haddock3/issues/970

So please add this one to the list there and I can handle them in another PR

src/haddock/modules/scoring/sasascore/__init__.py
rvhonorato: I just found out that pattern is found in many other places in the code, see here: https://github.com/haddocking/haddock3/issues/970

So please add this one to the list there and I can handle them in another PR

src/haddock/modules/scoring/sasascore/__init__.py
mgiulini: I am checking this over and over and I think this induces code duplication everywhere. now caprieval has both the `rearrange_output` function (when less_io = false) and this `extract_data_from_capri_class` (when less_io = true)..both doing the same thing in a completely different manner.
Everything would be much easier if fewer output files were generated by each module.

rvhonorato: It does yes, but it's already part of our technological debt because of how people decided this workflow execution would be done - unfortunately nothing we can do at this point except work around it 🤷🏽

but this pattern is in many places already, just add a comment in the code and add it to this list here: https://github.com/haddocking/haddock3/issues/970

mgiulini: there is something that can be done, which is simply having much fewer files produced by each analysis module. The code duplication this `less_io` pattern introduces is not worth the advantage imo. Probably something to be discussed in a code meeting. I won't add anything to that PR for the time being.

rvhonorato: I think I understand what you mean - here is the same case as https://github.com/haddocking/haddock3/pull/928#issuecomment-2346747672, the execution mode is being overwritten on the fly so there's no need to even consider this `less_io` parameter, the easier solution is to just set it to `local`, let the scheduler split the tasks and refactor this implementation to extract directly from the class.

src/haddock/modules/scoring/sasascore/__init__.py
rvhonorato: We don't need to think about `less_io` unless it's a CNS-based module, what you are doing here is duplicating the work of the scheduler - even if you split the tasks here, when they arrive at the scheduler they are re-split, so no need to do it twice and better to let the lower function do it

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -78,12 +58,14 @@ def _run(self) -> None:
...
-        # index_list for the jobs with linear scaling
-        ncores = parse_ncores(n=self.params['ncores'], njobs=nmodels)
-        idx_list = get_index_list(nmodels, ncores)
-
+        if self.params["mode"] == "mpi":
+            input_ncores_slurm = int(os.getenv('SLURM_CPUS_ON_NODE', 1))
+            input_ncores_pbs = int(os.getenv('PBS_CPUS_ON_NODE', 1))
+            input_ncores = max(input_ncores_slurm, input_ncores_pbs)
+        else:
+            input_ncores = self.params["ncores"]
```
VGPReys, line 66: Don't we want to have this machinery set in `libs/libparallel.py` ?

So it can be used by other modules.

```python3
def get_input_ncores(params: ParamDict) -> int:
    if self.params["mode"] == "mpi":
        input_ncores_slurm = int(os.getenv('SLURM_CPUS_ON_NODE', 1))
        input_ncores_pbs = int(os.getenv('PBS_CPUS_ON_NODE', 1))
        input_ncores = max(input_ncores_slurm, input_ncores_pbs)
    else:
        input_ncores = self.params["ncores"]
    return input_ncores
```

src/haddock/modules/scoring/sasascore/__init__.py
VGPReys, line 1: Few comments:
- Why not using the scoring class `ScoringModule` ? (`from haddock/modules/scoring import ScoringModule`).
- This class has the `self.output()` method, that could replace the writing of the file in `extract_data_from_accscore_class`.
- Also scores are not propagated in the `PDBFile` objects.

src/haddock/modules/scoring/sasascore/__init__.py
mgiulini: of course, but in another PR (if we agree on this implementation)

src/haddock/modules/scoring/sasascore/__init__.py
VGPReys: But then we cannot sort models based on the computed score... 😢

mgiulini: for reference: we decided to indeed overwrite the current score attribute with the sasa score

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+from haddock.libs.libontology import PDBFile
+from haddock.libs.libutil import parse_ncores
+from haddock.libs.libparallel import get_index_list, Scheduler
+from haddock.modules.scoring.sasascore.sasascore import (
+    AccScore,
+    extract_data_from_accscore_class
+    )
+
+RECIPE_PATH = Path(__file__).resolve().parent
+DEFAULT_CONFIG = Path(RECIPE_PATH, "defaults.yaml")
```
VGPReys, line 35:
```suggestion

from haddock.core.defaults import MODULE_DEFAULT_YAML
from haddock.core.typing import FilePath
from haddock.modules import get_engine
from haddock.modules.scoring import ScoringModule
from haddock.libs.libontology import PDBFile
from haddock.libs.libutil import parse_ncores
from haddock.libs.libparallel import get_index_list, Scheduler
from haddock.modules.scoring.sasascore.sasascore import (
    AccScore,
    extract_data_from_accscore_class
    )

RECIPE_PATH = Path(__file__).resolve().parent
DEFAULT_CONFIG = Path(RECIPE_PATH, MODULE_DEFAULT_YAML)
```

VGPReys (dismissed):

VGPReys (dismissed):

src/haddock/modules/scoring/sasascore/__init__.py
rvhonorato: It should actually be in the `ScoringModule` class:

```python
class ScoringModule(BaseHaddockModule):
    """Parent class for Scoring modules."""

    def __init__(self, order: int, path: Path, initial_params: FilePath) -> None:
        super().__init__(order, path, initial_params)

        if self.params["mode"] == "mpi":
            input_ncores_slurm = int(os.getenv("SLURM_CPUS_ON_NODE", 1))
            input_ncores_pbs = int(os.getenv("PBS_CPUS_ON_NODE", 1))
            self.params["ncores"] = max(input_ncores_slurm, input_ncores_pbs)
```

src/haddock/modules/scoring/sasascore/__init__.py
```diff
@@ -0,0 +1,122 @@
...
+                )
+        except Exception as e:
+            self.finish_with_error(e)
+        if self.params["mode"] == "mpi":
+            input_ncores_slurm = int(os.getenv('SLURM_CPUS_ON_NODE', 1))
+            input_ncores_pbs = int(os.getenv('PBS_CPUS_ON_NODE', 1))
+            input_ncores = max(input_ncores_slurm, input_ncores_pbs)
+        else:
+            input_ncores = self.params["ncores"]
+        ncores = parse_ncores(input_ncores, njobs=len(models_to_score))
```
rvhonorato, line 69: You should not call `parse_ncores` here, this is already handled by the scheduler:

https://github.com/haddocking/haddock3/blob/f9f008f9ccb3c1ecc9bbaa0f1bf54325dce24559/src/haddock/libs/libparallel.py#L159-L166

rvhonorato (changes requested):

src/haddock/modules/scoring/sasascore/__init__.py
rvhonorato: Actually since this affects all modules it should be lower, most likely inside `parse_cores`

rvhonorato (approved):

VGPReys (approved):
🚀

PR #868, opened 2024-04-18
Fair

### LINKED ISSUE #744: FAIRness
It would be nice to add the FAIR actions to the repo:
https://github.com/fair-software/howfairis-github-action

https://citation-file-format.github.io/
https://github.com/citation-file-format/doi2cff

VGPReys:
Close #744

- [X] LICENSE 
- [X] FAIR badge on `README.md`
- [X] RSD badge
- [X] `CITATION.CFF` file
- [x] Add a [OpenSSF Best Practices Badge Program](https://www.bestpractices.dev/en) badge !


.github/workflows/fair-software.yml
```diff
@@ -0,0 +1,15 @@
+name: fair-software
```
rvhonorato, line 1: This action will execute [this pipeline](https://github.com/fair-software/howfairis-github-action/blob/main/action.yml) on every push - however this is not automatically update the badge.

> This GitHub action will suggest a badge visualizing compliance with the FAIR Software recommendations as described on [fair-software.eu](https://fair-software.eu/).

I don't think this action is not needed, specially not on every push, this badge is something we can manually update.

CITATION.cff
```diff
@@ -0,0 +1,47 @@
+cff-version: 1.2.0
+title: HADDOCK3
+message: >-
+  If you use this software, please cite it using the
+  metadata from this file.
+type: software
+authors:
```
rvhonorato, line 7: This is wrong.

The citation should be something that matches a DOI, there is no publication so you need to use the one deposited in zenodo - https://zenodo.org/records/10527751

Then you can create a new `CITATION.cff` using `doi2cff`:

```bash
$ pip install git+https://github.com/citation-file-format/doi2cff
$ doi2cff init https://doi.org/10.5281/zenodo.10527751
$ catCITATION.cff
```

rvhonorato (changes requested):

CITATION.cff
VGPReys: ok nice

.github/workflows/fair-software.yml
VGPReys: What would you then suggest then?
- Remove this action ?
- Update manually ?
or
```
on:
  pull_request:
    branches:
      - main
``` 
?

rvhonorato: I'd just remove it

PR #874, opened 2024-04-24
improved postprocessing scaling

### LINKED ISSUE #857: Improve scaling of postprocessing tasks
some scaling experiments on supercomputers allowed us to spot some execution bottlenecks at the postprocessing level. In this case, when the postprocessing is called at the end of a caprieval-free workflow, the parallelisation is not great, as it takes the parameter proper to the batch mode execution

It should be easy to call the postprocessing with the current running parameters (mode + number of cores) from within haddock3

**Additional context**
This should save a lot of CPU time in these occasions. Goes also along the Bioexcel lines of improving the scaling of core applications.

mgiulini:

- [ ] Your PR is about CNS

---

Closes #857 by improving the scaling of the postprocessing analysis.

1. When caprieval folders are not present in the workflow, the postprocessing analysis uses the current mode and cores in the CAPRI calculations
2. pre-caprieval model unpacking (and post caprieval model compression) receive the same parameters

PS: the comparison between the results of `analysis/topoaa-clustfcc-test.cfg` as it is impossible to get full reproducibility in that case


src/haddock/clis/cli_analyse.py
```diff
@@ -183,6 +183,26 @@ def update_paths(
...
+    choices=["local", "batch", "mpi"],
+)
+
+ap.add_argument(
+    "--ncores",
+    help="number of cores to use",
+    required=False,
+    type=int,
+    default=1,
+    # maximum allowed value for ncores is 16
```
VGPReys, line 201: why is it limited to 16 ?

VGPReys (approved):
Ready to be merged ! 🚀

src/haddock/clis/cli_analyse.py
mgiulini: good question..cannot remember a good reason for that, I will put something like 100

amjjbonvin:
Isn’t that defined in a default.yaml file?

On the lumi supercomputer I changed that value to the max number of cores per node

mgiulini:
> Changing it in the yaml file did speed up the analysis when tested in lumi - so it seems there is a max defined/used

it used to depend on the run before: when caprieval was run within the workflow the scaling was noticeable, otherwise, if no caprieval was run, the postprocessing was launching capri calculations using few cores..with this PR there will not be any difference anymore

PR #875, opened 2024-04-24
Add `codacy/codacy-coverage-reporter-action` action to `tests.yml`

rvhonorato:
This add integration with Codacy - https://app.codacy.com/gh/haddocking/haddock3/, in line with the other tools from the group

### DIFF
--- a/.howfairis.yml
+++ b/.howfairis.yml
@@ -1,0 +1,9 @@
+## Uncomment a line if you want to skip a given category of checks
+
+#skip_repository_checks_reason: <reason for skipping goes here>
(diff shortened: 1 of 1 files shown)


mgiulini:
I think these checks are way too strict..I would add this action later on in the project when the production version is stable/published and the development slows down

rvhonorato:
There are no checks, it just display things over there and this is needed for FAIRness

rvhonorato:
That's a configuration in codacy, has nothing to do with this PR

VGPReys (approved):

PR #880, opened 2024-04-26
Save outputs in case of faulty tolerance

### LINKED ISSUE #879: Having to restart module from scratch when tolerance threshold is set too low
At the end of the module, once the computation is performed, if the tolerance threshold is set too low, the models are not accessible, and one has to restart the module from scratch.

Inform the user that the tolerance threshold is not met, stop the run, but also write down the models so the user can restart the run from next module instead of having to modify the tolerance threshold and restart the module.

VGPReys:
- [ ] Your PR is about CNS

---
Closes #879 
Allows to save generated outputs even in the case when tolerance is under faulty threshold.

### DIFF
--- a/src/haddock/modules/__init__.py
+++ b/src/haddock/modules/__init__.py
@@ -285,6 +285,9 @@ class BaseHaddockModule(ABC):
         io.add(self.previous_io.output, "i")
         # add the output models
         io.add(self.output_models, "o")
+        # Save outputs
+        io.save()
+        # Check if number of generated outputs is under the tolrance threshold
         faulty = io.check_faulty()
         if faulty > faulty_tolerance:
             _msg = (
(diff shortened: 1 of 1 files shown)


src/haddock/modules/__init__.py
```diff
@@ -285,14 +285,17 @@ def export_io_models(self, faulty_tolerance=0):
         io.add(self.previous_io.output, "i")
         # add the output models
         io.add(self.output_models, "o")
+        # Save outputs
+        io.save()
+        # Check if number of generated outputs is under the tolrance threshold
```
mgiulini, line 290:
```suggestion
        # Check if number of generated outputs is under the tolerance threshold
```

mgiulini (approved):

Issue #917, opened 2024-07-01
Reduce the number of files generated


rvhonorato:

rvhonorato:
Added a new global boolean parameter called `less_io` in #918, but still the CNS scripts produce a `.seed` file, this needs to also be changed

PR #918, opened 2024-07-01
Reduce number of files in `mode = "local"`

rvhonorato:

- [x] Your PR is about CNS

---

This PR partially addresses #917 by adding a new general parameter called `less_io` - which when triggered will *not* save the `.inp` and `.out` files generated as input/output for CNS.

To achieve this I had to do some refactoring of the `libsubprocess.CNSJob.run` method to allow for polymorphism. Tried to keep it simple, if `input_file` and `output_file` are `Path` - it will follow the default old using the `TextWrapperIO` objects as input in the subprocess call.

Also had to change a bit `libparallel` - tried to simplify it for readabality and also addedd a mechanism to return the result of the `run` method of the tasks. This should make it easier for other modules using this library to not rely on filesystem operations - but I'll go deeper into that in another PR.

I've also added unittests for both `libsubprocess` and `libparallel` since there were none. Besides that I've also added an integration test for the `CNSJob`. Executing its `run` method both via file-based and memory based scheduler, to make sure it all works fine. For that I had to add a  `.pdb` and a `.psf` file, hence why the large number of lines - you can safely ignore those and filter by `.py` when reviewing


src/haddock/libs/libparallel.py
```diff
@@ -183,3 +187,7 @@ def terminate(self) -> None:
             worker.terminate()
 
         log.info("The workers terminated in a controlled way")
+
+    def get_results(self) -> list:
+        """Return the results of the tasks."""
+        return self.queue.get()
```
mgiulini, line 193: is this necessary? I don't see it being called anywhere

tests/test_libparallel.py
```diff
@@ -0,0 +1,120 @@
...
+            Task(1),
+            Task(2),
+            Task(3),
+        ],
+    )
+
+
+@pytest.fixture
+def scheduler_files():
+
```
mgiulini, line 47: can you add a couple of words explaining what this test is doing?

tests/test_libsubprocess.py
```diff
@@ -0,0 +1,133 @@
...
+
+        yield CNSJob(
+            input_file=Path("input"),
+            output_file=Path("output"),
+            cns_exec=Path(f.name),
+        )
+
+
+def test_basejob_run(basejob, mocker):
+    basejob.make_cmd = lambda: None
```
mgiulini, line 48: same as above

mgiulini (approved):
very good job, clear and useful! I would just add a couple of words of description to the tricky tests for future reference (see comments)

tests/test_libparallel.py
rvhonorato: This is the fixture, not the test - but I'll add some lines

mgiulini (approved):

VGPReys (approved):

PR #913, opened 2024-06-26
adding test to checks for duplicated parameters in defaults.yaml

VGPReys:
You are about to submit a new Pull Request. Before continuing make sure you read the [contributing guidelines](CONTRIBUTING.md) and that you comply with the following criteria:

- [x] You have sticked to Python. Please talk to us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [ ] Your code is well documented: proper docstrings and explanatory comments for those tricky parts
- [x] You structured the code into small functions as much as possible. You can use classes if there is a (state) purpose
- [x] Your code follows our coding style
- [x] You wrote tests for the new code
- [x] `tox` tests pass. *Run `tox` command inside the repository folder*
- [x] `-test.cfg` examples execute without errors. *Inside `examples/` run `python run_tests.py -b`*
- [x] PR does not add any dependencies, unless permission granted by the HADDOCK team
- [x] PR does not break licensing
- [ ] Your PR is about writing documentation for already existing code :fire:
- [x] Your PR is about writing tests for already existing code :godmode:

---
Closes #911 and #912 

Adding runtime checks to:
 - [x] evaluate duplicated parameters in `default.yaml` files.
 - [x] validate `default.yaml` parameter schemes

Adding tests to:
 - [x] evaluate duplicated parameters in `default.yaml` files.
 - [x] validate `default.yaml` parameter schemes
 - [x] test new validation functions

PR #921, opened 2024-07-03
Workaround topoaa

VGPReys:

- [ ] Your PR is about CNS

---

This PR try to find a path to be able to do a haddock3 run without having to start with `topoaa`.

Basically, now the input molecules are handled by the `ModuleIO` class (in `haddock/libs/libontology.py`), that mimic the output of an `io.json`.
Small modifications had to be applied to the `topoaa` module to fit this new behavior.


src/haddock/clis/cli_score.py
```diff
@@ -103,7 +103,64 @@ def cli(ap: ArgumentParser, main: Callable[..., None]) -> None:
...
+                )
+        if value != default_emscoring[param]:
+            print(
+                f"* ATTENTION * Value ({value}) of parameter {param} "
+                f"different from default ({default_emscoring[param]})"
+                )
+            # get the type of default value
+            default_type = type(default_emscoring[param])
+            # convert the value to the same type
+            if default_type == bool:
```
rvhonorato, line 145: the way to check for a type is `isinstance(var, type)`

src/haddock/core/typing.py
```diff
@@ -30,6 +30,7 @@
     Generic,
     Iterable,
     Iterator,
+    List,
```
rvhonorato, line 33: `List` is not used

src/haddock/gear/prepare_run.py
```diff
@@ -315,6 +315,12 @@ def setup_run(
             dec_all=True,
         )
 
+    first_module_id = list(modules_params.keys())[0]
+    if (topoaa_module_id := "topoaa.1") in modules_params.keys():
+        topology_params = modules_params[topoaa_module_id]
+    else:
+        topology_params = {}
```
rvhonorato, line 322: I see in the old version this `topoaa.1` was also present, can you explain why?

src/haddock/gear/prepare_run.py
```diff
@@ -327,24 +333,27 @@ def setup_run(
     else:
         copy_molecules_to_topology(
             general_params["molecules"],
-            modules_params["topoaa.1"],
+            topology_params,
         )
+        # copy_molecules_to_topology(
+        #     general_params["molecules"],
+        #     modules_params[first_module_id],
+        # )
```
rvhonorato, line 342: Remove the comented code

src/haddock/gear/prepare_run.py
```diff
@@ -807,40 +821,48 @@ def copy_molecules_to_data_dir(
         Whether to preprocess input molecules. Defaults to ``True``.
         See :py:mod:`haddock.gear.preprocessing`.
     """
-    topoaa_dir = zero_fill.fill("topoaa", 0)
+    # Removes digit from module name
+    # Build regex to capture '<name>.<digit>'
+    name_digit_regex = re.compile(r"(\w+)\.\d+")
```
rvhonorato, line 826: This will not capture with the module name does not have a digit, is that the intended behaviour? If you want to capture names that might not have the `.`, the regex should be `(\w+)(\.\d+)?`

src/haddock/gear/prepare_run.py
```diff
@@ -789,8 +800,11 @@ def copy_molecules_to_topology(
 
 
 def copy_molecules_to_data_dir(
-    data_dir: Path, topoaa_params: ParamMap, preprocess: bool = True
-) -> None:
+        data_dir: Path,
+        topoaa_params: ParamMap,
+        _first_module_name: str,
```
rvhonorato, line 805: missing `_first_module_name` in the docstring

src/haddock/gear/prepare_run.py
```diff
@@ -1053,7 +1075,10 @@ def populate_topology_molecule_params(topoaa: ParamMap) -> None:
     return
 
 
-def populate_mol_parameters(modules_params: ParamMap) -> None:
+def populate_mol_parameters(
+        modules_params: ParamMap,
+        topology_params: ParamMap,
```
rvhonorato, line 1080: missing `topology_params` in the docstring

src/haddock/gear/prepare_run.py
```diff
@@ -1079,7 +1104,7 @@ def populate_mol_parameters(modules_params: ParamMap) -> None:
         Alter the dictionary in place.
     """
     # the starting number of the `mol_` parameters is 1 by CNS definition.
-    num_mols = range(1, len(modules_params["topoaa.1"]["molecules"]) + 1)
+    num_mols = range(1, len(topology_params["molecules"]) + 1)
```
rvhonorato, line 1107: Why can't you just re-use `modules_params` here?

src/haddock/libs/libontology.py
```diff
@@ -233,11 +255,226 @@ def remove_missing(self) -> None:
...
+            f"No molecules could be found in `{input_molecules_dir}`"
+        # Sort them by creation date (which is also input order)
+        input_molecules.sort(key=getmtime)  # FIXME: getctime ?
+        # Set input attribute
+        self.input = input_molecules
+
+        # Set parsing variables
+        molecules_list: list[dict[int, PDBFile]] = [
+            Molecule(input_file).pdb_files
+            for input_file in self.input
```
rvhonorato, line 289: `input_file` could be either a `dict` or a `PDBFile` - this will fail if it's not a dict

src/haddock/libs/libontology.py
```diff
@@ -233,11 +255,226 @@ def remove_missing(self) -> None:
...
+            ]
+        # And fake them to be the output of the previous io
+        self.output = molecules_list
+    
+
+class Molecule:
```
rvhonorato, line 295: Why did you move this class here?

src/haddock/libs/libpdb.py
```diff
@@ -110,24 +113,64 @@ def get_supported_residues(haddock_topology: FilePath) -> list[str]:
...
+    models_starts: int = 0
+    model_ends: int = 0
+    with open(pdb_file_path, 'r') as filin:
+        for line in filin:
+            if line.startswith("ENDMDL"):
+                model_ends += 1
+            elif line.startswith("MODEL"):
+                models_starts += 1
+    nb_models = max(1, model_ends)
+    assert max(1, models_starts) == nb_models
```
rvhonorato, line 172: Why this assert here?

src/haddock/libs/libontology.py
```diff
@@ -233,11 +255,226 @@ def remove_missing(self) -> None:
...
+        """Load first molecules at the stat of the workflow.
+
+        Parameters
+        ----------
+        input_molecules_dir : Path
+            Directory where the input molecules are stored.
+        """
+        # Gather all input molecules
+        input_molecules = list(input_molecules_dir.glob('*.pdb'))
+        assert input_molecules != [], \
```
rvhonorato, line 279: I think here would be best to raise an error instead of the assert

src/haddock/modules/topology/topoaa/__init__.py
```diff
@@ -173,79 +117,74 @@ def _run(self) -> None:
...
                     self.params,
                     parameters_for_this_molecule,
                     default_params_path=self.toppar_path,
                     write_to_disk=not self.params["less_io"],
-                )
-
-                self.log("Topology CNS input created")
+                    )
+                self.log(
+                    f"Topology CNS input created in {topoaa_input.name}"
```
rvhonorato, line 158: `generate_topology` might return a `str`, so this `.name` will fail

tests/test_gear_yaml2cfg.py
```diff
@@ -97,3 +112,19 @@ def test_yaml2cfg_test_no_header():
         shallow=False,
         )
     p.unlink()
+
+
+def test_yaml_duplicated_params(default_yaml_files):
```
rvhonorato, line 117: Isn't this being done in https://github.com/haddocking/haddock3/pull/913?

tests/test_libontology.py
```diff
@@ -81,6 +86,74 @@ def io_json_file(io_data) -> Generator[Path, None, None]:
         yield Path(f.name)
 
 
+@pytest.fixture
+def molecule():
+    return Molecule(None)
```
rvhonorato, line 91: Strange this is not raising a type error, but I see your point, just add a  `# type: ignore`

tests/test_libworkflow.py
```diff
@@ -19,13 +20,14 @@ def test_WorkflowManager(caplog):
...
-            ParamDict,
-            start=0,
-            other_params=Any,
-            )
-        workflow.postprocess()
-        first_log_line = str(caplog.records[0].message)
-        second_log_line = str(caplog.records[1].message)
-        assert first_log_line == "Reading instructions step 0_topoaa"
-        assert second_log_line == "Running haddock3-analyse on ./, modules [], with top_cluster = 10"  # noqa : E501
+        with working_directory(_tmpdir):
```
rvhonorato, line 23: Why do you need to change the context here?

integration_tests/test_alascan.py
```diff
@@ -17,7 +17,7 @@ def alascan_module():
     """Return a default alascan module."""
     with tempfile.TemporaryDirectory(dir=".") as tmpdir:
         alascan = AlascanModule(
-            order=0, path=".", initial_params=DEFAULT_ALASCAN_CONFIG
+            order=1, path=Path("."), initial_params=DEFAULT_ALASCAN_CONFIG
```
rvhonorato, line 20: Why do you need to change the order here? Also in the other tests

integration_tests/test_topoaa.py
```diff
@@ -1,31 +1,58 @@
...
+        initial_params=DEFAULT_TOPOAA_CONFIG,
         )
-        topoaa.__init__(path=tmpdir, order=0)
-        topoaa.params["molecules"] = [
-            Path(DATA_DIR, "docking-protein-protein/data/e2aP_1F3G.pdb"),
-            Path(DATA_DIR, "docking-protein-protein/data/hpr_ensemble.pdb"),
-        ]
-        topoaa.params["mol1"] = {"prot_segid": "A"}
-        topoaa.params["mol2"] = {"prot_segid": "B"}
+    #topoaa.__init__(path=modulename_path, order=0)
```
rvhonorato, line 48: remove comented code

src/haddock/clis/cli_score.py
```diff
@@ -103,7 +103,64 @@ def cli(ap: ArgumentParser, main: Callable[..., None]) -> None:
 
 def maincli() -> None:
     """Execute main client."""
-    cli(ap, main)
+    cli(_ap(), main)
+
+
+def get_parameters(kwargs: dict[str, Any]) -> dict[str, Any]:
```
rvhonorato, line 109: No test for this function

rvhonorato (changes requested):
It's not simple to untangle this part of the code, as you probably noticed - great job getting into it

src/haddock/gear/prepare_run.py
VGPReys: Required by the function `populate_topology_molecule_params()`, to map the `molX` parameters info to input molecules, in the case `topoaa` is the first module.
Without this, we loose the information.

VGPReys: Added comments on the code directly

src/haddock/gear/prepare_run.py
VGPReys: All of them have a '.' at this stage, separating `module name` and `module step`.
This is performed/ensured earlier by `load()` from `haddock.gear.config`.
But indeed, your regex is more flexible.

src/haddock/gear/prepare_run.py
VGPReys: Because now molecules are stored in the `topology_params["molecules"]` instead of modules params.
We cannot point to the first module parameters, as they do not have the `molecules` parameter defined in their `defaults.yaml` and therefore not valid later...
Not easy...

VGPReys: In the case of `topoaa`, we have defined the parameter `molecules`, with expert level `hidden`...
An other trick

src/haddock/libs/libontology.py
VGPReys: Because of recursive import issues...
`libontology` is:
- defining `PDBFile` class
- Requiring the use of the `Molecule` class

Before, the `Molecule` class was defined in `structure.py`, but now, is asking for the type `PDBFile` that is defined in `libontology`.

So, to my knowledge, moving it here was the only solution.

src/haddock/libs/libpdb.py
VGPReys: In the case of an ensemble: I am counting `ENDMDL` statements to define the number of models, and I just want to make sure that the number of `ENDMDL` statements is equal to the number of `MODEL` statements.
Otherwise, there should not be any `ENDMDL` nor `MODEL` statements, and I return `1`

tests/test_gear_yaml2cfg.py
VGPReys: I guess I messed up between PRs...

tests/test_libworkflow.py
VGPReys: otherwise the temporary directory is simply not used.

integration_tests/test_alascan.py
VGPReys: This is an issue related to this new implementation, and the inheritance of `AlascanModule` <- `HaddockModule.__init__()`
- If the order is `0` (first module in the workflow): It will search for input molecules stored somewhere that is not created for this integration test and `[src/haddock/libs/libontology.py](https://github.com/haddocking/haddock3/pull/921/files/38bb99b82ddf685e8f56d2da13e007821cbf6638#diff-15b98e8f9e924d3f27cf4aa6d3afebfcea090bc0a6594271706e7e134e3db4b2) line 279 will trigger a crash.
- If not, it will not search for it, and all fine

src/haddock/libs/libontology.py
VGPReys: In this case, `input_file` is a `Path` to the input molecules from the config file.

src/haddock/core/typing.py
VGPReys: Used in `libontology`.
Imported by other scripts.

mgiulini:
I tested the PR: it works well when the workflow is made of non-CNS modules, but when CNS modules are included a workflow without topoaa badly fails at the CNS preparation steps, without catching the error. Here an example output of [contmap-test](https://github.com/haddocking/haddock3/blob/main/examples/analysis/contmap-test.cfg) removing topoaa:
```
(haddock3) UU-CW4VKWDG2H:analysis Giuli003$ haddock3 contmap-test.cfg 
[2024-07-26 15:48:10,564 cli INFO] 
##############################################
#                                            #
#                 HADDOCK 3                  #
#                                            #
##############################################

Starting HADDOCK 3.0.0 on 2024-07-26 15:48:00

Python 3.9.18 (main, Sep 11 2023, 08:25:10) 
[Clang 14.0.6 ]

[2024-07-26 15:48:14,905 libworkflow INFO] Reading instructions step 0_rigidbody
[2024-07-26 15:48:14,905 libworkflow INFO] Reading instructions step 1_clustfcc
[2024-07-26 15:48:14,905 libworkflow INFO] Reading instructions step 2_contactmap
[2024-07-26 15:48:15,343 base_cns_module INFO] Running [rigidbody] module
[2024-07-26 15:48:15,344 __init__ INFO] [rigidbody] crossdock=true
[2024-07-26 15:48:15,344 __init__ INFO] [rigidbody] Preparing jobs...
[2024-07-26 15:48:15,344 libutil INFO] Selected 5 cores to process 20 jobs, with 8 maximum available cores.
[2024-07-26 15:48:15,354 libparallel INFO] Using 5 cores
Process Worker-1:
Process Worker-2:
Process Worker-3:
Process Worker-4:
  File "/Users/Giuli003/anaconda3/envs/haddock3/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/Users/Giuli003/software/haddock3/src/haddock/libs/libparallel.py", line 88, in run
    r = task.run()
  File "/Users/Giuli003/software/haddock3/src/haddock/libs/libparallel.py", line 72, in run
    return self.function(*self.args, **self.kwargs)
  File "/Users/Giuli003/software/haddock3/src/haddock/libs/libcns.py", line 307, in prepare_cns_input
    raise ValueError(f"Topology not found for pdb {pdb.rel_path}.")
  File "/Users/Giuli003/anaconda3/envs/haddock3/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
ValueError: Topology not found for pdb ../data/0_rigidbody/1a2k_r_u.pdb.
```

rvhonorato:
The error above means you are trying to execute a CNS module without having generated the topologies, not related to the contact module

Removing `topoaa` from being the "module 0" does add this dependency

amjjbonvin:
Basically rigidbody in this case can not be run if topoaa was not first run.

Need to define dependencies...

mgiulini:
yes, precisely. we need to catch this exception at the beginning asking the user to add topoaa to the workflow

VGPReys:
Thanks for the review.
I will make some modifications to solve this problem.

rvhonorato:
Defining this dependency graph is quite complex and definitely besides the scope of this pr, could you please handle this in another?

Remember this is a beta version and these kinds of uncaught exceptions are tolerable, it's a work in progress anyway :)

mgiulini (approved):

PR #928, opened 2024-07-10
Use `less_io` parameter in `caprieval`

rvhonorato:

- [ ] Your PR is about CNS

---

This PR updates the `caprieval` module to benefit from the `less_io` parameter. Now when setting this `less_io` parameter, no "intermediary" files will be generated.

To achieve this I had to refactor `CAPRI.run` to return a copy of itself so that an instance with the updated values (`self.fnat`, `self.irmsd`, etc) can be retrieved as a simple `results` property of the `Scheduler` execution engine. 

There is a `rearrange_ss_capri_output` which is responsible for joining the intermediate files - however this function was also ranking according to the score. So I removed this logic of ranking into another `rank_according_to_score` function. Then this refactored ranking function can be used in the `extract_data_from_capri_class` helper function that does the same as `rearrange_ss_capri_output` but without loading files.

I also added integration tests for the `caprieval` module and new tests for the new functions plus some linting and type fixes.

### DIFF
--- a/src/haddock/modules/analysis/caprieval/__init__.py
+++ b/src/haddock/modules/analysis/caprieval/__init__.py
@@ -27,21 +27,17 @@ The following files are generated:
 from pathlib import Path
 
 from haddock.core.typing import Any, FilePath, Union
-from haddock.modules import BaseHaddockModule, get_module_steps_folders
-from haddock.modules import get_engine
-from haddock.modules.analysis import get_analysis_exec_mode
-
-# from haddock.core.typing import List, PDBFile
 from haddock.libs.libontology import PDBFile
-
+from haddock.libs.libparallel import Scheduler
+from haddock.modules import BaseHaddockModule, get_engine
+from haddock.modules.analysis import get_analysis_exec_mode
 from haddock.modules.analysis.caprieval.capri import (
     CAPRI,
     capri_cluster_analysis,
-    get_previous_cns_step,
+    dump_weights,
+    extract_data_from_capri_class,
     merge_data,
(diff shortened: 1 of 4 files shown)


src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -49,101 +50,123 @@ class HaddockModule(BaseHaddockModule):
...
     def _run(self) -> None:
         """Execute module."""
         # Get the models generated in previous step
         if type(self.previous_io) == iter:
             _e = "This module cannot come after one that produced an iterable."
             self.finish_with_error(_e)
 
-        models = self.previous_io.retrieve_models(
-            individualize=True
+        models = self.previous_io.retrieve_models(individualize=True)
```
VGPReys, line 82: I think because of the parameter `individualize=True`, it will never be nested

VGPReys (approved):
Great

rvhonorato:
I completely missed this:
https://github.com/haddocking/haddock3/blob/844c7ca25996af1f40a83f68a51627c4f5d3d3fc/src/haddock/modules/analysis/caprieval/__init__.py#L112
https://github.com/haddocking/haddock3/blob/844c7ca25996af1f40a83f68a51627c4f5d3d3fc/src/haddock/modules/analysis/__init__.py#L34-L53

If the execution mode is being overwriten - there's absolute no need to have this branching logic here and we can just define that the mode of analysis modules is `local` and only use `extract_data_from_capri_class` 🤦🏽‍♂️

PR #935, opened 2024-07-11
Add logic to check for incompatible parameters

rvhonorato:

- [ ] Your PR is about CNS

---

This PR adds a logic to check for parameter incompatibilities. 

For example; with #918 I added a new parameter called `less_io` however you cannot use this parameter together with `mode=batch`.

So now you can define a new `incompatible` property of a given parameter in the `haddock.modules.defaults.yaml`;

```yaml
less_io:
  # ...
  incompatible:
    mode: batch
```

Which will then be validated when the run is being setup.

### DIFF
--- a/src/haddock/gear/prepare_run.py
+++ b/src/haddock/gear/prepare_run.py
@@ -944,7 +944,7 @@ def validate_module_names_are_not_misspelled(params: ParamMap) -> None:
     return
 
 
-def validate_parameters_are_not_incompatible(params: ParamMap):
+def validate_parameters_are_not_incompatible(params: ParamMap) -> None:
     """Validate parameters are not incompatible.
 
     Args:
--- a/src/haddock/gear/prepare_run.py
+++ b/src/haddock/gear/prepare_run.py
@@ -1,4 +1,5 @@
 """Logic pertraining to preparing the run files and folders."""
+
 import difflib
 import importlib
 import itertools as it
@@ -76,6 +77,7 @@ from haddock.modules import (
     modules_category,
     modules_names,
     non_mandatory_general_parameters_defaults,
+    incompatible_params,
 )
 from haddock.modules.analysis import (
     confirm_resdic_chainid_length,
@@ -268,6 +270,8 @@ def setup_run(
         general_params,
         reference_parameters=ALL_POSSIBLE_GENERAL_PARAMETERS,
     )
+
+    validate_parameters_are_not_incompatible(general_params)
 
     # --extend-run configs do not define the run directory
     # in the config file. So we take it from the argument.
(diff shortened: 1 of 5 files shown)


src/haddock/modules/defaults.yaml
```diff
@@ -148,3 +148,5 @@ less_io:
     a network file system where I/O operations are slow.
   group: "execution"
   explevel: easy
+  incompatible:
```
amjjbonvin, line 151: Nice!  Will the syntax allow for multiple entries under the incompatible category? I assume yes

amjjbonvin (approved):

src/haddock/gear/yaml2cfg.py
```diff
@@ -214,3 +215,28 @@ def flat_yaml_cfg(cfg: ParamMap) -> ParamDict:
         new[param] = new_value
 
     return new
+
+
+def find_incompatible_parameters(yaml_file: Path) -> dict[str, dict[str, str]]:
```
VGPReys, line 220: I would change the return type to: dict[str, ParamDict]

src/haddock/gear/prepare_run.py
```diff
@@ -76,6 +77,7 @@
     modules_category,
     modules_names,
     non_mandatory_general_parameters_defaults,
+    incompatible_params,
```
VGPReys, line 80: I would change the name of that guy to: incompatible_defaults_params

src/haddock/modules/__init__.py
```diff
@@ -52,25 +53,28 @@
...
 # module where the module definition overwrites global definition. Not all
 # modules will use these parameters. It is the responsibility of the module to
 # extract the parameters it needs.
 # the config file is in modules/defaults.cfg
-non_mandatory_general_parameters_defaults = read_from_yaml_config(modules_defaults_path)  # noqa : E501
+non_mandatory_general_parameters_defaults = read_from_yaml_config(
+    modules_defaults_path
+)  # noqa : E501
+
+incompatible_params = find_incompatible_parameters(modules_defaults_path)
```
VGPReys, line 68: BIS: I would change the name of that guy to: incompatible_defaults_params

src/haddock/gear/prepare_run.py
```diff
@@ -940,6 +944,27 @@ def validate_module_names_are_not_misspelled(params: ParamMap) -> None:
     return
 
 
+def validate_parameters_are_not_incompatible(params: ParamMap) -> None:
+    """Validate parameters are not incompatible.
+
+    Args:
```
VGPReys, line 950: In the rest of the haddock3 code, we are using the `numpy` docstring format

VGPReys (dismissed):

sverhoeven:
In clustrmsd module the value of criterion parameter determines which other module parameters are ignored see https://github.com/haddocking/haddock3/blob/a11a7c37a96b12d203717941607daf36bf966eaa/src/haddock/modules/analysis/clustrmsd/defaults.yaml#L50 . Could we use incompatible parameters there as well? Or use clustrrmsd approach here?

VGPReys:
> In clustrmsd module the value of criterion parameter determines which other module parameters are ignored see
> 
> https://github.com/haddocking/haddock3/blob/a11a7c37a96b12d203717941607daf36bf966eaa/src/haddock/modules/analysis/clustrmsd/defaults.yaml#L50
> . Could we use incompatible parameters there as well? Or use clustrrmsd approach here?

I don't know, because they have default values, and only in some case (based on `criterion` value) some parameters will be used but not the other ones.

src/haddock/modules/defaults.yaml
rvhonorato: Default `yaml` format syntax applies here

amjjbonvin (dismissed):
We will have to remember to edit the default.yaml files when needed to define incompatible parameters, but now at least the machinery to check those is in place - nice!

VGPReys (approved):

amjjbonvin (approved):

AnnaKravchenko (approved):

PR #937, opened 2024-07-12
Clean intermediate files in batch mode

### LINKED ISSUE #936: Cleaning does not remove all relevant files in batch mode
**Describe the bug**

When executing haddock3 in batch mode, the cleaning of files does not remove all the generated job files and the .err files returned by slurm.  All job files are still present, even when `clean=true`. 

**To Reproduce**

Run for example the docking-protein-protein example, and look into the 01_rigidbody directory. You will find files like:

```
rigidbody_156.err
rigidbody_156.job
```

**Expected behavior**

All .err and .job files should be deleted.

VGPReys:

- [ ] Your PR is about CNS

---

Close #936 

In batch mode, more files are created (e.g: `.job` and `.err`).
They were not part of the file extensions to be removed during the cleaning process.
This is now fixed !

### DIFF
--- a/src/haddock/gear/clean_steps.py
+++ b/src/haddock/gear/clean_steps.py
@@ -65,6 +65,8 @@ def clean_output(path: FilePath, ncores: int = 1) -> None:
         '.inp.gz',
         '.out',
         '.out.gz',
+        '.job',
+        '.err',
         ]
 
     for extension in files_to_delete:


AnnaKravchenko:
Why is it `# deletes all except the first one` in line 62 in haddock3/src/haddock/gear/clean_steps.py?
I don’t understand why '.inp’ is in files_to_delete, since it is not to be deleted. From checking clean_steps.py I don’t see why ‘.inp’ should be in files_to_delete in general. 

Is there a reason for it, or can we remove it and change this `# deletes all except the first one` to just `# deletes all`?

amjjbonvin:
We want to limit the number of files generated, but still keep the first one to be able to check things (e.g. that there is no issue in the generated CNS script)

There is now a general parameter called `less_io` that avoid writing to disk all those files.
Only works in local mode though (and MPI should work too in principle)

AnnaKravchenko:
Ok, I understand that we want to keep '.inp' to be able to check that CNS runs fine, and that we still need this clean for batch mode.
 
But why then add `.inp` to `files_to_delete` to begin with? 
```
    # Files to delete
    # deletes all except the first one 
     files_to_delete = [
        '.inp',
        '.inp.gz',
        '.out',
        '.out.gz',
        '.job',
        '.err',
        ]

    for extension in files_to_delete:
        flist = glob_folder(path, extension)
        for file_ in flist[1:]:
            Path(file_).unlink()
```

Wouldn’t it be more logical to:
```
    # Files to delete
     files_to_delete = [
        '.inp.gz',
        '.out',
        '.out.gz',
        '.job',
        '.err',
        ]

    for extension in files_to_delete:
        flist = glob_folder(path, extension)
        for file_ in flist:
            Path(file_).unlink()
```
So it’s the same pattern for delete/archive/compress?

amjjbonvin:
Because we only want to keep 1 and not 1 per model

AnnaKravchenko:
Thank you, fully understrand:)

AnnaKravchenko (approved):
I see only [module]_1.job and [module]_1.err when running in haddock in batch mode, everything is as expected.  
For some reason, the outcome is the same with the main brunch. Maybe it has something to do with `less_io` Alex mentioned?

mgiulini (approved):

PR #968, opened 2024-08-06
added integration tests for scoring modules

mgiulini:
You are about to submit a new Pull Request. Before continuing make sure you read the [contributing guidelines](CONTRIBUTING.md) and that you comply with the following criteria:

- [ ] You have sticked to Python. Please talk to us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [ ] Your code is well documented: proper docstrings and explanatory comments for those tricky parts
- [ ] You structured the code into small functions as much as possible. You can use classes if there is a (state) purpose
- [ ] Your code follows our coding style
- [x] You wrote tests for the new code
- [x] `tox` tests pass. *Run `tox` command inside the repository folder*
- [x] `-test.cfg` examples execute without errors. *Inside `examples/` run `python run_tests.py -b`*
- [ ] PR does not add any dependencies, unless permission granted by the HADDOCK team
- [ ] PR does not break licensing
- [ ] Your PR is about writing documentation for already existing code :fire:
- [x] Your PR is about writing tests for already existing code :godmode:

---

Closes #967 by adding two simple integration tests for the CNS scoring modules. Simple checks on the output are performed. Due to the chaotic nature of the score calculation, the score attribute cannot be checked precisely.

PR #969, opened 2024-08-07
Enhance flexible definition in config files / defaults.yaml

### LINKED ISSUE #919: Flexible segments definition
Definition of flexible/semi-flexible segments in refinement modules is somewhat enigmatic:
One has to use parameter `nfleX` with X matching a sequential number of the molecule to be flexible, i.e. if 
`molecules =  [‘molecule1’, ‘molecule2’]`
and one wants 1 segment of molecule2 to be flexible, one needs to define parameter `nfle2 = 1` and not, for example, `nfle1`, `nfle3` etc. 
This way of defining a segment feels much less intuitive compared to a definition of the symmetrical segments. 
Plus it’s not at all explained at www.bonvinlab.org/haddock3/  

Would it be possible to define flexible and semi-flexible segments similarly to the symmetry segment, i.e. using chain/segment id? Let’s say mol2.pdb contains chain B, then, current definition of two flexible segments looks like:
```
molecules = [“mol1.pdb”, “mol2.pdb”]
# molecule 2 has 2 flexible segments
nfle2 = 2 
# 1st flexible segment of molecule 2 starts with residue 1
fle_sta_2_1 = 1 
# 1st flexible segment of molecule 2 ends with residue 4
fle_end_2_1 = 4 
# 2nd flexible segment of molecule 2 starts with residue 6
fle_sta_2_2 = 6 
# 2nd flexible segment of molecule 2 starts with residue 18
fle_sta_2_2 = 18 

```
Possible simplified definition of the flexible segment could look like:
```
# 1st flexible segment belongs to chain B
flex_seg_1 = ‘B’ 
# 1st flexible segment starts with residue 1 (within chain B)
flex_1_sta = 1 
# 1st flexible segment ends with residue 4 (within chain B)
flex_1_end = 4 
# 2st flexible segment belongs to chain B
flex_seg_2 = ‘B’ 
# 2nd flexible segment starts with residue 6 (within chain B)
flex_2_sta = 6 
# 2nd flexible segment ends with residue 18 (within chain B)
flex_2_end = 18 
```

Alternatively, a better description of current semi/flexibility definition should be provided. 

Current definition:
```
nfle1
default: 0
type: integer
title: Number of fully flexible segments
min: 0
max: 1000
short description: This defines the number of fully flexible segments.
long description: This parameter defines the number of fully flexible 
segments for the specified molecule. If >=1 then those must be defined 
manually with starting and end residue numbers in the fle_sta_* and 
fle_end_* variables.
group: flexibility
explevel: expert
```
Possible enhanced definition:
```
nfle1
default: 0
type: integer
title: Number of fully flexible segments in the 1st molecule
min: 0
max: 1000
short description: This defines the number of fully flexible segments 
of the 1st molecule.
long description: This parameter defines the number of fully flexible 
segments for the molecule that is the 1st entry in the ‘molecules' 
parameter. If >=1 then those must be defined manually with starting 
and end residue numbers in the fle_sta_* and fle_end_* variables.
group: flexibility
explevel: expert
```

amjjbonvin:

- [X] Your PR is about CNS

---

Simplified the fully flexible segments definitions, removing the per-molecule definition. 
This removes a lot of parameters from the defaults.yaml files.  
Also cleaned up the related CNS scripts, removing unused parameters.

Closes #919


amjjbonvin:
There is a reason why only `nfle` has been changed and not `nsegX`:  By having molecule-specific `nsegX` parameters we can control the flexibility per molecule with the following options:

-  -1 -> automatic definition of semi-flexible segments
-  0 -> molecule treated as rigid
-  >1 -> manual definition of semi-flexible segments.

About computing nlfe on the fly, this should be then implemented at the python level in the module and the corresponding parameters passed to CNS. You are welcome to try :-)

As for the set vs chain naming, formally the chainID comes after the residue name, while the segID comes at the end and this is what CNS used internally. This is also the reason why we should may-be stick to moleculeID in the documentation to cover both chainID and segID.

amjjbonvin:
And we are speaking here about full flexibility of defined segment, which means  flexible from the start of the simulated annealing process. This is different than the semi-flexible segments that are only treated as flexible in the last two cooling phases, first only side-chains and then side-chains and backbone.

rvhonorato:
I've added the missing integrations tests to account for this parameters, `test_mdref`, `test_emref`, `test_flexref`

integration_tests/test_emref.py
```diff
@@ -0,0 +1,110 @@
...
+    emref_module.params["fle_seg_1  "] = "A"
+
+    emref_module.params["fle_sta_2 "] = 66
+    emref_module.params["fle_end_2 "] = 71
+    emref_module.params["fle_seg_2  "] = "B"
+
+    emref_module.run()
+
+    assert Path(emref_module.path, "emref_1.pdb").exists()
+    assert Path(emref_module.path, "emref_1.out.gz").exists()
```
mgiulini, line 110: here I would add a fnat check (like the ones in https://github.com/haddocking/haddock3/pull/968) to be sure that the EM preserved more or less the contacts.

integration_tests/test_flexref.py
```diff
@@ -0,0 +1,110 @@
...
+    flexref_module.params["fle_seg_1  "] = "A"
+
+    flexref_module.params["fle_sta_2 "] = 66
+    flexref_module.params["fle_end_2 "] = 71
+    flexref_module.params["fle_seg_2  "] = "B"
+
+    flexref_module.run()
+
+    assert Path(flexref_module.path, "flexref_1.pdb").exists()
+    assert Path(flexref_module.path, "flexref_1.out.gz").exists()
```
mgiulini, line 110: same as before (but probably less strict, as flexref can change many contacts)

integration_tests/test_mdref.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def test_mdref_defaults(mdref_module):
+
+    mdref_module.previous_io = MockPreviousIO(Path(golden_data))
+
+    mdref_module.run()
+
+    assert Path(mdref_module.path, "mdref_1.pdb").exists()
+    assert Path(mdref_module.path, "mdref_1.out.gz").exists()
```
mgiulini, line 76: same as before

mgiulini (approved):

rvhonorato:
when adding `test_mdref`, `test_emref`, `test_flexref` I noticed there was some contamination between the integration tests - so here I also refactored them to avoid this.

VGPReys (approved):

rvhonorato:
People who added code should not review :)

VGPReys (approved):

VGPReys (dismissed):

VGPReys (approved):

src/haddock/modules/refinement/mdref/defaults.yaml
```diff
@@ -2184,483 +2184,19 @@ seg_end_20_1:
...
+  long: This parameter defines the number of fully flexible segments for all molecules.
+    If >=1 then those must be defined manually with starting and end residue numbers in the fle_sta_* and fle_end_* variables
+    and segment ID (segid) in the fle_seg_* variable.
   group: 'flexibility'
   explevel: expert
-fle_sta_16_1:
+fle_sta_1:
   default: .nan
   type: integer
   min: -9999
```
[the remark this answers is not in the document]
amjjbonvin: Yes a residue number can be negative

VGPReys (approved):

AnnaKravchenko (approved):

Issue #970, opened 2024-08-08
Don't split tasks in the modules


rvhonorato:
## **Desired feature/enhancement**

Refactor code that is splitting tasks in the modules

## **Motivation**

`libparallel`  recieves a set of tasks and handles the split and balancing of them, however this is also being done elsewhere in the code. 

https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/modules/analysis/ilrmsdmatrix/__init__.py#L110
https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/modules/analysis/ilrmsdmatrix/__init__.py#L308
https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/modules/analysis/rmsdmatrix/rmsd.py#L50
https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/libs/libalign.py#L1289
https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/modules/analysis/alascan/__init__.py#L63
https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/modules/analysis/rmsdmatrix/__init__.py#L88
https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/modules/analysis/rmsdmatrix/__init__.py#L141

Related implementations that generate multiple files relying on the task split to retrieve information;

https://github.com/haddocking/haddock3/blob/7a9434e2b73587d9686b31cec3920940516782de/src/haddock/modules/analysis/caprieval/capri.py#L615

https://github.com/haddocking/haddock3/blob/01eb9ae3e3c97b06248a9a90ca414d2f41c70d5f/src/haddock/modules/analysis/ilrmsdmatrix/__init__.py#L285

## **Description**

For maintanability, consistency and simplicity this should be handled in a single library.

## **Additional context**

Would be good if we could make this clear into some developer manual

rvhonorato:
[...]

PR #978, opened 2024-08-23
Propagate cluster related properties in `caprieval`

rvhonorato:

- [ ] Your PR is about CNS

---

This PR adds the extraction and combination of cluster-related information from the models into the `extract_data_from_capri_class` method. I've also updated the test to cover this behaviour and added a deprecation notice to the soon-to-be-removed branch with `less_io=false` - see #970 

Also keep in mind that this retrieval of clustering information should not be done via the model object - by doing this we are breaking the modular design by using the PDBFile object as means to transport information instead of the ontology object. Should be addressed in a separate PR tho

### DIFF
--- a/src/haddock/modules/analysis/caprieval/__init__.py
+++ b/src/haddock/modules/analysis/caprieval/__init__.py
@@ -144,6 +144,10 @@ class HaddockModule(BaseHaddockModule):
             )
 
         else:
+            self.log(
+                msg="DEPRECATION NOTICE: This execution mode (less_io=False) will no longer be supported in the next version.",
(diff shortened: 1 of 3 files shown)


VGPReys (approved):
Should fix the bug

mgiulini (approved):
does not solve the code duplication problem but solves the bug

rvhonorato:
> does not solve the code duplication problem but solves the bug

You probably mean feature duplication, since the code is different. Again see #970, I've expanded the list to also include this sort of branching that causes feature duplication in relation to running modes

PR #987, opened 2024-08-30
Refactor `validate_parameters_are_not_incompatible` to check for the incompatible value

rvhonorato:
You are about to submit a new Pull Request. Before continuing make sure you read the [contributing guidelines](CONTRIBUTING.md) and that you comply with the following criteria:

- [x] You have sticked to Python. Please talk to us before adding other programming languages to HADDOCK3
- [ ] Your PR is about CNS
- [x] Your code is well documented: proper docstrings and explanatory comments for those tricky parts
- [x] You structured the code into small functions as much as possible. You can use classes if there is a (state) purpose
- [x] Your code follows our coding style
- [x] You wrote tests for the new code
- [x] `tox` tests pass. *Run `tox` command inside the repository folder*
- [ ] `-test.cfg` examples execute without errors. *Inside `examples/` run `python run_tests.py -b`*
- [x] PR does not add any dependencies, unless permission granted by the HADDOCK team
- [x] PR does not break licensing
- [ ] Your PR is about writing documentation for already existing code :fire:
- [x] Your PR is about writing tests for already existing code :godmode:

---

This PR updates `validate_parameters_are_not_incompatible` that was not checking which value should trigger the incompatibility. Effectively it would break any workflow using the parameters mark as incompatible.

PR #991, opened 2024-09-02
Parameter incompatibilities: Increase flexibility in incompatibilities handles.

### LINKED ISSUE #986: `validate_parameters_are_not_incompatible` ignoring incompatible value

`gear.prepare_run`

## **Expected behavior**

When reading the incompatible parameter, it should check what is the value that raises the incompatibility

## **Actual behavior**

It does not check the value and breaks the execution

Run a workflow with the following `toml` file:

```toml
run_dir = "validate-bug"
mode = "batch"
concat = 1
queue_limit = 100
molecules =  [
    "data/e2aP_1F3G.pdb",
    "data/hpr_ensemble.pdb"
    ]
[topoaa]
```

Add a new `value` property to signal which is the value that raises the incompatibility

## **Version**

`2d2f02f38c08c5c1cb41869f99851b8c29974749`

VGPReys:

- [ ] Your PR is about CNS

---

Refactor the way incompatible parameters are handled and written.
Hopefully allows for more flexibility and usage outside of the global parameters.

- [x] more flexible: multiple incompatible parameters can be set for the same parameter
- [x] can be used in other modules: when validating modules parameters, incompatible ones are also checked !

Closes #986 

# How to write an incompatible parameter ?

```toml
parameter1:
  default: "value1_X"
  incompatible:
    value1_X:
      parameter2: "value2_Y"
    value1_Y:
      parameter2: "value2_X"
parameter2:
  default: "value2_X"
```

# How to read / write / understand it ?

`parameter1` has an incompatibility when its value is `value1_X` with `parameter2=value_Y`.
Also, if `parameter1` is taking the value `value1_Y`, `parameter2` cannot be set to  `value2_X`.

In this case, the default configuration file is fine, as `parameter2` is having `value2_X`.

But if the user is putting `value2_Y` for `parameter2`, an error is thrown !


src/haddock/modules/defaults.yaml
```diff
@@ -149,5 +149,5 @@ less_io:
   group: "execution"
   explevel: easy
   incompatible:
-    value: true
-    mode: batch
+    true:
```
sverhoeven, line 152: This format change will need to fixed at https://github.com/i-VRESSE/workflow-builder/blob/f59b3d1f20a04dd1afe87427eb47a9e3316eb53d/packages/haddock3_catalog/generate_haddock3_catalog.py#L128-L144

VGPReys: outch...
Maybe not the best solution then...
But nevertheless I believe it would allow more flexibility in the way we can handle incompatibilities

sverhoeven: does not hurt that bad, but it will take some time to adjust. Mostly the comment was a reminder for myself.

rvhonorato (dismissed):
Held by #987

rvhonorato (dismissed):

rvhonorato (approved):

mgiulini (approved):

Issue #993, opened 2024-09-02
Logic inconsistency when checking if file exists


rvhonorato:
## **Module/Workflow/Library affected**

All modules that use CNS's `fileexist` function;
- `emref`
- `flexref`
- `mdref`
- `rigidbody`
- `emscoring`
- `mdscoring`
- `topoaa`

## **Expected behavior**

After checking checking if a file exists, update the value of the `$result` variable

## **Actual behavior**

After checking if an empty file exists, CNS will evaluate that a file with an empty string also exists.

### CNS logic
1. Create an empty file; `touch empty-file.txt`
2. Create a non-empty file `echo 00 > non-empty-file.txt`
3. Run the following `file.cns`
```fortran
set message=off echo=off end
eval ($empty_file = "empty-file.txt")
eval ($empty_string = "")
eval ($non_existing_file = "non-existing-file.txt")
eval ($non_empty_file = "non-empty-file.txt")

! Check if the `empty_string` exists as a file
fileexist $empty_string end
display RESULT=$result ! should be FALSE, `empty_string` is not a file

! Check if the `empty_file` exists as a file
fileexist $empty_file end
display RESULT=$result ! should be TRUE, `empty_file` is a file

! Check if the `empty_string` exists as a file
fileexist $empty_string end
display RESULT=$result ! should be FALSE, `empty_string` is not a file

! Check if the `non_empty_file` file exists as a file
fileexist $non_empty_file end
display RESULT=$result ! should be TRUE, `non_empty_file` is a file

! Check if the `non_existing_file` file exists as a file
fileexist $non_existing_file end
display RESULT=$result ! should be FALSE, `non_existing_file` is not a file

! Check if the `empty_string` exists as a file
fileexist $empty_string end
display RESULT=$result ! should be FALSE, `empty_string` is not a file

stop
```

Execute:
```bash
0] $ ~/cns.linux < file.cns | grep -E 'TRUE|FALSE'
1] RESULT=FALSE ! should be FALSE, `empty_string` is not a file
2] RESULT=TRUE ! should be TRUE, `empty_file` is a file
3] RESULT=TRUE ! should be FALSE, `empty_string` is not a file
4] RESULT=TRUE ! should be TRUE, `non_empty_file` is a file
5] RESULT=FALSE ! should be FALSE, `non_existing_file` is not a file
6] RESULT=FALSE ! should be FALSE, `empty_string` is not a file
```

[...]

PR #995, opened 2024-09-02
Address issue #993

amjjbonvin:

- [X] Your PR is about CNS

Added check for empty string before `fileexist` statements in CNS code (only for instances where there is the possibility of an empty string).

### DIFF
--- a/src/haddock/modules/refinement/emref/cns/read_data.cns
+++ b/src/haddock/modules/refinement/emref/cns/read_data.cns
@@ -10,7 +10,6 @@
 ! * as part of this package.                                            *
 ! ***********************************************************************
 !
-
 ! Ambiguous and unambiguous distance restraints, hbonds
 noe
     reset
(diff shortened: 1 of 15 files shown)


rvhonorato (approved):
Amazing! Fixing the issue directly at the source instead of patching it upstream 🎉

src/haddock/modules/refinement/emref/cns/read_data.cns
```diff
@@ -23,22 +22,30 @@ fileexist $filenam0 end
...
+    end if
 end if
 
-fileexist $hbond_fname end
-if ($result eq true) then
-    noe @@$hbond_fname end
+if ($hbond_fname # "") then
+    fileexist $hbond_fname end
+    if ($result eq true) then
+        noe @@$hbond_fname end
```
VGPReys, line 47: why not `noe class dist @@$hbond_fname end` in this case ?

VGPReys (approved):

src/haddock/modules/refinement/emref/cns/read_data.cns
amjjbonvin: Good point! In haddock2.X the unambiguous file was always existing, even if empty. In haddock3 this is no longer the case. So the hbond will not inherit the dist class in case unambig_fname is not defined.

amjjbonvin: actually I will open a new issue for this

PR #1001, opened 2024-09-04
remove dead code related to DNA in `cm_restraints.cns` 

amjjbonvin:
Also added a protein-DNA cm-restrained docking test.
Checked all CNS output files for errors, none detected

- [X] Your PR is about CNS
- [X] You wrote CNS tests for the new code
- [X] `tox` or `pytest` tests pass. *Run `tox` command inside the repository folder*

---

### DIFF
--- a/examples/docking-protein-DNA/docking-protein-DNA-cmrest-test.cfg
+++ b/examples/docking-protein-DNA/docking-protein-DNA-cmrest-test.cfg
@@ -1,0 +1,71 @@
+# ====================================================================
+# protein-DNA docking example
+
+# directory in which the scoring will be done
+run_dir = "run1-cmrest-test"
+
+# execution mode
+mode = "local"
+ncores = 40
+
+# molecules to be docked
+molecules =  [
+    "data/cro.pdb",
+    "data/dna.pdb"
+    ]
+
+# ====================================================================
+# Parameters for each stage are defined below, prefer full paths
+# ====================================================================
+[topoaa]
+autohis = true
+
+[rigidbody]
+tolerance = 20
+cmrest = true
+sampling = 20
+epsilon = 78
+dielec = "cdie"
+randremoval = false
+
+[caprieval]
+reference_fname = "data/target.pdb"
+
+[seletop]
+select = 5
+
+[flexref]
+tolerance = 20
+cmrest = true
+epsilon = 78
+dielec = "cdie"
+randremoval = false
+dnarest_on = true
+
+[caprieval]
+reference_fname = "data/target.pdb"
+
+[emref]
+tolerance = 20
+contactairs = true
+randremoval = false
+dnarest_on = true
+
+[clustfcc]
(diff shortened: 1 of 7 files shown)


rvhonorato:
Could you please change the title to reflect what was actually changed in the code? I'm usually changing after you, but for this one even I can't understand

amjjbonvin:
Actually the issue description gives more details - the title cannot simply reflect the changes.

There were  undefined parameter for DNA (remaining of haddock2.X version) - but this was never noticed since we never did CM protein-DNA docking with haddock3.

The CNS part was simplified as the same behaviour is followed for ligand and nucleic acids.

examples/docking-protein-DNA/docking-protein-DNA-cmrest-test.cfg
```diff
@@ -0,0 +1,71 @@
+# ====================================================================
+# protein-DNA docking example
+
+# directory in which the scoring will be done
```
rvhonorato, line 4:
```suggestion
# TODO: Add this as an integration test
# directory in which the scoring will be done
```

rvhonorato (approved):
Keep in mind that while `examples/docking-protein-DNA/docking-protein-DNA-cmrest-test.cfg` executes a workflow that demonstrate the changes proposed in this PR, it is not actually testing for anything.

The proper way to test it would be to expand the integration tests for rigidbody and flexref, similar to what is being done here:

https://github.com/haddocking/haddock3/blob/b2dd2197d2ee996ebf1ef97a09e077b1ca026454/integration_tests/test_rigidbody.py#L134-L137

And add the assertions - I've suggested a `TODO` so this can be addressed later

examples/docking-protein-DNA/docking-protein-DNA-cmrest-test.cfg
amjjbonvin: Ok - but it was needed to actually allow to check the CNS outputs for error messages. It is like all other tests in the examples dir.  Do we have already integration tests for all other test cfg files?

For integration tests we can have much simpler workflows that one check one specific module.

amjjbonvin:
Ok - we will have then to define what to test on as for this and all other CNS-related tests, we need to check coordinates or energies in some way, allowing for some variation because of the chaotic nature of the computations. And there are so many different scenarios possible for each module.

rvhonorato:
Yep - this is already being done, here for example checking for fnat (pending on #969)

https://github.com/haddocking/haddock3/blob/a644729d6161d576c6746f5168f222ea898a68c2/integration_tests/test_flexref.py#L69-L83

Other capri metrics can also be checked and other key values can also be validated within a range

rvhonorato (approved):

rvhonorato:
oopsie, marked the wrong team - sorry guys 😅

VGPReys (approved):
Nice, this will help me writing the manual ;)

PR #1013, opened 2024-09-10
re-work haddock3 installation

rvhonorato:
This PR drastically changes the haddock3 installation.

We no longer use submodules. For `fcc` the code has already been aded to the project (#1019) and for `fastrmsd-matrix` @mgiulini gave us permission to copy the source here, thanks! 

This external code is built with a `CustomBuild` in `setup.py`.

Another main pain point is CNS, here I've added logic to determine the system's arch and download a provided pre-compiled binary - keep in mind that the installation will *fail* for unsupported architecture. I understand there might be possible to have a workflow without CNS after #921, the installation should then be reviewed.

To test it, follow these commands:

```bash
# Make a fresh clone and checkout this branch
$ git clone https://github.com/haddocking/haddock3 haddock3-installation-test
$ cd haddock3-installation-test
$ git checkout installation

# Setup the python environment
$ python -V # make sure you are on python3.9
Python 3.9.16
$ python -m venv .venv
$ source .venv/bin/activate

# With conda:
## $ conda create -n haddock3-installation-test python=3.9
## $ conda activate haddock3-installation-test

# Install haddock3
(.venv) $ pip install .

# Run one example
(.venv) $ cd examples/docking-protein-protein
(.venv) $ haddock3 docking-protein-protein-test.cfg
# This should finish without issues 🎉 
```

@haddocking/haddock-developers please look at this one very carefully


rvhonorato:
> Tried this in a conda python 3.9 environment.
> After ```pip install .``` I get the following error :
> ```
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> matplotlib 3.9.0 requires pyparsing>=2.3.1, which is not installed.
> matplotlib 3.9.0 requires numpy>=1.23, but you have numpy 1.22.2 which is incompatible.
> 
> Running the example then also does not work

I cannot replicate this, and seems like there's an issue with your package management, as written in the error message above - so it's probably related to that

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
```

"Works on my machine" both with native python and anaconda

rversin (approved):

src/haddock/fcc/__init__.py
amjjbonvin, line 1: empty file, is it needed then?

amjjbonvin (approved):

src/haddock/fcc/__init__.py
rvhonorato: yep

PR #1018, opened 2024-09-11
Detection of known CNS errors when failed run

### LINKED ISSUE #1012: Improve CNS-related error messages

A better description of the errors found during a `.inp` file execution

## **Motivation**

Currently when one of the CNS-based modules fail to produce it's output, the error is caught later during the tolerancy check, raising this `RuntimeError` exception which is not very informative (https://github.com/haddocking/haddock3/issues/1011 and others)

```
[2024-09-09 20:04:33,709 libutil ERROR] 100.00% of output was not generated for this module and tolerance was set to 5.00%.
  File "/data/foldseek/af-data/haddock3/src/haddock/libs/libutil.py", line 335, in log_error_and_exit
    yield
  File "/data/foldseek/af-data/haddock3/src/haddock/clis/cli.py", line 192, in main
    workflow.run()
  File "/data/foldseek/af-data/haddock3/src/haddock/libs/libworkflow.py", line 43, in run
    step.execute()
  File "/data/foldseek/af-data/haddock3/src/haddock/libs/libworkflow.py", line 162, in execute
    self.module.run()  # type: ignore
  File "/data/foldseek/af-data/haddock3/src/haddock/modules/base_cns_module.py", line 61, in run
    self._run()
  File "/data/foldseek/af-data/haddock3/src/haddock/modules/sampling/rigidbody/__init__.py", line 246, in _run
    self.export_io_models(faulty_tolerance=self.params["tolerance"])
  File "/data/foldseek/af-data/haddock3/src/haddock/modules/__init__.py", line 300, in export_io_models
    self.finish_with_error(_msg)
  File "/data/foldseek/af-data/haddock3/src/haddock/modules/__init__.py", line 308, in finish_with_error
    raise RuntimeError(reason)
RuntimeError: 100.00% of output was not generated for this module and tolerance was set to 5.00%.
```

## **Description**

The CNS-based modules can fail to produce output for multiple reasons, however over the years @amjjbonvin has collected several outputs related to a specific reason for failure. See here [`haddock25/tools/check-error-messages.sh`](https://github.com/haddocking/haddock25/blob/main/tools/check-error-messages.sh) (only for @haddocking/haddock-developers)

These checks can be added with a new `parse_errors` method of `CNSJob`, and integrated into it's `run` method, parsing the stdout of the `p.communicate` call

## **Additional context**

When implementing this it's important to consider efficiency as these `.out` files can be quite large and with _a lot_ of lines.

Since the errors are going to be at the end, it's a good idea to read the file backwards and stop when an error is found - for example:

```python
error_dict= {
    "exceeded allocation for NOE-restraints":  "check your active/passive definition",
    "ROTMAT error encountered: rotation vector has zero length": "try turning off the sampling of 180 degrees rotattion"
    # ... etc
}

def parse_errors(filename: str) -> str:
    with open(filename, 'rb') as file:
        file.seek(0, 2)
        size = file.tell()
        chunk_size = 4096
        buffer = b''
        for i in range(size - 1, -1, -chunk_size):
            file.seek(max(i - chunk_size, 0))
            chunk = file.read(min(chunk_size, i + 1))
            buffer = chunk + buffer
            lines = buffer.split(b'\n')
            for line in reversed(lines[1:]):
                decoded_line = line.decode('utf-8', errors='replace')
                for error, cause in error_dict.items():
                    # Check if this error is known
                    if error in decoded_line:
                        # return the cause
                        return cause
```

VGPReys:
- [x] Your PR is about CNS

---

This PR add a Known CNS error detection machinery triggered when `tolerance` threshold is not met (basically when the run will stop), which is often the case when the user will provide some weird input files.
In the module output directory, `*.cnserr` / `*.cnserr.gz` files are read and searched for known errors.

The new script `src/haddock/gear/known_cns_errors.py` holds the check functions and the list of known errors.
Can be simply updated by newly discovered error types by adding the string to search as key and the hint message to be returned to the user as values.

The `CNSJob` class has been updated to search for those errors at runtime, when STDOUT is still in memory.
If an error is detected, dump a `.cnserr` file that is later compressed, containing the error.
CNS modules have been modified to send the path to the potential `xxxx.cnserr` file, so we know where to write it.

Also has uni and integration tests to check that it is properly functional.

Closes #1012


src/haddock/gear/known_cns_errors.py
```diff
@@ -0,0 +1,157 @@
...
+from pathlib import Path
+
+from haddock.core.exceptions import KnownCNSError
+from haddock.core.typing import FilePath, Optional, Union
+
+# Dictionary of known errors
+# as key:    How to catch it in the cns.out
+# as value:  Message to user
+KNOWN_ERRORS = {
+    "CHAIN LENGHT FOR SYMMETRY RESTRAINTS DO NOT MATCH": (
```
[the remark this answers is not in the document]
VGPReys: yes, indeed:
`symmultimer.cns` files contain the typo, I will update those too

src/haddock/gear/known_cns_errors.py
```diff
@@ -0,0 +1,157 @@
...
+        "and passive residues. "
+        "Make sure to filter those for solvent accessibility."
+        ),
+    "SELRPN error encountered: parsing error": (
+        "Check your restraint files."
+        ),
+    "PARSER error encountered: Encountered too many parsing errors": (
+        "Encountered too many parsing errors."
+        ),
+    "XMREAD error encountered:  sectioning of map incompatible with resolution": (  # noqa : E501
```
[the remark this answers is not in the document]
VGPReys: Same here, this is used to catch the CNS error.

src/haddock/gear/known_cns_errors.py
```diff
@@ -0,0 +1,157 @@
...
+        "Too many distance restraints defined. "
+        "Try to reduce this number by checking your definition of active "
+        "and passive residues. "
+        "Make sure to filter those for solvent accessibility."
+        ),
+    "SELRPN error encountered: parsing error": (
+        "Check your restraint files."
+        ),
+    "PARSER error encountered: Encountered too many parsing errors": (
+        "Encountered too many parsing errors."
```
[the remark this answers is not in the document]
VGPReys: Yes, indeed.
Our cns code is not generating this error, this one is pure CNS output.
I could find a solution to point the the file (but could also be multiple of them, and therefore be a very long and boring list of files)

src/haddock/gear/known_cns_errors.py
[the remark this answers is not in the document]
VGPReys: Also, EM maps are not yet ported to haddock3.

src/haddock/gear/known_cns_errors.py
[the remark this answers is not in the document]
amjjbonvin: The hint is given in the CNS output: 
        "Check your input molecules and symmetry restraints."

amjjbonvin (dismissed):

src/haddock/gear/known_cns_errors.py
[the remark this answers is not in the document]
VGPReys: ok done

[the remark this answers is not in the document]
VGPReys: OK so this one is also a symmetry related error

src/haddock/gear/known_cns_errors.py
```diff
@@ -0,0 +1,158 @@
...
+        "Check your symmetry restraints definition."
+        ),
+    "error in SYMMETRY potential, check NOE table": (
+        "Check your symmetry restraints definition."
+        ),
+    "exceeded allocation for NOE-restraints": (
+        "Too many distance restraints defined. "
+        "Try to reduce this number by checking your definition of active "
+        "and passive residues. "
+        "Make sure to filter those for solvent accessibility."
```
amjjbonvin, line 31: Add another suggestion:
"Or alternatively increase the nres parameter in the noe statements in the relevant CNS scripts."

src/haddock/modules/__init__.py
amjjbonvin, line 1: Will this also work in local and MPI modes? And if `lessio` is set to true?
Is it performing the checks before the out files are gzipped?

src/haddock/gear/known_cns_errors.py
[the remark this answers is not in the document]
amjjbonvin: Nope - put my comment in the wrong part... :-)
This is generic to the noe function in CNS. Any distance restraints or symmetry restraints are part of the noe function.

src/haddock/modules/__init__.py
VGPReys: I works on local.
In MPI, probably too, as it reads the `*.out` files within the module directory.
I have to see what `less_io` is outputing
It is performing the checks before the out files are gzipped, just at the end of a module, and not at the end of the workflow.

VGPReys: it seems that the `*.out` files are not generated when `less_io = true`.

VGPReys: Nice catch.
The `.out` are compressed by the `CNSJob` class and not by the cleaning step.
So I should search of `.out.gz` and uncompress them when faulty !

amjjbonvin:
This will add additional I/O / CPU requirements... May-be only worth to do it if an error is detected and not systematically for all out.gz files

VGPReys:
Yes, already the case
Reading the `*.out` or `.out.gz` is only triggered when the `tolerance` threshold is not met and the run is about to fail.

src/haddock/core/exceptions.py
```diff
@@ -36,6 +39,22 @@ class CNSRunningError(HaddockError):
 
     pass
 
+class KnownCNSError(CNSRunningError):
```
rvhonorato, line 42: nice :)

src/haddock/gear/known_cns_errors.py
```diff
@@ -0,0 +1,173 @@
...
+    Raises
+    ------
+    KnownCNSError
+        An exception for known CNS errors, with its hint on how to solve it!
+    """
+    # Find file size
+    file_handle.seek(0, 2)
+    size = file_handle.tell()
+    buffer = b''
+    parsed_lines = 99999
```
rvhonorato, line 118: If I understand, this number here sets the max number of lines that will be read, right? Maybe add a comment here to make it clear for prosperity

src/haddock/gear/known_cns_errors.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Returns
+    -------
+    all_errors : dict[str, dict[str, Union[list[FilePath], KnownCNSError]]]
+        Dictionary containing all errors found in this directory.
+    """
+    all_errors: dict[str, dict[str, Union[int, KnownCNSError]]] = {}
+    # Gather list of all `.out` and `.out.gz` files present in directory
+    all_cns_out_files = list(Path(directory_path).glob("*.out.gz"))
+    all_cns_out_files += list(Path(directory_path).glob("*.out"))
```
rvhonorato, line 162: The logic of this function is correct but this is in the wrong place, it needs to go here: https://github.com/haddocking/haddock3/blob/a15dc5e7569d53421c861dcef4980149d6cc90d4/src/haddock/libs/libsubprocess.py#L170-L174

And either parse the file or the output coming out of the subprocess

src/haddock/modules/__init__.py
```diff
@@ -297,6 +300,16 @@ def export_io_models(self, faulty_tolerance=0):
...
+            # Try to detect CNS errors
+            if detected_errors := find_all_cns_errors(self.path):
+                _msg += linesep
+                for error in detected_errors.values():
+                    _msg += (
+                        # f'An error was detected in {len(error["files"])} files'
+                        # f'({",".join(error["files"][:3])}...).{linesep}'
+                        f'{str(error["error"])}{linesep}'
+                        )
+            # Show final error message
```
rvhonorato, line 312: This is also in the wrong place, need to raise this error in the base `CNSJob` class;

https://github.com/haddocking/haddock3/blob/a15dc5e7569d53421c861dcef4980149d6cc90d4/src/haddock/libs/libsubprocess.py#L93-L94

rvhonorato (changes requested):

src/haddock/modules/__init__.py
rvhonorato: In the current form this would only work if the `.out` files are generated since its being applied later in the call, when checking the tolerance. You need to check this earlier, just after the subprocess is called directly in the `CNSJob` class

rvhonorato:
> This will add additional I/O / CPU requirements... May-be only worth to do it if an error is detected and not systematically for all out.gz files

Unfortunately there's no way of knowing if the CNS subprocess failed or not without parsing something, as it will exit with status 0 even if there are erros. Consider this example:

```fortran
! test.inp, this will cause cns to exit with errors
display $non_defined_variable
```

and execute and check the exit status
```bash
$ cns < test.inp > /dev/null && echo $?
0
```

So we cannot directly check the process status to trigger the check, we need to either parse the output (object or file) and/or check if the expected output is generated.

Need to think carefully here what will be the trigger condition as this changes the implementation quite dramatically. imo it's best to check the subprocess output (object or file) directly. Doing this directly in the `CNSJob` class means that this check will be done in the context of a Task, thus it will be already paralelized so the overhead might be quite low. If you check later then it will be serialized and it might indeed create a bottleneck.

src/haddock/modules/__init__.py
VGPReys: I would keep it here, as we just want to check what errors are present when the workflow is about to terminate, and not after each CNS jobs.
So it limits the number of `.out` files that we read, and only parse them when models have not been generated

src/haddock/gear/known_cns_errors.py
VGPReys, line 118:
```suggestion
    buffer = b''
    # Set the number of lines to parse
    # initiated with high value to read all lines the first time
    # updated to the number of lines that were already parsed
    # after each chunk iteration, so we do not parse the same lines
    parsed_lines = 99999
```

amjjbonvin: Note that the errors are always appearing at the end of the out files. Anyway to specify to only read the last e.g. 1000 lines?

rvhonorato: The code is reading the file backwards and stopping when there's an error - up to a upper limit `parsed_lines = 99999`

amjjbonvin: Perfect then!

VGPReys:
⚠️  We should discuss about how to deal with `less_io = true`, especially in the case of the web application ⚠️

VGPReys:
In the end, we will read the `*.out`, `*.out.gz` files only when:

- `tolerance` is not met ( when the run is about to terminate )
- `less_io = false` (aka `debug = true`), otherwise they are not written.

This will limit the overhead by not reading the STDOUTs in memory.

For the web-app, we will need to find a clever machinery, maybe by restarting the last module that failed with `less_io = false` (or `debug = true`), with the hope of returning something useful to the user.

This PR can be merged

rvhonorato:
> In the end, we will read the `*.out`, `*.out.gz` files only when:
> 
> * `tolerance` is not met ( when the run is about to terminate )
> * `less_io = false` (aka `debug = true`), otherwise they are not written.
> 
> This will limit the overhead by not reading the STDOUTs in memory.
> 
> For the web-app, we will need to find a clever machinery, maybe by restarting the last module that failed with `less_io = false` (or `debug = true`), with the hope of returning something useful to the user.
> 
> This PR can be merged

I think this is a lazy solution that will cause problems in the future and would be better if this is addressed now, specially considering how this will affects both the throughput and the support. I don't see why new features should be half-measures. Your choice, but I won't approve it as is

VGPReys:
It is now reading the `stdout` returned by the subprocess, and writing a `.err` file when one is detected.

src/haddock/gear/clean_steps.py
amjjbonvin, line 1: Why double quote for err and not for the other extensions?

amjjbonvin (commented):
Did you test in batch modo to check that the .err files generated by slurm are not interfering with this. Those should be cleaned by the machinery ideally (if not empty)

src/haddock/gear/clean_steps.py
VGPReys: I was lazy.
best practice would tell to put all of them as double quote, but did not want to modify other ones.

amjjbonvin:
> > Did you test in batch modo to check that the .err files generated by slurm are not interfering with this. Those should be cleaned by the machinery ideally (if not empty)
> 
> Good point. Did not try. Should I modify the extension of CNS related `.err` files to `.cnserr` (or `.cns.err`) ?

Maybe indeed `.cnserr`   - then the `.err` files (if generated) can be cleaned.

src/haddock/gear/clean_steps.py
VGPReys: done

src/haddock/gear/clean_steps.py
```diff
@@ -61,12 +62,11 @@ def clean_output(path: FilePath, ncores: int = 1) -> None:
     # Files to delete
     # deletes all except the first one
     files_to_delete = [
-        '.inp',
-        '.inp.gz',
-        '.out',
-        '.out.gz',
-        '.job',
-        '.err',
```
amjjbonvin, line 69: I would still leave .err as a file to delete as it can be generated by slurm in batch mode and it makes no sense to keep it.

src/haddock/gear/clean_steps.py
```diff
@@ -61,12 +62,11 @@ def clean_output(path: FilePath, ncores: int = 1) -> None:
...
+        ".inp",
+        ".inp.gz",
+        ".out",
+        ".out.gz",
+        ".job",
```
VGPReys, line 69:
```suggestion
        ".job",
        ".err",
```

amjjbonvin (approved):

src/haddock/libs/libsubprocess.py
```diff
@@ -229,7 +243,34 @@ def run(
...
         return out
+
+    @staticmethod
+    def contains_cns_stdout_error(out: bytes) -> bool:
+        # Decode end of STDOUT
+        # Search in last 24000 characters (300 lines * 80 characters)
+        sout = out[-24000:].split(bytes(os.linesep, "utf-8"))
+        print(sout)
+        print(type(sout))
+        print()
```
rvhonorato, line 267: Should these prints be here?

src/haddock/libs/libsubprocess.py
```diff
@@ -229,7 +243,34 @@ def run(
...
+        # Reverse loop on lines (read backward)
+        for bytes_line in reversed(sout):
+            line = bytes_line.decode("utf-8")
+            if "         ^^^^^^^" in line:
```
rvhonorato, line 271: The error tag should be more generic, maybe just `"^"`

src/haddock/libs/libsubprocess.py
```diff
@@ -229,7 +243,34 @@ def run(
...
+                return True
+            # Check if a known error is found
```
rvhonorato, line 273: Add a comment here saying that this check is also needed because known errors will not have error markers

src/haddock/gear/known_cns_errors.py
```diff
@@ -0,0 +1,177 @@
...
+    else:
+        return None
+
+
+def _find_cns_errors(
+        file_handle: Union[gzip.GzipFile, BufferedReader],
+        known_errors: dict[str, str],
+        chunk_size: int = 4096,
+        filepath: FilePath = "",
+        ) -> None:
```
rvhonorato, line 95: Here you should also handle the situation in which there was no known error, maybe then raise a generic one directing people to one of the helper resources (manual/forum/etc)

src/haddock/libs/libsubprocess.py
```diff
@@ -121,12 +127,16 @@ def __init__(
         """
         self.input_file = input_file
         self.output_file = output_file
+        self.error_file = error_file
         self.envvars = envvars
         self.cns_exec = cns_exec
 
     def __repr__(self) -> str:
+        input_file = self.input_file
```
rvhonorato, line 135: maybe `_input_file` as a variable name just to avoid confusion, but good as is

src/haddock/modules/__init__.py
```diff
@@ -297,6 +300,16 @@ def export_io_models(self, faulty_tolerance=0):
                 f"{faulty:.2f}% of output was not generated for this module "
                 f"and tolerance was set to {faulty_tolerance:.2f}%."
             )
+            # Try to detect CNS errors
+            if detected_errors := find_all_cns_errors(self.path):
+                _msg += linesep
+                for error in detected_errors.values():
+                    _msg += (
+                        # f'An error was detected in {len(error["files"])} files'
+                        # f'({",".join(error["files"][:3])}...).{linesep}'
```
rvhonorato, line 309: No need for these comments here

rvhonorato (changes requested):
Much better! Just some comments

src/haddock/gear/known_cns_errors.py
VGPReys: Good point, especially that now we known that files with the `.cnserr` will always contain an error.

rvhonorato (approved):

amjjbonvin (approved):

Issue #1000, opened 2024-09-04
Reduce default tadfactor value


amjjbonvin:
## **Desired feature/enhancement**

During the torsion angle dynamics in the `flexref` module, the time steps are increased by a factor defined by `tadfactor` for the rigid body stages of the protocol.  This in combination with fully flexible segment definitions might lead to failures (the MD becomes unstable and crashes). In haddock2.X those cases were handled by the python code and the model was rerun with a smaller factor.

The `flexref` CNS code should automatically detected the presence of fully flexible segments and reduce the `tadfactor` from the default value of 8 to e.g. 4.

## **Motivation**

This will cause less model failures and more stable runs.

Shadi201711:
try this
https://gofile.io/d/yMHabA
select the desired compiler in the installer

amjjbonvin:
Best to implement this once issue #969 has been solved/merged.

amjjbonvin:
Will add an integration test once the integration-tests branch is merged, using the test_flexref.py as basis.

PR #1020, opened 2024-09-13
Automatically reduce the tadfactor to 4 when fully flexible segments are defined

amjjbonvin:

Solves issue #1000.
In the `flexref` module, the tad factor defining the time step for an MD integrations step is now automatically reduced from the default of 8 to 4 for the high temperature initial stage of the simulated annealing protocol. This will reduce possible job failures (models crashing) in the flexref module.

Integration test will be added once the `integration-tests` branch is merged.

### DIFF
--- a/src/haddock/modules/refinement/flexref/cns/flexref.cns
+++ b/src/haddock/modules/refinement/flexref/cns/flexref.cns
@@ -366,6 +366,10 @@ if ($failure eq false) then
 
     {* Stage 1 ======================================= rigid body high temperature search*}
 
+    if ($nfle > 0) then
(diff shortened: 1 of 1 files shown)


VGPReys (dismissed):
This overwrites any user defined values of this `tadfactor` parameter?
Or it applies only at high temperatures ?

VGPReys (approved):

amjjbonvin:
> This overwrites any user defined values of this `tadfactor` parameter? Or it applies only at high temperatures ?

In the current implementation it will take the minimum value of the user-defined value and 4 for the high temperature stage only. For the last two stages with flexible side-chains and backbone the value was already set to 1

Issue #1027, opened 2024-09-16
[attract] as a sampling module


AnnaKravchenko:
## **Desired feature/enhancement**
Add a rigid fragment-based protein-ssRNA doking pipeline, performed by the ATTRACT docking engine, as a sampling module of HADDOCK.

## **Motivation**
1. Introduces a dedicated docking workflow for protein-ssRNA complexes within HADDOCK, eliminating the need for CNS-based sampling;
2. Could potentially be executed in an anchored manner, when a single fragment is docked using restraints (obtained using experimental data, etc.), and consecutive fragments are docked using restraints towards the previous fragment(s); 
3. Could potentially be expanded to protein-protein and protein-DNA docking (supported by ATTRACT but outside of my current competency).

## **Description**
[attract] module should handle the full docking workflow, consisting of the following steps:
1. Preparation of the input folder, protein coarse-graining
2. Sampling, a per-fragment rigid-body energy minimisation, with or without restraints
4. Scoring with ATTRACT scoring function or with HIPPO
5. Assembly of RNA fragments into a full-chain docking models
6. Conversion of a limited number of models back to all-atom representation

## **Additional context**
Paper about [fragment-based protein-ssRNA doking](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1004697) 
Paper about [anchored docking ](https://academic.oup.com/nar/article/44/10/4565/2517079#96182036)

PR #1034, opened 2024-09-18
[attract] module

AnnaKravchenko:

## Checklist

So far [attract] module can:
- Make sure ATTRACT is installed
- Coarse-grain protein and RNA
- Fragment RNA and prepare required for ATTRACT files. 

It requires strictly 1 protein and 1 RNA molecules as an input. 

Installation guidlines are in`haddock3/src/haddock/modules/sampling/attract/install.md`

## Related Issue
[#1027](https://github.com/haddocking/haddock3/issues/1027)

## Additional Info
Currently required environment variables:
```
$ATTRACTDIR=/trinity/login/arha/tools/attract/bin
$ATTRACTTOOLD=/trinity/login/arha/tools/attract/tools
$LIBRARY=/trinity/login/arha/tools/nalib 
$RANDSEARCH=/trinity/login/arha/data/ssRNA/randsearch
```

If you import these variables - [attract] may run* on tintin. 
*now only create required by ATTRACT folder structure and pass two models to the next module


AnnaKravchenko:
* Add instructions on how to install ATTRACT
* Move small functions in separate .py 
* Move 'nalib check' to ‘general installation check’ with corresponding message 
* Try to do all I need to do with pdb within 1 function (to not open/close file multiple time)
* Use sys.executable rather than subprocess.run 
* Be more consistent with the style (particularly replace `open(file,'r’)` with `with open(file,w) as`)
* Think if it can/should be a class?

src/haddock/modules/_template_cat/README.md
```diff
@@ -4,25 +4,25 @@ To develop your own HADDOCK3 module follow the patterns used in other modules.
 
 1. If your module belongs to a new category, create a folder for that category
    under the `modules` folder.
-1. Else, create a folder for your module inside its relevant category.
+2. Else, create a folder for your module inside its relevant category.
```
rvhonorato, line 7: This is on-purpose, markdown will render it numerically independent of the sequence

AnnaKravchenko: I know, but also it is not displayed properly in a simple text editor, while actual numeration looks good in both type of text editors

Issue #1038, opened 2024-09-20
In offline mode replace inline plotlyjs with import


sverhoeven:
## **Desired feature/enhancement**

Running a caprieval analysis in offline mode uses lots of diskspace.

For example I ran https://github.com/haddocking/haddock3/blob/main/examples/scoring/capri-scoring-test.cfg with `offline = true`.
<details>
<summary>The capriscoring-test/analysis/7_caprieval_analysis directory uses 136Mb.</summary>

```
3.6M    air_clt.html
3.6M    bsa_clt.html
4.0K    capri_clt.tsv
4.0K    capri_ss.tsv
3.6M    desolv_clt.html
3.6M    dockq_air.html
3.6M    dockq_clt.html
3.6M    dockq_desolv.html
3.6M    dockq_elec.html
3.6M    dockq_score.html
3.6M    dockq_vdw.html
3.6M    elec_clt.html
3.6M    fnat_air.html
3.6M    fnat_clt.html
3.6M    fnat_desolv.html
3.6M    fnat_elec.html
3.6M    fnat_score.html
3.6M    fnat_vdw.html
3.6M    ilrmsd_air.html
3.6M    ilrmsd_clt.html
3.6M    ilrmsd_desolv.html
3.6M    ilrmsd_elec.html
3.6M    ilrmsd_score.html
3.6M    ilrmsd_vdw.html
3.6M    irmsd_air.html
3.6M    irmsd_clt.html
3.6M    irmsd_desolv.html
3.6M    irmsd_elec.html
3.6M    irmsd_score.html
3.6M    irmsd_vdw.html
3.6M    lrmsd_air.html
3.6M    lrmsd_clt.html
3.6M    lrmsd_desolv.html
3.6M    lrmsd_elec.html
3.6M    lrmsd_score.html
3.6M    lrmsd_vdw.html
7.2M    report.html
3.6M    score_clt.html
160K    summary.tgz
3.6M    vdw_clt.html
```

</details>

It would be nice if offline caprieval analysis html pages would use less space.

In #1025 I am importing report.bundle.js file so each caprieval module can reuse that js file. I think the same idea can be applied to the plotly at https://github.com/haddocking/haddock3/blob/64d625bce6826be0164bc678544cd420bc9f49ef/src/haddock/libs/libplots.py#L131
By instead of embedding the plotlyjs write it to a file once and importing it. 

## **Motivation**

The upside is going from 136Mb to 3.6Mb (the size of a single plotly js file)
[...]

PR #1041, opened 2024-09-23
Import plotly offline mode

VGPReys:

## Checklist

Modify the `offline` mode for plotly to reduce the overall size of the js bundle by witting it in a separated file that is then imported by all the plots.

## Related Issue

#1038


amjjbonvin (dismissed):

src/haddock/libs/libplots.py
```diff
@@ -1433,33 +1453,70 @@ def fig_to_html(
...
+
+    Returns
+    -------
+    plotly_js_import : str
+        HTML solution for the importation of the plotly javascript content.
+    """
+    # Case where offline isrequired
+    if offline:
+        # Obtain directory where the figure should be written
+        fig_dir = Path(fpath).parent
```
sverhoeven, line 1504: Lot of the examples have multiple caprieval modules. 

This will cause a plotly.js file in analysis/XX_caprieval, 

would be nice if the plotly.js could be shared between modules to reduce size even more.

VGPReys: I totally agree with that.
Also other modules generating plots will have to use this file.

the issue is:
- I do not know how to access the relative path of it.
- for analysis directory, it would be `../../data/jsbundle/plotly.js`
- for other modules, it would be `../data/jsbundle/plotly.js`

I could not figure it out yet...

sverhoeven: When code is run the cwd is run_dir/XX_caprieval_analysis so the path on fs and in http is different see https://github.com/haddocking/haddock3/blob/d24a510d70727488465ce28f01a5bef3a649b92c/src/haddock/libs/libplots.py#L994-L995

rvhonorato: > I totally agree with that. Also other modules generating plots will have to use this file.
> 
> the issue is:
> 
> * I do not know how to access the relative path of it.
> * for analysis directory, it would be `../../data/jsbundle/plotly.js`
> * for other modules, it would be `../data/jsbundle/plotly.js`
> 
> I could not figure it out yet...

Why not something simple like this:

```python
def find_file(filename: str) -> Optional[Path]:
    search_paths = [Path('..').parent, Path('..')]
    for path in search_paths:
        full_path = path / filename
        if full_path.is_file():
            return full_path
    return None
```

VGPReys: Its an idea, that must be fine tuned, but why not

VGPReys: OK, tried an implementation following the suggestion idea

VGPReys:
Maybe the search of the `data/ui` is not the best solution, and having the `plotly.js` locally is a simpler solution that would also allow to detach the directory from the `rundir` and still be functional.
I guess I will revert the commits to [bdbb596](https://github.com/haddocking/haddock3/pull/1041/commits/bdbb596ad9a4671335a9aff02a0ac9cd3514d0bd) version

mgiulini (commented):
looks good and does the job! CI is broken though

src/haddock/libs/libplots.py
```diff
@@ -1433,33 +1453,70 @@ def fig_to_html(
...
-def make_traceback_plot(tr_subset, plot_filename):
+def offline_js_manager(fpath: FilePath, offline: bool) -> str:
+    """Build string to access plotly javascript content.
+
+    Parameters
+    ----------
+    fpath : FilePath
+        Path to the figure about to be written.
+    offline : bool
+        Should this figure use the offline approache, or not.
```
VGPReys, line 1494:
```suggestion
        if True use the offline approach.
```

rvhonorato (approved):
lgtm - but what is the point of having all this code to handle an offline scenario? what would be the situation in wich someone doing data analysis will not have access to the internet?

VGPReys:
> lgtm - but what is the point of having all this code to handle an offline scenario? what would be the situation in wich someone doing data analysis will not have access to the internet?

- Mainly companies, that often have a closed system to prevent intrusions or data leakage.
- Also Alex in a plane 🛩️  😄

mgiulini (approved):

PR #1053, opened 2024-09-25
Further implementing the topocg CNS scripts

amjjbonvin:

## Checklist

Fixed the CNS scripts of the topocg module, adding missing components.

Added an example (docking-protein-protein-CG) as working dir to further develop the coarse graining option.

topoaa and topocg step are now working.

Does not yet fix the DSSP issue - but bypasses it


amjjbonvin:
PS: The large number of lines comes mainly from newly added examples and missing CNS scripts.

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -493,15 +493,15 @@ def martinize(input_pdb, output_path, skipss):
     pdbf_path = os.path.realpath(input_pdb)
     aa_model = p.get_structure("aa_model", pdbf_path)
 
-    # set ALL bfactors to 0
+    # set ALL bfactors to 1
     for model in aa_model:
         for chain in model:
             if chain.id == " ":
                 print("+ ERROR: Empty chain id detected")
                 exit()
```
VGPReys, line 501: I would rather add a `raise ModuleError` here instead of the `exit()` to terminate the run properly.
requires:
- to import the custom exception at the top of the file:
```python3
from haddock.core.exceptions import ModuleError
```
- to catch this error when calling `martinize()`

src/haddock/modules/topology/topocg/defaults.yaml
VGPReys, line 1: I still think that both `ligand_param_fname` and `ligand_top_fname` should be removed, as yet no possibility to handle these with martini (from my knowledge)

amjjbonvin: Yes possibly - but at this time I would keep them to have the machinery in place if needed

rversin: If we want to add this option we will have to add them in the parameter files probably. And there will be a very limited number of ligand. Additionally, not the ligands are supported depended on which version of Martini (based on what I found in the gromacs parameter files of Martini).

src/haddock/modules/topology/topocg/aa2cg.py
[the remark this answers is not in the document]
rversin: These are things I could work on once @amjjbonvin finished fixing CNS issues :)

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -418,8 +418,8 @@ def determine_ss(structure, skipss, pdbf_path):
                 dssp = DSSP(model, pdbf_path)
             except:  # TODO: think about making this exception more specific
                 # no secondary structure detected for this model
-                print("+ ERROR: SS could not be assigned, check under the hood or add --skipss")
-                exit()
+                print("+ ERROR: SS could not be assigned, assigning code 1 to all residues")
+                continue
```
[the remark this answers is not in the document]
amjjbonvin: will fix those issues in a separate branch - for now the focus is on getting the CNS machinery working.

src/haddock/modules/topology/topocg/aa2cg.py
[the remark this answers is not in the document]
amjjbonvin: will fix those issues in a separate branch - for now the focus is on getting the CNS machinery working.

src/haddock/modules/topology/topocg/defaults.yaml
amjjbonvin: The ligand.param and ligand.top files allow to add parameters separately of the main topo/param files. We should keep this flexibility for convenience, even if not yet used.

amjjbonvin:
Tests now passing :-)   Since I deleted old Martini param/top files I had to adapt the core module that checks the list of supported amino acids and bases.

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -19,13 +19,15 @@
 import random
 import subprocess
 import warnings
+import logging
```
rvhonorato, line 22: Here you are using python's logger, it should be using haddock one instead

`from haddock import log`

And then adapt the calls in the code

rvhonorato (changes requested):

src/haddock/modules/topology/topocg/aa2cg.py
amjjbonvin: Done

rvhonorato (approved):

Issue #1056, opened 2024-09-27
Mismatch between readme and available config files for cyclic peptides


VGPReys:
## **Module/Workflow/Library affected**

In the cyclic-peptides, one example is missing,
but all the input files are there

## **Expected behavior**

Add an other example using the 3WNE structure

VGPReys:
closed by #1057

PR #1057, opened 2024-09-27
update cyclic peptides examples

VGPReys:

Add an additional example for cyclic peptides

## Related Issue

#1056


VGPReys:
I wonder why we use `unambig_fname = "data/3wne_unambig.tbl"` in the last `[flexref]`, while the bond should be formed by the topology and minimized by then ?

amjjbonvin (approved):

amjjbonvin:
Probably as a backup in case the bond was not yet formed for some reason. Doesn't hurt but could be removed. The two examples should be consistent.

rvhonorato (approved):

PR #1085, opened 2024-10-08
Add pipeline to publish image to `ghcr.io`

rvhonorato:

## Checklist

This PR adds a `Dockerfile` and a github action to automatically build and publish the image to the `ghcr.io` when a new tag is created in the repository.

Adding this pipeline will make the integration much simpler with the web-app and also on other tools.

With it people can also execute haddock via docker without any installation needed (instructions to be added in another PR) - something like
```
docker run -v `pwd`/data ghcr.io/haddock3 <configuration-file.toml>
```

And easily build a self-contained executable-like singularity/apptainer image with
```
apptainer build haddock3.sif docker://ghcr.io/haddock3:latest
```

And then just run it as
```
haddock3.sif <configuration-file.toml>
```


Dockerfile
VGPReys, line 1: There is no need to specify the linux distribution ?
Nor to install the libc / gcc / gfortan libs ?

Also the only entry point is haddock3 ?
all other CLI cannot be used ?

sverhoeven: Correct you have to use `--entrypoint haddock3-analyse` to run something else.

rvhonorato: well yes, you need to override the entrypoint to use another entrypoint :)


> There is no need to specify the linux distribution ?
> Nor to install the libc / gcc / gfortan libs ?

Of course not, why would things be necessary?

VGPReys: to run CNS and compile clustfcc ?

rvhonorato: ???

rvhonorato: Why would you need the gfortran libs to run cns? haha and the build-essentials are default in the python image

rvhonorato: Sure, but here we are using the pre-compiled static binary, so no need for any links

rvhonorato: I think it's just a lack of understanding how conteinarization works in general - this is a very good first resource for you to learn @VGPReys https://docs.docker.com/get-started/

rvhonorato: I've had this conversation with @amjjbonvin already and he did make new binaries that were static - another one for him to fix them.

But as for this PR, works fine

rvhonorato: Just checked https://github.com/haddocking/haddock3/pkgs/container/haddock3/286666577?tag=2024.10.0b6

```
$ docker pull ghcr.io/haddocking/haddock3:2024.10.0b6
$ docker image ls | grep haddock3
ghcr.io/haddocking/haddock3   2024.10.0b6   68cb10f2206f   35 minutes ago   1.96GB
```

So ~2Gb without any of the extra dependencies

PR #1088, opened 2024-10-09
Remove `self.params["debug"]` logic in `caprieval`

### LINKED ISSUE #1081: wrong numbers in capri_clt
caprieval capri_clt files

## **Expected behavior**
the capri_clt files should be consistent with the capri_ss files

## **Actual behavior**
when debug = false (default settings) haddock3 gives wrong numbers in capri_clt files

technically, the capri jobs are not being merged properly here https://github.com/haddocking/haddock3/blob/97dfc828312908e9dc234be1f0fc8f4733589c14/src/haddock/modules/analysis/caprieval/__init__.py#L144-L149 and this causes this other function to mix up things https://github.com/haddocking/haddock3/blob/97dfc828312908e9dc234be1f0fc8f4733589c14/src/haddock/modules/analysis/caprieval/capri.py#L1020-L1021

1. Go to the docking-protein-glycan/ example directory
2. Run the following command: `haddock3 docking-protein-glycan-test.cfg`
3. open the capri_ss file `head run1-test/2_rigidbody/capri_ss.tsv`
4. open the capri_clt file `head run1-test/2_rigidbody/capri_clt.tsv`
5. realise the numbers do not match at all

jobs have to be merged properly in caprieval! otherwise the zipping fails

rvhonorato:

## Checklist

This PR removes the logic path that used the `less_io` parameter - which was then changed to `debug` - to decide how to retrieve the information from the Scheduler.

It also fixes #1081 just by changing the type of `CAPRI.identificator` to an integer, this ensures the `engine.result` list will be in the expected order. To check for this bug I added a new integration test `test_ss_clt_relation` that checks if the metrics in the `capri_ss.tsv` and `capri_clt.tsv` match, so I also had to add a bunch of PDBs as golden data.


integration_tests/test_caprieval.py
```diff
@@ -284,6 +297,62 @@ def _check_capri_clt_tsv(
     _compare_polymorphic_data(expected_data, oberseved_data)
 
 
+def _check_means_match(
+    capri_ss_f: Path,
+    capri_clt_f: Path,
+    target_metric: str,
+    top_n: int,
+):
```
VGPReys, line 305:
```suggestion
) -> None:
```

src/haddock/modules/analysis/caprieval/capri.py
VGPReys, line 1: debug is still present in the CAPRI class:
- l.180
- l.226
- l.655
Even tho functional as using default kwargs `debug = False`

mgiulini (approved):

PR #1090, opened 2024-10-09
Correction of the usage of DSSP in the topocg module

rversin:

## Checklist

Here, I adapt the code to produce DSSP usable pdb files and remove the usage of the DSSP function from the BioPython library, instead using the command line version.

## Related Issue
Not related to an issue

## Additional Info
The warning "No SS definition found for residue: ..." was removed, since if DSSP is installed and works, a secondary structure will be applied. 
I will look into the Codacy reports issues with the aa2cg.py script for a following pull request.

### DIFF
--- a/src/haddock/modules/topology/topocg/aa2cg.py
+++ b/src/haddock/modules/topology/topocg/aa2cg.py
@@ -32,7 +32,7 @@ from haddock.modules.topology.topocg.helper import *
 
 warnings.filterwarnings("ignore")
 
-CRYST_LINE = "CRYST1 \n"
+CRYST_LINE = "CRYST1 " + os.linesep
 
 
 def add_dummy(bead_list, dist=0.11, n=2):
--- a/src/haddock/modules/topology/topocg/aa2cg.py
+++ b/src/haddock/modules/topology/topocg/aa2cg.py
@@ -441,13 +441,13 @@ def determine_ss(structure, output_path, skipss, pdbf_path):
                 p = subprocess.Popen(["dssp", tmp_file_name, "--output-format", "dssp"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
                 dssp_raw, errors = p.communicate()
                 dssp_raw = dssp_raw.split('#')[1].split('\n')[1:-1]
-                if Path(tmp_file_name).exists(): Path(tmp_file_name).unlink()
             except:  # TODO: think about making this exception more specific
                 # no secondary structure detected for this model
-                if Path(tmp_file_name).exists(): Path(tmp_file_name).unlink()
                 log.warning('SS could not be assigned, assigning code 1 to all residues')
                 continue
-        
+            finally:
+                if Path(tmp_file_name).exists(): Path(tmp_file_name).unlink()
+
         dssp = {}
 
         for line in dssp_raw:
@@ -467,7 +467,7 @@ def determine_ss(structure, output_path, skipss, pdbf_path):
 
             # transform MARTINI > HADDOCK
             # and add it to the bfactor col
-            for residue, ss in zip(chain, martini_types):
+            for residue, ss in zip(chain, martini_types): # chain and martini_types order must match
                 code = ss_to_code[ss]
                 # for atom in residue.get_atoms():
                 for atom in residue:
--- a/src/haddock/modules/topology/topocg/aa2cg.py
+++ b/src/haddock/modules/topology/topocg/aa2cg.py
@@ -436,12 +436,12 @@ def determine_ss(structure, skipss, pdbf_path):
     """
     # calculate SS
     for model in structure:
-        tmp_file_name = create_file_with_cryst(pdbf_path)
 
         if skipss:
             continue
         else:
             try:
+                tmp_file_name = create_file_with_cryst(pdbf_path)
                 p = subprocess.Popen(["dssp", tmp_file_name, "--output-format", "dssp"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
                 dssp_raw, errors = p.communicate()
                 dssp_raw = dssp_raw.split('#')[1].split('\n')[1:-1]
--- a/src/haddock/modules/topology/topocg/aa2cg.py
+++ b/src/haddock/modules/topology/topocg/aa2cg.py
@@ -401,21 +401,26 @@ def extract_groups(pair_list):
     out.close()
 
 
-def create_file_with_cryst(pdb_file: str, pdb_file_copy: str) -> None:
+def create_file_with_cryst(output_path: str, pdb_file: str) -> None:
     """
     This function creates a new pdb because the CRYST line is missing from the pdf file.     
     This line is necessary for DSSP.
 
     Args:
+        output: str
         pdb_file: str
-        pdb_file_copy: str
 
     Returns:
+        pdb_file_copy: str
 
     """
+    pdb_file_copy = f"../{output_path}/{pdb_file.split('/')[-1][:-4]}_tmp.pdb"
+    
     with open(pdb_file, "r") as file_in, open(pdb_file_copy, "w") as file_out:
         content = file_in.read()
         file_out.write(CRYST_LINE + content)
+    
+    return(pdb_file_copy)
 
 
 def determine_ss(structure, output_path, skipss, pdbf_path):
@@ -427,12 +432,12 @@ def determine_ss(structure, output_path, skipss, pdbf_path):
         pdbf_path:
 
     Returns:
+        structure:
 
     """
     # calculate SS
     for model in structure:
-        tmp_file_name = f"../{output_path}/{pdbf_path.split('/')[-1][:-4]}_tmp.pdb"
-        create_file_with_cryst(pdbf_path, tmp_file_name)
+        tmp_file_name = create_file_with_cryst(output_path, pdbf_path)
 
         if skipss:
             continue
--- a/src/haddock/modules/topology/topocg/aa2cg.py
+++ b/src/haddock/modules/topology/topocg/aa2cg.py
@@ -19,6 +19,7 @@ import os
 import random
 import subprocess
 import warnings
+from pathlib import Path
 from haddock import log
 
 from Bio.PDB import Entity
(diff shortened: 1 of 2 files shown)


src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -399,7 +400,27 @@ def extract_groups(pair_list):
...
+
+    Returns:
+
+    """
+    file_in = open(pdb_file, "r")
+    file_out = open(pdb_file_copy, "w")
+    content = file_in.read()
+    file_out.write(CRYST_LINE + content)
+    file_in.close()
+    file_out.close()
```
rvhonorato, line 420: It's best to use the `with` construction when handling files - it automatically ensures the files are closed even if there is any error, its overall better resource management.

Here since you have two opens, you can use both in one line;


```python
with open(pdb_file, "r") as file_in, open(pdb_file_copy, "w") as file_out:
    content = file_in.read()
    file_out.write(CRYST_LINE + content)
```

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -410,45 +431,49 @@ def determine_ss(structure, skipss, pdbf_path):
     Returns:
 
     """
+
     # calculate SS
     for model in structure:
+        tmp_file_name = f"../{output_path}/{pdbf_path.split('/')[-1][:-4]}_tmp.pdb"
+        create_file_with_cryst(pdbf_path, tmp_file_name)
```
rvhonorato, line 438: Here it's best to also use python to generate the temporary file:

```python
tmp_file_name = tempfile.NamedTemporaryFile(delete=False, suffix=".pdb").name
```

This will make sure the temporary file always exists and it's always valid, regardless of paths and such

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -410,45 +431,49 @@ def determine_ss(structure, skipss, pdbf_path):
...
         else:
             try:
-                dssp = DSSP(model, pdbf_path)
+                p = subprocess.Popen(["dssp", tmp_file_name, "--output-format", "dssp"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
             except:  # TODO: think about making this exception more specific
                 # no secondary structure detected for this model
                 log.warning('SS could not be assigned, assigning code 1 to all residues')
                 continue
+        
+        dssp_raw, errors = p.communicate()
```
rvhonorato, line 450: This `communicate` needs to go just after the `subprocess.Popen`, else it can become unbound

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -399,7 +400,27 @@ def extract_groups(pair_list):
     out.close()
 
 
-def determine_ss(structure, skipss, pdbf_path):
+def create_file_with_cryst(pdb_file, pdb_file_copy):
```
rvhonorato, line 403: Best to add the type signatures here:

```python
def create_file_with_cryst(pdb_file: str, pdb_file_copy: str) -> None:
```

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -24,14 +24,15 @@
...
 from Bio.PDB import PDBParser
-from Bio.PDB.DSSP import DSSP
 from Bio.PDB.StructureBuilder import StructureBuilder
 
 from haddock.core.exceptions import ModuleError
 from haddock.modules.topology.topocg.helper import *
 
 warnings.filterwarnings("ignore")
 
+CRYST_LINE = "CRYST1 \n"
```
rvhonorato, line 34: Use `os.linesep` instead of `\n` to avoid problems in the future and to keep in line with the rest of the code

```python
CRYST_LINE = "CRYST1 " + os.linesep
```

src/haddock/modules/topology/topocg/aa2cg.py
rversin: yes, I did this function quickly, I will correct it

src/haddock/modules/topology/topocg/aa2cg.py
rversin: I will add this too :)

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -412,43 +431,47 @@ def determine_ss(structure, skipss, pdbf_path):
...
         else:
             try:
-                dssp = DSSP(model, pdbf_path)
+                p = subprocess.Popen(["dssp", tmp_file_name, "--output-format", "dssp"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+                dssp_raw, errors = p.communicate()
+                dssp_raw = dssp_raw.split('#')[1].split('\n')[:-1]
+                if Path(tmp_file_name).exists(): Path(tmp_file_name).unlink()
             except:  # TODO: think about making this exception more specific
                 # no secondary structure detected for this model
+                if Path(tmp_file_name).exists(): Path(tmp_file_name).unlink()
```
rvhonorato, line 447: You don't need to repeat this one, you can put it outside this try/except

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -412,43 +431,47 @@ def determine_ss(structure, skipss, pdbf_path):
...
-                    code = ss_to_code[ss]
-                    # for atom in residue.get_atoms():
-                    for atom in residue.get_atom():
-                        atom.bfactor = code
+            dssp_ss = "".join(dssp[chain.id])
+            _, martini_types = ss_classification(dssp_ss)  # ancestral function, keep it there
+
+            # transform MARTINI > HADDOCK
+            # and add it to the bfactor col
+            for residue, ss in zip(chain, martini_types):
```
rvhonorato, line 470: Just a note here, this `zip` expects that `chain` and `martini_types` are already in the expected order - maybe its good to add an extra check for it, but if it's too complicated to just add a comment saying that `chain` and `martini_types` order must match

src/haddock/modules/topology/topocg/aa2cg.py
rversin: But if the SS cannot be assigned for whatever reason, the file isn't removed in the folder (because of the continue in the except). So I put it twice. But I can also place it outside of the for loop, before the return.

rversin: I guess I don't really understand where exaclty you want to put it, since the other options I'm trying are not working if DSSP is failing

rvhonorato: line 450

src/haddock/modules/topology/topocg/aa2cg.py
rvhonorato: you can actually create the temporary file inside the function, think it makes it more clean

```python
tmp_file_name = create_file_with_cryst(pdbf_path)
```

rversin: I actually need the name of the temporary file in the next steps, but I will think of a solution

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -401,21 +401,26 @@ def extract_groups(pair_list):
...
         pdb_file: str
-        pdb_file_copy: str
 
     Returns:
+        pdb_file_copy: str
 
     """
+    pdb_file_copy = f"../{output_path}/{pdb_file.split('/')[-1][:-4]}_tmp.pdb"
+    
     with open(pdb_file, "r") as file_in, open(pdb_file_copy, "w") as file_out:
```
rvhonorato, line 419:
```python
with open(pdb_file, "r) as file_in, tempfile.NamedTemporaryFile(delete=False) as file_out:
    content = file_in.read()
    file_out.write(CRYST_LINE + content)

return file_out.name
```

then you don't need two arguments for this function

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -408,47 +431,51 @@ def determine_ss(structure, skipss, pdbf_path):
...
+            finally:
+                if Path(tmp_file_name).exists(): Path(tmp_file_name).unlink()
+
+        dssp = {}
+
+        for line in dssp_raw:
+            if line[11:12] in dssp.keys():
+                dssp[line[11:12]].append(line[16:17])
+            else:
+                dssp[line[11:12]] = [line[16:17]]
```
rvhonorato, line 461: you can remove this whole if/else block if you use the [`setdefault`](https://docs.python.org/3.12/library/stdtypes.html#dict.setdefault) method, and by doing `line[11:12]` what you are actually doing is `line[11]`, so you can also cast it directly to variables:

```python
for line in dssp_raw:
    var_a, var_b = line[11], line[16]
    dssp.setdefault(var_a, []).append(var_b)
```

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -399,7 +402,27 @@ def extract_groups(pair_list):
...
+
+    """
+    with open(pdb_file, "r") as file_in, tempfile.NamedTemporaryFile(mode = "w", delete=False) as file_out:
+        content = file_in.read()
+        file_out.write(CRYST_LINE + content)
+    
+    return(file_out.name)
+
+
+def determine_ss(structure, output_path, skipss, pdbf_path):
```
rvhonorato, line 425: I think `output_path` is not used in this function anymore

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -408,47 +431,51 @@ def determine_ss(structure, skipss, pdbf_path):
         pdbf_path:
 
     Returns:
+        structure:
 
     """
     # calculate SS
     for model in structure:
+        tmp_file_name = create_file_with_cryst(pdbf_path)
```
rvhonorato, line 439: I think this should go in L445 inside the try/except block

src/haddock/modules/topology/topocg/aa2cg.py
rversin: yes I forgot to remove it !

rvhonorato (approved):
still somethings to be improved but good enough for now

rversin:
should I merge the pool request ?

amjjbonvin:
Go for it!

rversin:
Thanks :D

PR #1125, opened 2024-10-31
Minor fix of aa2cg script

rversin:

## Checklist

### DIFF
--- a/src/haddock/modules/topology/topocg/aa2cg.py
+++ b/src/haddock/modules/topology/topocg/aa2cg.py
@@ -508,11 +508,21 @@ def rename_nucbases(structure):
             for chain in model:
                 for r in chain.get_residues():
                     if r.resname in ref_dic.keys():
-                        # rename!
+                        # rename
                         r.resname = ref_dic[r.resname]
 
 
 def martinize(input_pdb, output_path, skipss):
+    """
+    
+    Args:
+        input_pdb: str
+        output_path: str
+        skipss: boolean
+    
+    Returns:
+        cg_pdb_name: str
+    """
 
     if not input_pdb:
         emsg = "No input file detected"
--- a/src/haddock/modules/topology/topocg/aa2cg.py
+++ b/src/haddock/modules/topology/topocg/aa2cg.py
@@ -207,14 +207,14 @@ def center_of_mass(entity, geometric=False):
 
     if geometric:
         return [sum(coord_list) / len(masses) for coord_list in positions]
-    else:
-        w_pos = [[], [], []]
-        for atom_index, atom_mass in enumerate(masses):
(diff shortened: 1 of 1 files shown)


src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -201,18 +202,19 @@ def center_of_mass(entity, geometric=False):
     # If there is a single atom with undefined mass complain loudly.
     if "ukn" in set(masses) and not geometric:
         raise ValueError("Some Atoms don't have an element assigned.\n"
```
VGPReys, line 204: Use `os.linesep` instead of `\n`, for OS interoperability.
```suggestion
        raise ValueError(f"Some Atoms don't have an element assigned.{os.linesep}"
```

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -486,9 +489,11 @@ def rename_nucbases(structure):
     Returns:
 
     """
-    chainresdic = dict([(c.get_id(), [r.get_resname() for r in c.get_residues()]) for m in structure for c in m])
+    chainresdic = dict([(c.get_id(),
+                       [r.get_resname() for r in c.get_residues()]) for m in structure for c in m])
 
-    nucleotide_list = ["CYT", "C", "DC", "THY", "T", "DT", "ADE", "A", "DA", "G", "GUA", "DG", "U", "URI"]
+    nucleotide_list = ["CYT", "C", "DC", "THY", "T", "DT", "ADE",
+                       "A", "DA", "G", "GUA", "DG", "U", "URI"]
```
VGPReys, line 496: Be careful here: "U" is also used for Uranium Ion (see [https://rascar.science.uu.nl/haddock2.4/library](https://rascar.science.uu.nl/haddock2.4/library)).
Even tho, rarely used...

src/haddock/modules/topology/topocg/aa2cg.py
```diff
@@ -612,7 +628,7 @@ def martinize(input_pdb, output_path, skipss):
             n_l = line
         out.write(n_l)
     out.close()
-    subprocess.call(f"mv temp.pdb {cg_pdb_name}", shell=True)
+    if Path("temp.pdb").exists(): Path("temp.pdb").unlink()
```
VGPReys, line 631: What about this version ?
```suggestion
    Path("temp.pdb").unlink(missing_ok=True)
```

src/haddock/modules/topology/topocg/aa2cg.py
rversin: Maybe this is something we should talk with Alex, we should probably make changes for uranium instead of the nucleic acid base in my opinion.

Issue #1119, opened 2024-10-29
Allow the use of srun instead of mpirun for MPI execution


amjjbonvin:
## **Desired feature/enhancement**

The current version of haddock3 requires `mpirun` to be present on the system to run in MPI mode.
As seen on e.g. the Lumi HPC system, `mpirun` can be replaced by `srun` if slurm is installed on the system.

`libmpi.py` should allow for both modes and try to use `srun` in case `mpirun` is not found.

## **Motivation**

This will avoid having to manually edit `libmpi` for installation on some HPC systems.

PR #1127, opened 2024-10-31
Solves issue #1119 related to MPI execution with mpirun or run

amjjbonvin:

## Checklist

It does add `setuptools` as dependency. Present by default in Python3.9 but not 3.12

This pull request modifies the `lipmpi.py` to check for the presence of `mpirun` and if not found for `srun` instead (as used on the Lumi HPC system). 

## Related Issue

Issue #1119

### DIFF
--- a/src/haddock/libs/libmpi.py
+++ b/src/haddock/libs/libmpi.py
@@ -10,6 +10,7 @@ from shutil import which
 from haddock import log
 from haddock.core.exceptions import HaddockTermination
 
+
 class MPIScheduler:
     """Schedules tasks to be executed via MPI."""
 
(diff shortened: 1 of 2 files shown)


mgiulini:
I did some googling and it seems that setuptools and distutils are marked as deprecated and will be removed from future python version: the better alternative here seems to replace 
`from distutils.spawn import find_executable
`
with
`from shutil import which`
and then using 
`if which("mpirun") is not None:
`
rather than
`if find_executable("mpirun") is not None:`

amjjbonvin:
Thanks Marco - will implement and test

amjjbonvin:
Tested on Lumi, using `shutil` and `which` - works fine. All tests passing.

mgiulini (approved):

VGPReys (dismissed):
I modified the termination process, to finalize the haddock3 run properly.

mgiulini (approved):

Issue #1132, opened 2024-11-06
`[caprieval]` with ensemble as reference


VGPReys:
When providing an ensemble as in the parameter `reference_fname`, no computation is performed, as there is a bug in the alignment procedure.

## **Module/Workflow/Library affected**

`[caprieval]`

## **Actual behavior**

part of the `log`
```
Running [caprieval] module
[2024-11-05 16:40:36,091 __init__ INFO] [caprieval] Found previous CNS step: 1_rigidbody
[2024-11-05 16:40:36,192 __init__ INFO] [caprieval] Saved scoring weights to: weights_params.json
[2024-11-05 16:41:17,429 libutil INFO] Selected 30 cores to process 1782 jobs, with 32 maximum available cores.
[2024-11-05 16:41:17,431 libparallel INFO] Using 30 cores
[2024-11-05 16:41:17,468 capri WARNING] Alignment failed between ../data/4_caprieval/reference.pdb and [pdb|2024-11-05 15:50:45] /hddb/roxana/protein-RNA_tutorial/7q4l_redock_bound/3_seletopclusts/3_seletopclusts/cluster_23_model_3.pdb, skipping...
[2024-11-05 16:41:17,473 capri WARNING] Alignment failed between ../data/4_caprieval/reference.pdb and [pdb|2024-11-05 15:50:46] /hddb/roxana/protein-RNA_tutorial/7q4l_redock_bound/3_seletopclusts/3_seletopclusts/cluster_38_model_3.pdb, skipping...
```

## **Steps to reproduce the behavior**

```toml
[caprieval]
reference_fname = "ensemble.pdb"
```

## **Suggestions on how to fix it**

- Parse the `reference_fname`
- Select first model
- Make comparison against all models ?

## **Additional context**

Discovered by our best user: Roxana !

mgiulini:
I would probably go with selecting the first model within the ensemble

VGPReys:
Yes, me too, much simpler.

But there is an interesting situation where you would like to compare your result to a NMR structure represented by an ensemble of conformation, and where there is no trivial ways to assess the quality of your results.

VGPReys:
We could create a new parameter to tackle this purpose.

mgiulini:
yes, but handling that case would make things very complicated code-wise (multiple capri files to be written etc), while a much simpler solution is to run caprieval sequentially over all the elements of the ensemble.
we just have to put a warning when an ensemble is detected, maybe something along these lines:
[...]

PR #1133, opened 2024-11-06
Handle reference issue when multiple conformation provided in `[caprieval]`

VGPReys:
## Checklist

Add machinery to handle multiple conformation in input `reference_fname` in `[caprieval]` module.

When multiple conformations are provided:
- the `capri_ss.tsv` file will contain an additional column `ref_id` allowing to traceback what was the best reference for each complexes
- an additional file `capri_ss_multiref.tsv` will be created: it contains all the performance metrics against all references. Therefore each input complex will have X number of entries (X being the number of reference conformations).

## Related Issue

#1132


src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -69,6 +73,92 @@ def is_nested(models: list[Union[PDBFile, list[PDBFile]]]) -> bool:
...
+        wc_return = sys.stdout.read()  # Read output
+        # Restore original stdout
+        sys.stdout.close()
+        sys.stdout = original_stdout
+        # Parse output
+        for line in wc_return.split("\n"):
+            if "No. models" in line:
+                sline = line.strip().split()
+                nb_models = int(sline[-1])
+                break
```
mgiulini, line 113: extremely complex indeed...

mgiulini (approved):
quite complicated syntax but I guess unavoidable..let's go for it

src/haddock/modules/analysis/caprieval/__init__.py
VGPReys: yes, the pdb_tools `pdb_wc` is not returning anything... :'(

VGPReys:
About to create a new PR to be merged to this one, expending the approach to evaluate on multiple references instead of only selecting the first.

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -69,6 +73,92 @@ def is_nested(models: list[Union[PDBFile, list[PDBFile]]]) -> bool:
...
+        with open(reference, "r") as fh:
+            pdb_wc(fh, "m")
+        # Get output
+        sys.stdout.seek(0)  # Jump to the start
+        wc_return = sys.stdout.read()  # Read output
+        # Restore original stdout
+        sys.stdout.close()
+        sys.stdout = original_stdout
+        # Parse output
+        for line in wc_return.split("\n"):
```
VGPReys, line 109:
```suggestion
        # using here `\n` (and not os.linesep) as it is the output for pdb_wc
        for line in wc_return.split("\n"):
```

mgiulini:
> About to create a new PR to be merged to this one, expending the approach to evaluate on multiple references instead of only selecting the first.

why? didn't we agree that evaluating on multiple references would be confusing?

VGPReys:
> > About to create a new PR to be merged to this one, expending the approach to evaluate on multiple references instead of only selecting the first.
> 
> why? didn't we agree that evaluating on multiple references would be confusing?

Yes, but in the end I think I found an ok way to handle this without a very minimum number of modifications.
It also expends the possibilities of the module, which I feel is nice.
Because splitting the references into X files and running X `[caprieval]` and having to parse X files to gather the data seems (to me) difficult for unprepared users.
Of course, they can still do it if they want.

mgiulini:
I tried this on an independent example:
I see that capri calculations are performed, but I cannot see where the CAPRI results for the various "non-best" models are saved. If we calculate these numbers then we have to provide them to the users somehow.

VGPReys:
> I tried this on an independent example: I see that capri calculations are performed, but I cannot see where the CAPRI results for the various "non-best" models are saved. If we calculate these numbers then we have to provide them to the users somehow.

To not disturb the user, the outputs from `caprieval`  are the same.
The only difference (introduced by https://github.com/haddocking/haddock3/pull/1136) is the fact that reported values only corresponds to the best reference, which may be different for each model.

Do you think that additionally we should create a `ref_X_capri_ss.tsv` for each `X` reference conformation ?

mgiulini:
> > I tried this on an independent example: I see that capri calculations are performed, but I cannot see where the CAPRI results for the various "non-best" models are saved. If we calculate these numbers then we have to provide them to the users somehow.
> 
> To not disturb the user, the outputs from `caprieval` are the same. The only difference (introduced by #1136) is the fact that reported values only corresponds to the best reference, which may be different for each model.
> 
> Do you think that additionally we should create a `ref_X_capri_ss.tsv` for each `X` reference conformation ?

If this is the desired logic, I think so. Otherwise it would be impossible to trace back which numbers correspond to which reference, correct?

VGPReys:
> > > I tried this on an independent example: I see that capri calculations are performed, but I cannot see where the CAPRI results for the various "non-best" models are saved. If we calculate these numbers then we have to provide them to the users somehow.
> > 
> > 
> > To not disturb the user, the outputs from `caprieval` are the same. The only difference (introduced by #1136) is the fact that reported values only corresponds to the best reference, which may be different for each model.
> > Do you think that additionally we should create a `ref_X_capri_ss.tsv` for each `X` reference conformation ?
> 
> If this is the desired logic, I think so. Otherwise it would be impossible to trace back which numbers correspond to which reference, correct?

Indeed, for now it is not possible to trace back which numbers corresponds to which reference.
But this is intended to behave that way, as the question solved is: what is the quality of my docking model against any of the reference conformations (mostly used for NMR ensemble).

To solve the traceback issue, I see multiple solutions:
- add the machinery to generate multiple capri_ss / capri_clt for each reference conformation.
- add a `ref_id` column in the capri_ss, referencing to the index of the best reference in the conformational ensemble for each input model
- do both

amjjbonvin:
I would vote for for adding a ref_id column.  

> 
> add a ref_id column in the capri_ss, referencing to the index of the best reference in the conformational ensemble for each input model

amjjbonvin:
Make sure to check this is not breaking any of the analysis part, nor the haddock3-webapp display of results.

VGPReys:
> Make sure to check this is not breaking any of the analysis part, nor the haddock3-webapp display of results.

Very good point !

Should I also update the analysis scripts to display the `ref_id` data in the table ?

amjjbonvin:
Not needed - this only makes sense in a benchmarking scenario - in real applications you won’t know the reference…
We could for simplicity even think of writing a separate file that contains ref_id (only in case where an ensemble is provided)


> Should I also update the analysis scripts to display the ref_id data in the table ?
>

VGPReys:
Analysis plots are rendering without issues.

Also, now, when multiple conformations are provided:
- the `capri_ss.tsv` file will contain an additional column `ref_id` allowing to traceback what was the best reference for each complexes
- an additional file `capri_ss_multiref.tsv` will be created: it contains all the performance metrics against all references, and the `ref_id` column allowing for traceback. Therefore each input complex will have X number of entries (X being the number of conformations in the reference file).

mgiulini (approved):
working as expected!

PR #1136, opened 2024-11-08
`[caprieval]` comparisons against multiple references

VGPReys:
## Checklist

This PR in made to handle comparisons with multiple references.

This is performed by turning on the parameter `multiple_references = true` (default false).
In this scenario, reported values in `capri_ss.tsv` are the ones obtained against the best reference.

If `multiple_references = false` (default), the first reference will be used as reference. 

I had to add the `__eq__` and `__lt__` methods to the CAPRI class, so comparisons could be performed by functions/methods such as `sorted([CAPRI, CAPRI])` or `[CAPRI, CAPRI].sort()`.

## Related Issue

#1133


src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -177,30 +200,33 @@ def _run(self) -> None:
...
-                    "of `model_to_be_evaluated`"
+            # Loop over references
+            for reference in references:
+                # `models_to_be_evaluated` cannot be a list,
+                # `CAPRI` class is expecting a single model
+                if isinstance(model_to_be_evaluated, list):
+                    raise ValueError(
+                        "CAPRI module cannot handle a list "
+                        "of `model_to_be_evaluated`"
+                    )
```
VGPReys, line 220:
```suggestion
            # `models_to_be_evaluated` cannot be a list,
            # `CAPRI` class is expecting a single model
            if isinstance(model_to_be_evaluated, list):
                raise ValueError(
                    "CAPRI module cannot handle a list "
                    "of `model_to_be_evaluated`"
                    )
            # Loop over references
            for reference in references:
```

examples/analysis/topoaa-caprieval_ensemble-test.cfg
```diff
@@ -0,0 +1,28 @@
...
+molecules = "data/ensemble_4G6M.pdb"
+
+# ====================================================================
+# Parameters for each stage are defined below, prefer full paths
+# ====================================================================
+[topoaa]
+
+[caprieval]
+# Comparison agains first reference
+reference_fname = "data/ensemble_4G6M.pdb"
```
mgiulini, line 21: I find this example confusing for users: why using the same input ensemble as a reference in caprieval? the fact that we support this does not mean we should encourage this..this could become an integration test!

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -8,6 +8,18 @@ reference_fname:
...
+  default: false
+  type: boolean
+  title: Allow the use of multiple reference conformations.
+  short: Performs caprieval computation on multiple reference conformations.
+  long: Performs caprieval computation on multiple reference conformations.
+    Final values will contain the best performing metrics with respect to
+    all reference conformations. Usefull when evaluating docking poses against
+    NMR structures, for example.
+  group: analysis
+  explevel: expert
```
mgiulini, line 21: why this parameter? isn't it easier to evaluate the quality of whatever is provided as `reference_fname` in caprieval? maybe limiting the number of reference structures to 10 or 20

examples/analysis/topoaa-caprieval_ensemble-test.cfg
VGPReys: Yes, indeed !
At least we see that we obtain 1 in dockq and 0 in rmsd.
I will find an other example !

src/haddock/modules/analysis/caprieval/defaults.yaml
VGPReys: To make sure the user is willing to compare against multiple references.
Could be removed and always applied.

mgiulini: I would remove it..I think the logic gets too complicated otherwise

examples/analysis/topoaa-caprieval_ensemble-test.cfg
mgiulini: you can also remove the example and we only have this as integration test..depends on how much we want to advertise this feature

VGPReys: OK

src/haddock/modules/analysis/caprieval/defaults.yaml
VGPReys: OK

mgiulini (approved):

PR #1172, opened 2024-12-09
modify way of getting the ncores

### LINKED ISSUE #1168: number of core determination :: potential overcommit
Hello,

haddock3 use `os.cpu_coun()` in `libutil.py` in `parse_ncores` function. this may cause problems in some environment, eg: taskset, batch system like slurm, etc...

python `os.cpu_count()` returns the number of physical cores available on the computer it runs on. BUT this is not the number of **AVAILABLE** cores for the running process.
eg in slurm allocation, you can run on a node with 96 cores, but only have access to **n** cores regarding the allocation.

I would suggest to use `len(os.sched_getaffinity(0))` instead of `os.cpu_count()`

a demo worth all the words.

```
# slurm allocation with 4 cores
maestro-submit:~ > srun -c 4 python3  -c "import os; print(os.cpu_count(),  len(os.sched_getaffinity(0)))"
96 4
```

means you may have for example  a job submitted on a 96 cores HPC nodes, in a single core slurm allocation: 
you will have more  threads fighting for the usage of this core.

regards

Eric

VGPReys:

Modifying the way number of cores are detected, from `os.cpu_count()` to `os.sched_getaffinity(0)`.

## Related Issue

Closes #1168 

## Additional Info

Based on user suggestion.


src/haddock/libs/libutil.py
```diff
@@ -128,6 +128,31 @@ def remove_dict_keys(d: ParamMap, keys: Container[str]) -> ParamDict:
...
+
+    Note: pid 0 == current process
+
+    FIXME: from python3.13, better to use os.process_cpu_count()
+
+    Returns
+    -------
+    process_ncores : int
+        Number of cores allocated to the process pid.
+    """
```
mgiulini, line 144: shall we add a test for this new function

VGPReys: I do not know how, because we need to know what setup is running the test

mgiulini (approved):

PR #1183, opened 2024-12-13
back to basics for macos users when detecting number of cores

### LINKED ISSUE #1182: The last pull request #1172 about core detection broke the execution on OSX
The latest pull request (#1172) broke the execution on OSX because 'sched_getaffinity' is not defined.

```
[2024-12-12 11:38:37,889 libutil ERROR] module 'os' has no attribute 'sched_getaffinity'
  File "/Users/abonvin/haddock_git/haddock3/src/haddock/libs/libutil.py", line 146, in cpu_count
    process_ncores = int(os.process_cpu_count())
                         ^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'os' has no attribute 'process_cpu_count'
```

libutil.py

## **Expected behavior**

A check (try) if `sched_getaffinity` is available on the system, and if not revert to the old method of getting the number of cores

Execute one of the test example under OSX

VGPReys:

Try to solve the problem related to core detection on MacOS.
The implemented solution is to go back to the original function `os.cpu_count()`.
While this may not be fully accurate, it allows to at least return the number of cores available on MacOS machines.
On Linux systems, the `os.sched_getaffinity(0)` function is used, returning the number of cores allocated to the process, which is a better solution for HPC and SLURM executions.

## Related Issue

Closes #1182

### DIFF
--- a/src/haddock/libs/libutil.py
+++ b/src/haddock/libs/libutil.py
@@ -142,14 +142,32 @@ def cpu_count() -> int:
     process_ncores : int
         Number of cores allocated to the process pid.
     """
+    def _cpu_count() -> int:
+        """Detect number of cores available.
+        
+        Returns
+        -------
+        ncores : int
+            Maximum number of cores that can be used.
+        """
+        try:
+            # Try to obtain the number of cpu allocated to the process
+            _process_ncores = os.sched_getaffinity(0)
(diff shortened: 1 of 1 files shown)


amjjbonvin (approved):
Works perfectly

PR #1193, opened 2025-01-24
Cg haddock cg2aa

rversin:

## Checklist

The aatocg module was added. In order to have the aatocg module fully functional, modifications had to be made to the libs library. The topocg and rigisbody modules had to be adapted as well.


src/haddock/modules/refinement/cgtoaa/defaults.yaml
VGPReys, line 1: Are all those parameters required to the `cgtoaa` module ?
I think most of them could be removed (e.g: ligand_param_fname, maxmodels, etc...)

src/haddock/modules/refinement/cgtoaa/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+                    model,
+                    self.path,
+                    self.recipe_str,
+                    self.params,
+                    "cgtoaa",
+                    native_segid=True,
+                    debug=self.params["debug"],
+                    seed=model.seed if isinstance(model, PDBFile) else None,
+                    cgtoaa=True
+                )
```
VGPReys, line 88: I think this seed value must be modified for each sampled models `sampling_factor`
```suggestion
            for s_ind in range(self.params["sampling_factor"]):
                cgtoaa_input = prepare_cns_input(
                    idx,
                    model,
                    self.path,
                    self.recipe_str,
                    self.params,
                    "cgtoaa",
                    native_segid=True,
                    debug=self.params["debug"],
                    seed=(model.seed + s_ind) if isinstance(model, PDBFile) else None,
                    cgtoaa=True
                )
```

src/haddock/libs/libontology.py
```diff
@@ -69,6 +69,8 @@ def __init__(
         self,
         file_name: Union[Path, str],
         topology: Optional[Any] = None,
+        aa_topology: Optional[Any] = None,
+        aatocg_tbl:Optional[Any] = None,
```
VGPReys, line 73: Should it be called more `cgtoaa_tbl` instead of `aatocg_tbl` ?
Because the restraints are used to convert back from CG to all atoms.

src/haddock/libs/libcns.py
```diff
@@ -322,10 +326,46 @@ def prepare_cns_input(
         psf_fname = pdb.topology.rel_path
         psf_list.append(psf_fname)
 
+    aa_psf_list: list[Path] = []
+    aatocg_tbl_list: list[Path] = []
+    if cgtoaa==True:
+        if isinstance(input_element.aa_topology, (list)):
+            for psf in input_element.aa_topology:
+                if psf is None:
+                    raise ValueError(f"AA Topology not found {input_element.rel_path}.")
```
VGPReys, line 335: Here the user will get a none-trivial error to be understood.
In my opinion, I would say that we should let the user know that trying to convert back from Coarse grained to All Atoms require that an initial topology has been created (using `[topoaa]` + `[aatocg]` or [`topocg`]).

Also, I think that abbreviations should be avoided, so the user may understand what's going on:
- AA -> All Atoms 
- CG -> Coarse Grained

src/haddock/modules/refinement/cgtoaa/defaults.yaml
amjjbonvin: Good point - I did already remove quite a bit - will check the remaining ones.

src/haddock/libs/libontology.py
amjjbonvin: More logical indeed - should also be adapted in the topocg module

src/haddock/modules/refinement/cgtoaa/__init__.py
rversin: This was copied from the emref module, should it also be changed in the emref then ?

src/haddock/libs/libontology.py
rversin: I put aatocg because of the name of the file produced by aatocg, but i can also adapt the name of this file

src/haddock/modules/refinement/cgtoaa/__init__.py
amjjbonvin: In minimisation there is no seed used. For the cgtoaa module we use molecular dynamics and thus a starting seed. So only needed here.  You should also find it in the mdref module for example

src/haddock/libs/libontology.py
amjjbonvin: Might be more logical to adapt as we are using it to convert from CG to AA

rversin: ok will do !

rversin:
I think I will be merging this today, if it's okay for everyone :)

amjjbonvin:
Yes please

PR #1207, opened 2025-02-19
added nanobody example data

### LINKED ISSUE #1204: add nanobody-specific examples
add examples about nanobody-antigen complex modelling

## **Motivation**
The workflow and the restraints might differ slightly from the antibody ones.

mgiulini:
## Checklist

This PR adds a folder to the examples directory with an example nanobody-antigen modelling case.
The considered example comes from PDB file 7tgf, with the nanobody ensemble comes from clustering alphafold and immunebuilder models and the antigen coming from PDB file 7y0v

Several scenarios are present in terms of restraint files:

- true paratope vs true epitope
- exposed CDR vs loosely defined epitope
- exposed CDR vs mutagenesis-mapped epitope
- mixed nanobody restraints: in a single ambig.tbl.tgz are present two ambig.tbl files. In the first, the standard exposed CDR are defined as active, but in the second only the H3 loop is fully active and the other two loops and some regions in the framwork are ambiguously defined as active. This is particularly useful in this case, as H1 and H2 loops are not very relevant and three residues in the framework are in contact with the antigen.

The standard HADDOCK protocol is shown.

## Related Issue
Closes #1204


docs/examples.md
```diff
@@ -42,6 +43,18 @@ Three different protocols/workflows are illustrated:
 
 The `caprieval` module is called at various stages during the workflow to assess the quality of the models with respect to the known reference structure.
 
+## docking-nanobody-antigen
```
amjjbonvin, line 46: A rather lengthy description compared to the other tutorials. May-be shorten it and add instead a README file into the directory with all the details.

mgiulini: agreed and done!

amjjbonvin (approved):

PR #1223, opened 2025-03-13
update `caprieval` to handle coarse-grained structures

rversin:

## Checklist

This pull request corresponds to the adaptation of the caprieval module for coarse-grained structures.


VGPReys (changes requested):
Is it really needed to add the `cgffversion` parameter ? Can't we detect that we are already working with coarse-grain ?

There is an issue with the variable `cg` set in the __init__.py:
- It is set to be an integer, but the CAPRI class and other functions are defined receiving `bool` types.
- Also `cg` integer is used to point a specific position in lists, and therefore cannot be of type `bool`.

Instead of lists, I would rather use dicts (in `libalign.py`):
```python3
PROT_ATOMS = [
    PROT_ATOMS_AA,
    PROT_ATOMS_MARTINI2,
    PROT_ATOMS_MARTINI2 # The backbone particle is identical between MARTINI 2 and 3
]
```
to 
```python3
PROT_ATOMS = {
    "aa": PROT_ATOMS_AA,
    "martini2": PROT_ATOMS_MARTINI2,
    "martini3": PROT_ATOMS_MARTINI2 # The backbone particle is identical between MARTINI 2 and 3
}
```
(this also applies for DNA and RNA lists)

and make `cg` of type `str`, taking possible values `aa`, `martini2` (and in the future `martini3`)
I think it will be easier to understand what is happening.

src/haddock/modules/analysis/caprieval/defaults.yaml
```diff
@@ -197,3 +197,18 @@ allatoms:
...
+  minchars: 0
+  maxchars: 100
+  choices:
+    - None
+    - martini2
+  title: MARTINI Force-field version
+  short: Force-field version for the coarse-grain mapping and parameters.
+  long: Force-field version for the coarse-grain mapping and parameters. To be 
+        used when working with coarsed-grained structures.
+  group: module
```
VGPReys, line 213:
```suggestion
  group: force-field
```

rversin:
For the cgffversion parameter, I chose this option instead of opening and closing the pdb files in the caprieval module directly, which was the only other solution I could think of to check if it was CG or not. This pdb opening or closing felt tedious, especially because it is tricky to differentiate between martini2 and martini3 structures based on the pdb.

rversin:
For the variable cg , I indeed made a mistake, at first it was a bool then I changed my mind but didn't change the definition of the variable  
I will correct it :)

VGPReys:
> For the cgffversion parameter, I chose this option instead of opening and closing the pdb files in the caprieval module directly, which was the only other solution I could think of to check if it was CG or not. This pdb opening or closing felt tedious, especially because it is tricky to differentiate between martini2 and martini3 structures based on the pdb.

Don't you have access to the type of topology based on the content of the `io.json` ?

src/haddock/libs/libalign.py
```diff
@@ -590,13 +710,13 @@ def get_atoms(pdb: PDBPath, full: bool = False) -> AtomsDict:
...
+    atom_dic.update((r, PROT_ATOMS[cg]) for r in PROT_RES)
+    atom_dic.update((r, DNA_ATOMS[cg]) for r in DNA_RES)
+    atom_dic.update((r, RNA_ATOMS[cg]) for r in RNA_RES)
     if full:
-        atom_dic.update(PROT_SIDE_CHAINS_DICT)
-        atom_dic.update(DNA_FULL_DICT)
-        atom_dic.update(RNA_FULL_DICT)
+        atom_dic.update(PROT_SIDE_CHAINS_DICT[cg])
+        atom_dic.update(DNA_FULL_DICT[cg])
+        atom_dic.update(RNA_FULL_DICT[cg])
```
rvhonorato, line 720: be careful here, this change breaks a core function in the analysis workflows - this is a confusing implementation which adds unecessary complexity to the types. if what you want to do is to just update the `atom_dic` based on a condition, do not change the type structure of the function but instead add simple if/else block.

rversin: ok will do ! thanks

rvhonorato (changes requested):
This PR introduces changes to an important analysis module but there are no tests covering the new code - do not merge this without tests.

mgiulini:
Hi @rversin : I am not sure whether you're still working on this, but I would suggest to simply add a couple of tests in [test_module_caprieval.py](https://github.com/haddocking/haddock3/blob/fc7d224d0c4f05fd1cd424a084db11927193df7f/tests/test_module_caprieval.py) rather than creating a brand new test file. For caprieval, the CG is just a property of the structures, which should be tested in the tests proper to the module. We can save some lines and complexity without having tons of separate test files :)

rversin:
Hello everyone,
I did the last checks, and hopefully I didn't miss anything :)
If you all agree, we should be able to merge the branch into CG-HADDOCK.

examples/docking-protein-protein-CG/docking-protein-protein-CG-full.cfg
```diff
@@ -0,0 +1,73 @@
...
+#queue_limit = 100
+debug = True
+
+# molecules to be docked
+molecules =  [
+    "data/e2a_1F3G.pdb",
+    "data/hpr_ensemble_B.pdb"
+    ]
+
+postprocess = false
```
VGPReys, line 22: I would remove the part including the `batch` mode and the commented lines.

```suggestion

# molecules to be docked
molecules =  [
    "data/e2a_1F3G.pdb",
    "data/hpr_ensemble_B.pdb"
    ]
```

rversin: oh right thanks !

src/haddock/modules/analysis/caprieval/capri.py
```diff
@@ -112,6 +112,7 @@ def load_contacts(
     cutoff: float = 5.0,
     numbering_dic: Optional[dict[str, dict[int, int]]] = None,
     model2ref_chain_dict: Optional[dict[str, str]] = None,
+    cg: str = "aa"
```
VGPReys, line 115: missing docstring for `cg`. (And also other arguments, but it's not coming from your implementation.)
Also, I would call it `ff` or `force_field` as: `cg == coarse grained`, but `cg="aa"` == `coarse grained = all atoms` which is weird in my opinion.

rversin: yes, it comes from the fact that at first it was a true/false boolean

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -182,7 +183,24 @@ def _run(self) -> None:
...
 
         # Get reference file
-        references = self.get_reference(models)
+        try:
+            cgffversion = models[0].topology[0].rel_path.as_posix().split("_")[-1].split(".")[0]
+        except TypeError:
+            try:
+                cgffversion = models[0].topology.rel_path.as_posix().split("_")[-1].split(".")[0]
+            except AttributeError:
+                cgffversion = False
```
VGPReys, line 192: Can't this line simply take the value `cgffversion = 'aa' ?
Then you send `cgffversion` to the `CAPRI` class directly (`cg=cgffversion`)

src/haddock/modules/analysis/caprieval/capri.py
```diff
@@ -179,6 +180,7 @@ def __init__(
         reference: PDBPath,
         params: ParamMap,
         ref_id: int = 1,
+        cg: str = "aa"
```
VGPReys, line 183: again, missing some docstring for this new parameter

src/haddock/modules/analysis/caprieval/capri.py
```diff
@@ -659,6 +663,7 @@ def _load_atoms(
         model: PDBPath,
         reference: PDBPath,
         full: bool = False,
+        cg: str = "aa",
```
VGPReys, line 666: missing docstring

src/haddock/modules/analysis/caprieval/capri.py
```diff
@@ -688,6 +693,7 @@ def _load_atoms(
     def identify_interface(
         pdb_f: PDBPath,
         cutoff: float = 5.0,
+        cg: str = "aa",
```
VGPReys, line 696: missing docstring

tests/test_module_caprieval.py
```diff
@@ -292,8 +352,64 @@ def test_protprot_allatoms(protprot_allatm_caprimodule):
...
+    # fnat
+    protprot_caprimodule_cg.calc_fnat(cutoff=7.0)
+    print(protprot_caprimodule_cg.fnat)
+    assert np.isclose(protprot_caprimodule_cg.fnat, 0.083, atol=0.01)
+
+
+def test_protprot_global_rmsd_cg(protprot_caprimodule_cg):
+    """Test protein-protein global rmsd calculation."""
+    protprot_caprimodule_cg.calc_global_rmsd()
+    print(protprot_caprimodule_cg.rmsd)
```
VGPReys, line 380: removed print
```suggestion
```

tests/test_module_caprieval.py
```diff
@@ -292,8 +352,64 @@ def test_protprot_allatoms(protprot_allatm_caprimodule):
...
 
+def test_protprot_allatoms_cg(protprot_allatm_caprimodule_cg):
+    """Test protein-protein all atoms calculations."""
+    # Compute values with all atoms calculations
+    protprot_allatm_caprimodule_cg.calc_fnat()
+    protprot_allatm_caprimodule_cg.calc_lrmsd()
+    protprot_allatm_caprimodule_cg.calc_irmsd()
+    protprot_allatm_caprimodule_cg.calc_ilrmsd()
+    # process checks
+    print(protprot_allatm_caprimodule_cg.lrmsd)
```
VGPReys, line 410: removed print
```suggestion
```

tests/test_module_caprieval.py
```diff
@@ -458,6 +619,21 @@ def test_identify_protprotinterface(protprot_caprimodule, protprot_input_list):
...
         assert sorted(observed_interface[ch]) == sorted(expected_interface[ch])
 
+def test_identify_protprotinterface_cg(protprot_caprimodule_cg, protprot_input_list_cg):
+    """Test the interface identification."""
+    protprot_complex = protprot_input_list_cg[0]
+
+    observed_interface = protprot_caprimodule_cg.identify_interface(
+        protprot_complex, cutoff=7.0, cg='martini2'
+    )
+    print(observed_interface)
```
VGPReys, line 629: removed print
```suggestion
```

tests/test_module_caprieval.py
```diff
@@ -458,6 +619,21 @@ def test_identify_protprotinterface(protprot_caprimodule, protprot_input_list):
     for ch in expected_interface.keys():
         assert sorted(observed_interface[ch]) == sorted(expected_interface[ch])
 
+def test_identify_protprotinterface_cg(protprot_caprimodule_cg, protprot_input_list_cg):
+    """Test the interface identification."""
+    protprot_complex = protprot_input_list_cg[0]
+
+    observed_interface = protprot_caprimodule_cg.identify_interface(
+        protprot_complex, cutoff=7.0, cg='martini2'
```
VGPReys, line 627:
```suggestion
        protprot_complex, cutoff=7.0, cg="martini2",
```

tests/test_module_topocg.py
```diff
@@ -0,0 +1,107 @@
...
+    return Path(golden_data, "protein.pdb")
+
+
+@pytest.fixture(name="topocg")
+def fixture_topocg(monkeypatch):
+    """topocg module fixture"""
+    #tempdir = tempfile.mkdtemp()  # does NOT auto-delete
+    #monkeypatch.chdir(tempdir)
+    #print(f"Temp dir: {tempdir}")  # helpful for debugging
+    #yield Topocg(order=1, path=Path("."), initial_params=topocg_params)
```
VGPReys, line 58: removed commented section
```suggestion
```

src/haddock/libs/libalign.py
```diff
@@ -572,7 +691,9 @@ def load_coords(
     return coord_dic, chain_ranges
 
 
-def get_atoms(pdb: PDBPath, full: bool = False) -> AtomsDict:
+def get_atoms(pdb: PDBPath, 
+              full: bool = False,
+              cg: str = "aa") -> AtomsDict:
```
VGPReys, line 696: missing docstring for this new argument

src/haddock/modules/analysis/caprieval/__init__.py
rversin: This doesn't work, because the value of cgffversion is not automatically false when it is an all-atom structure.
When it is an all-atom structure, the value is actually "haddock", so I can't put cg=cgffversion, i should put cg='aa'.
I will change the name of cg though

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -182,7 +183,24 @@ def _run(self) -> None:
         dump_weights(self.order)
 
         # Get reference file
-        references = self.get_reference(models)
+        try:
+            cgffversion = models[0].topology[0].rel_path.as_posix().split("_")[-1].split(".")[0]
```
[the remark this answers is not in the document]
rversin: it does not work, I have the following error :
`AttributeError: 'TopologyFile' object has no attribute 'stem'`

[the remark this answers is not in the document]
VGPReys: ow, its not a `Path` object but a `TopologyFile` object, by bad !

tests/test_module_caprieval.py
```diff
@@ -441,6 +557,51 @@ def test_protdna_allatoms(protdna_caprimodule):
...
+    assert np.isclose(protdna_caprimodule_cg.irmsd, 2.09, atol=0.01)
+    protdna_caprimodule_cg.calc_lrmsd()
+    assert np.isclose(protdna_caprimodule_cg.lrmsd, 6.32, atol=0.01)
+    protdna_caprimodule_cg.calc_ilrmsd()
+    assert np.isclose(protdna_caprimodule_cg.ilrmsd, 6.05, atol=0.01)
+    protdna_caprimodule_cg.calc_fnat(cutoff=7.0)
+    assert np.isclose(protdna_caprimodule_cg.fnat, 0.58, atol=0.01)
+
+
+# TODO: Move to integration tests
```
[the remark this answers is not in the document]
rversin: this is not from me, it was already here

rversin:
Hi, 
If it's fine with everyone, I would like to merge this now :)

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -182,7 +183,24 @@ def _run(self) -> None:
...
+        except TypeError:
+            try:
+                cgffversion = models[0].topology.rel_path.as_posix().split("_")[-1].split(".")[0]
+            except AttributeError:
+                cgffversion = "aa"
+
+        if cgffversion == "martini2":
+            references = []
```
VGPReys, line 195: You can remove this line

```suggestion
```

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -182,7 +183,24 @@ def _run(self) -> None:
...
+            references = [
+                Path(martinize(ref_aa, self.path, False))
+                for ref_aa in self.get_reference(models)
+                ]
+            ff = cgffversion
+        else:
+            references = self.get_reference(models)
+            ff = "aa"
```
VGPReys, line 203: Why using both `cgffversion` and `ff` if `ff = cgffversion` ?

```suggestion
        try:
            ff = models[0].topology[0].rel_path.as_posix().split("_")[-1].split(".")[0]
        except TypeError:
            try:
                ff = models[0].topology.rel_path.as_posix().split("_")[-1].split(".")[0]
            except AttributeError:
                ff = "aa"

        if ff == "martini2":
            references = [
                Path(martinize(ref_aa, self.path, False))
                for ref_aa in self.get_reference(models)
                ]
        else:
            references = self.get_reference(models)
```

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -182,7 +183,24 @@ def _run(self) -> None:
...
 
         # Get reference file
-        references = self.get_reference(models)
+        try:
+            cgffversion = models[0].topology[0].rel_path.as_posix().split("_")[-1].split(".")[0]
+        except TypeError:
+            try:
+                cgffversion = models[0].topology.rel_path.as_posix().split("_")[-1].split(".")[0]
+            except AttributeError:
+                cgffversion = "aa"
```
VGPReys, line 192: I would make a function for the section that searches for the force-field.

```python3
def get_ff_version(model: PDBFile) -> str:
    try:
        cgffversion = model.topology[0].rel_path.as_posix().split("_")[-1].split(".")[0]
    except TypeError:
        try:
            cgffversion = model.topology.rel_path.as_posix().split("_")[-1].split(".")[0]
        except AttributeError:
            cgffversion = "aa"
    return cgffversion
ff = get_ff_version(models[0])
```

src/haddock/modules/analysis/caprieval/capri.py
```diff
@@ -671,14 +680,16 @@ def _load_atoms(
             PDB file of the model to have its atoms identified
         full : bool
             If False, only backbone atoms will be retrieved, otherwise all atoms
+        ff : string
+            Force-field information : all-atom, martini2 or martini3
```
VGPReys, line 684: Because the desired value for all-atom is "aa", I would put "aa" in the docstring.
Also, martini3 is not implemented yet. But I get the point of putting it now already as it might be scheduled for the near future.

This applied for all docstrings describing theis parameter
```suggestion
            Force-field information : aa (all-atom), martini2 or martini3
```

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -72,6 +73,17 @@ def is_nested(models: list[Union[PDBFile, list[PDBFile]]]) -> bool:
             if isinstance(model, list):
                 return True
         return False
+    
+    @staticmethod
+    def find_ff(models: PDBFile) -> str:
```
VGPReys, line 78:
```suggestion
    def find_ff(models: list[PDBFile]) -> str:
```

rversin: right ! I did it too fast

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -182,7 +194,16 @@ def _run(self) -> None:
...
+        ff = self.find_ff(models)
+
+        if ff == "martini2":
+            references = [
+                Path(martinize(ref_aa, self.path, False))
+                for ref_aa in self.get_reference(models)
+                ]
+        else:
+            references = self.get_reference(models)
+            ff = "aa"
```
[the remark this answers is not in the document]
rversin: Actually no ! when we work with all atom structures, the output of find_ff is not "aa" but "haddock". 
I could modify the function so that it outputs "aa" when we don't have "martini" in the name, as we don't want to change the original naming system (I assume).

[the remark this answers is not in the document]
VGPReys: I think it makes sense to indeed return "aa" if "martini" is not in the name
because returning "haddock" from the `find_ff()` function is quite weird as haddock is not a force-field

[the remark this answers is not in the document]
rversin: yes ! changing it now :)

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -72,6 +73,21 @@ def is_nested(models: list[Union[PDBFile, list[PDBFile]]]) -> bool:
             if isinstance(model, list):
                 return True
         return False
+    
+    @staticmethod
+    def find_ff(models: list[PDBFile]) -> str:
```
VGPReys, line 78: Sorry that I forgot to mention it before, but this method lacks docstring

rversin: right !

src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -72,6 +73,23 @@ def is_nested(models: list[Union[PDBFile, list[PDBFile]]]) -> bool:
...
+        """Finds the force-field information (all-atom or martini) from the topology 
+        associated to the first model (the force-fields will be identical between models)"""
```
VGPReys, line 80:
```suggestion
        """Finds the force-field information (all-atom or martini) from the topology 
        associated to the first model (the force-fields will be identical between models).
        
        Parameters
        ------------
        models : list[PDBFile]
            List of models where to find to topology
        
        Return
        -------
        ff : str
            The force-field used in those models.
        """
```

VGPReys (approved):

src/haddock/modules/analysis/caprieval/__init__.py
rversin: thanks victor !

PR #1226, opened 2025-03-14
Adding new filtering module `[filter]`

### LINKED ISSUE #1225: Filtering module

It would be nice to have a new module allowing to filter complexes based on the score.

## **Motivation**

`[seletop]` might not be sufficient, especially if we already know what criteria must be fulfilled.
I could allow the filtering of:
- structures with clashes
- satisfying a certain type of restraints
- with a threshold on BSA
- with a threshold on buried/exposed residues computed by the `[sasascore]` module

VGPReys:
## Checklist

- [X] User manual was updated (https://github.com/haddocking/haddock3-user-manual/pull/4)

This PR adds a new module `[filter]` allowing to filter input models based on selected score.
A single parameter must be tuned: `threshold`, allowing to set the threshold value above which models will be filtered OUT. The tested condition is `if score <= threshold: Keep Model`.

## Related Issue

Closes #1225


src/haddock/modules/analysis/filter/defaults.yaml
```diff
@@ -0,0 +1,25 @@
...
+  default: "score"
+  type: string
+  minchars: 1
+  maxchars: 20
+  choices:
+    - score
+    - sasascore
+  title: Name of the scoring function to be used.
+  short: Name of the scoring function to be used to filter complexes.
+  long: Name of the scoring function to be used to filter complexes. The score must be present and computed before hand.
```
AnnaKravchenko, line 11: i’d change 'to filter complexes’ to 'to filter models’, in both short and long descriptions

src/haddock/modules/analysis/filter/defaults.yaml
```diff
@@ -0,0 +1,25 @@
...
+  group: analysis
+  explevel: easy
+
+cutoff:
+  default: 0.00
+  type: float
+  min: -99999.99
+  max: 99999.99
+  precision: 2
+  title: Cutoff value to filter complexes.
```
AnnaKravchenko, line 21: Same here, ‘models’ rather than ‘complexes’ in title, short and long

src/haddock/modules/analysis/filter/defaults.yaml
```diff
@@ -0,0 +1,25 @@
+by:
+  default: "score"
+  type: string
+  minchars: 1
+  maxchars: 20
+  choices:
+    - score
```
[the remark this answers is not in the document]
amjjbonvin: Good point indeed!

src/haddock/modules/analysis/filter/defaults.yaml
amjjbonvin: Another good point - we can also use haddock for single structures. So model makes more sense

src/haddock/modules/analysis/filter/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+                " that produced an iterable"
+                )
+        models: list[PDBFile] = [
+            p
+            for p in self.previous_io.output
+            if p.file_type == Format.PDB
+            ]
+        
+        # Get the filter by parameter
+        filter_by = self.params["by"]
```
AnnaKravchenko, line 55: Changing "score" to “haddockscore" in .yaml will require changing `filter_by` parameter back to “score", as it is an actual name of the attribute. But this is trivial, normally

```
if filter_by == ‘haddockscore’:
    filter_by = ‘score'
```

src/haddock/modules/analysis/filter/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+        threshold = self.params["cutoff"]
+
+        # Make sure we can access this attribute on models
+        models_with_attributes: list[PDBFile] = [
+            m for m in models
+            if getattr(m, filter_by, "UnexpectedValue") != "UnexpectedValue"
+            ]
+        # Check how many of them are available
```
AnnaKravchenko, line 63: Changing "score" to “haddockscore" in .yaml will require changing `filter_by` parameter back to “score", as it is an actual name of the attribute. But this is trivial, normally

```
if filter_by == ‘haddockscore’:
    filter_by = ‘score'
```

src/haddock/modules/analysis/filter/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+            p
+            for p in self.previous_io.output
+            if p.file_type == Format.PDB
+            ]
+        
+        # Get the filter by parameter
+        filter_by = self.params["by"]
+        threshold = self.params["cutoff"]
+
+        # Make sure we can access this attribute on models
```
AnnaKravchenko, line 58: Changing "score" to “haddockscore" in .yaml will require changing `filter_by` parameter back to “score", as it is an actual name of the attribute. But this is trivial, normally

```
if filter_by == ‘haddockscore’:
    filter_by = ‘score'
```

src/haddock/modules/analysis/filter/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+        models_with_attributes: list[PDBFile] = [
+            m for m in models
+            if getattr(m, filter_by, "UnexpectedValue") != "UnexpectedValue"
```
AnnaKravchenko, line 61: May-be log the % of model that have desired attribute ? 

Not sure if it's possible for only some - but not all - models to have it. But if so, could this info be handy for general user awareness and/or debuggin?

tests/test_module_filter.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Once a module is initialized, it should have the following attributes
+    assert filter.path == Path("0_anything")
+    assert filter._origignal_config_file == DEFAULT_CONFIG
+    assert type(filter.params) == dict
+    assert len(filter.params.keys()) != 0
+
+
+def test_filter_expected_behavior(filter, mocker, scored_models):
+    """Test finish_with_error due to wrong cluster parameter type."""
```
AnnaKravchenko, line 80: Seems here and lower functions’ descriptions are off

src/haddock/modules/analysis/filter/defaults.yaml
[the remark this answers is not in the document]
VGPReys: This was still under development, and not yet matured.

Maybe I will remove this parameter, as, for now, scoring modules (or CNS ones) are updaing the `.score` attributes to the PDBFile class.
This allows the `seletop` module and any kind of sorting to use this attribute to perform the task.

If we want to use an other score, then we need to write those score with a different attribute.
But some modules may modify complexes, and therefore a `.sasascore` score will no more be accurate with the current model.

So I think the `by` parameter will be simply dropped of, and we will point always to the `.score` attribute, which will be the last one computed.

src/haddock/modules/analysis/filter/__init__.py
VGPReys: no more relevant as there is no more `by` parameter

tests/test_module_filter.py
```diff
@@ -0,0 +1,127 @@
...
+    # In the mocked method, add the arguments that are called by the original
+    #  method that is being tested
+    def retrieve_models(self, individualize: bool = False):
+        """Provide a set of models."""
+        return self.models
+
+
+@pytest.fixture(name="scored_models")
+def fixture_scored_models() -> list[PDBFile]:
+    """Set of clustered PDBfiles."""
```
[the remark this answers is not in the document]
VGPReys: good catch, updated the doc strings

tests/test_module_filter.py
VGPReys: good catch, updated the doc strings

src/haddock/modules/analysis/filter/__init__.py
VGPReys: OK added

tests/test_module_filter.py
```diff
@@ -0,0 +1,115 @@
...
+    assert len(filter.output_models) == 5
+    # Check that their values is under the threshold
+    for m in filter.output_models:
+        assert m.score <= threshold
+
+
+def test_filter_cutoff_too_strigent(filter, mocker, scored_models):
+    """Test expected error when filtering cutoff is too strigent.
+    
+    Here we expect that if not models are passed to the next module,
```
mgiulini, line 101:
```suggestion
    Here we expect that if no models are passed to the next module,
```

mgiulini (approved):

tests/test_module_filter.py
VGPReys: not applying suggestion so you can still review !
But applied the suggestion on a different commit

AnnaKravchenko (approved):

PR #1232, opened 2025-03-28
adding random removal in haddock3-restraint cli

### LINKED ISSUE #1240: Add random removal of restraints

I would be nice to have a tool to manage the restraints a different way than the `npart` in CNS.

## **Motivation**

This would allow the fine tuning of the number of restraints, and not limit ourselves to 50%, 66%, 75%, 80%, etc...

VGPReys:
## Checklist

This pull request adds the `random_removal` sub-command to the `haddock3-restraints` CLI.

Given an input ambiguous file (.tbl), this subcommand will generate an archive containing multiple tbl files, each containing a subset of the initial ones.

The subset is tuned by the optional argument --ratio (-r), defining the ratio of restraints to be removed (between 0 and 1), 0 not doing anything and 1 removing everything...
The number of generated files in the archive is tuned by the argument --nb-tbl (-n)
The random seed can be tuned using --seed (-s), for reproducibility issues

Usage:
`haddock3-restraints random_removal <tbl_file> [-r <ratio>] [-n <nb-tbl>] [-s <seed>]`

```
options:
  -h, --help            show this help message and exit
  -r RATIO, --ratio RATIO
                        Ratio of restraints to be randomly removed.
  -s SEED, --seed SEED  Pseudo-random seed.
  -n NB_TBL, --nb-tbl NB_TBL
                        Number of ambiguous files to generate in the archive.
```

## Related Issue

Closes #1240

## Additional Info

The functions managing the random removal are placed in `librestraints.py`, allowing for their usage outside of this CLI.
For example, directly in the CNS modules, such as rigidbody, flexref, etc...


rvhonorato:
There is no issue associated with this PR - please add one before merging

src/haddock/clis/restraints/random_removal.py
```diff
@@ -0,0 +1,155 @@
...
+    Parameters
+    ----------
+    tblfile : Union[str, Path]
+        Path the the input ambiguous file
+    ratio : float
+        Ration of restraints to be removed
+    nb_tbl : int, optional
+        Number of ambig files to generate in the archive, by default 10
+    seed : int, optional
+        Initial random seed, by default 420
```
AnnaKravchenko, line 142: I suggest to include these defaul values to the help message

src/haddock/clis/restraints/random_removal.py
```diff
@@ -0,0 +1,155 @@
...
+        default=0.5,
+        type=float,
+        )
+
+    rand_removal_subcommand.add_argument(
+        "-s",
+        "--seed",
+        help="Pseudo-random seed.",
+        required=False,
+        default=917,  # Same as the one in various modules (iniseed)
```
AnnaKravchenko, line 56: Here default seed is 917, but later in code default is 420. I’d say - pick your favourite number out of two:)

src/haddock/clis/restraints/random_removal.py
```diff
@@ -0,0 +1,155 @@
...
+        type=str,
+        help="input tbl restraint file.",
+        )
+    
+    rand_removal_subcommand.add_argument(
+        "-r",
+        "--ratio",
+        help="Ratio of restraints to be randomly removed.",
+        required=False,
+        default=0.5,
```
AnnaKravchenko, line 47: This value also should be in the help message

AnnaKravchenko (changes requested):
One issue (or I misunderstood how ratio should work):
* My tbl contains 10 restraints and with default parameters, I get 10 files of 4 restraints each. I would expect to see 5 restraints in each file, not 4. Same for ratios of 0.2 and 0.7 - it’s always 1 restrain less than expected. Is it normal, or is there some rounding issue? 

And one suggestion (optional):
* I would add an extra foolproof layer - allow to generate only unique combinations of restraints. I.e., inclute max cut off for `-n` equal to the number of combination without repetition given total number of restraints and sample size.

VGPReys:
> One issue (or I misunderstood how ratio should work):
> 
>     * My tbl contains 10 restraints and with default parameters, I get 10 files of 4 restraints each. I would expect to see 5 restraints in each file, not 4. Same for ratios of 0.2 and 0.7 - it’s always 1 restrain less than expected. Is it normal, or is there some rounding issue?

Well pointed !
Issue in the code, the last assign statement was missing from the reading of the file.
Now corrected and added a test.

> And one suggestion (optional):
> 
>     * I would add an extra foolproof layer - allow to generate only unique combinations of restraints. I.e., inclute max cut off for `-n` equal to the number of combination without repetition given total number of restraints and sample size.

I get the idea, but because I do not enumerate all possible combination and loop over them, I cannot make sure they have all been generated, as the process picks random set of restraints instead of looping over the list of possible combinations.
But you are right, to some extend, they may be redundancies if you have a low number of restraints and a high number of files to be generated.

AnnaKravchenko:
> I get the idea, but because I do not enumerate all possible combination and loop over them, I cannot make sure they have all been generated, as the process picks random set of restraints instead of looping over the list of possible combinations. But you are right, to some extend, they may be redundancies if you have a low number of restraints and a high number of files to be generated.

Ok! Then could you add little note at the end of sucesful run? Smth simple, e.g. in src/haddock/clis/restraints/random_removal.py, 
```

print(
        f"Generated archive path: {archive_fpath}!{os.linesep}”
        f”Note that resulting restraints might be redundant"
        f"By using this file, we suggest to:{os.linesep}"
        f"- turn off random removal (random_removal = false){os.linesep}"
        "- turn on the previous ambig in later stages (previous_ambig = true)"
        )
```

AnnaKravchenko:
I am testing on the file with 10 restraints (`/trinity/login/arha/dev-h3/test-pr/ambig-test-10.tbl`) and smth is off with ratio vs actual number of kept restraints:

1.  `-r 0.8` returns files with only 1 restraint, while it should contain 2 restraints. 
2. `-r 0.9` and `-r 0.9 -n 1` fails with:
```
  File "/trinity/login/arha/tools/miniforge3/envs/hdd3/bin/haddock3-restraints", line 8, in <module>
    sys.exit(maincli())
  File "/trinity/login/arha/dev-h3/haddock3/src/haddock/clis/cli_restraints.py", line 125, in maincli
    args.func(**cmd)
  File "/trinity/login/arha/dev-h3/haddock3/src/haddock/clis/restraints/random_removal.py", line 149, in main
    archive_fpath = random_removal(tblfile, ratio, nb_tbl=nb_tbl, seed=seed)
  File "/trinity/login/arha/dev-h3/haddock3/src/haddock/clis/restraints/random_removal.py", line 110, in random_removal
    subset_restraints = next(subsets_restraints)
  File "/trinity/login/arha/dev-h3/haddock3/src/haddock/libs/librestraints.py", line 792, in get_restraint_subset
    raise ValueError(f"Error: Subset of restraints equal to 0")
ValueError: Error: Subset of restraints equal to 0

```

3. Another foolprofing that could be nice: atm  -n can be zero or negative, e.g.:
```
haddock3-restraints random_removal ambig-test-10.tbl -r 0.5 -n -3
Generated archive path: /trinity/login/arha/dev-h3/test-pr/ambig-test-10.tbl.tgz!
By using this file, we suggest to:
- turn off random removal (random_removal = false)
- turn on the previous ambig in later stages (previous_ambig = true)
```
In this case the archive is empty. Would be nice to give ‘invalid value’ error message for this case

VGPReys:
Good detective, thanks for trying the subcommand so deeply.
Indeed, this time it was rounding issues.
- I tried to fix it.
- I also improved the error messages
- implemented to check of number of files to generate to be `>0`

PR #1317, opened 2025-07-03
Per-residue parallelization in alasan 

### LINKED ISSUE #1307: Parallelize computations in [alascan]
Mutation of each residure in the interface is performed sequentially, so it’s rather long, e.g. 9 min for 40 interface residues. I expect parallelization to speed things us sigificantly.

AnnaKravchenko:
`not applicable` >> Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)

Alascan already parallelizes per-model computations. But if we’re working with a few models, a large number of interface residues, and a sufficient number of available cores (after per-model parallelization), then per-residue parallelization speeds up alascan.
This is achieved by introducing InterfaceScanner class, which creates a list of mutation jobs per model. Nextly, all mutations are merged in a single list of jobs, and the ModelPointMutation class takes care of these individual mutations. Lastly, performed mutations are associated with the relevant model and saved in a .tsv with model_name.
For both classes, parallelization is handled by the haddock engine.

Additionally, there is a parameterised library_mode in InterfaceScanner class, which is set to False in alascan.inint.py, and the pipeline runs as described above. However, if library_mode = True (default in scan.py), then InterfaceScanner.run() won’t just return a list of mutation jobs, but perform these jobs sequentially, writing results to the file (1 file per model, as usual). This allows to run alascan as a stand-alone library by simply exporting scan.py
## Related Issue
Closes #1307

## Additional info
During review of this PR, the idea of parallelizing calculations for each cluster arose. It is now a separate isse #1333

### DIFF
--- a/src/haddock/modules/analysis/alascan/scan.py
+++ b/src/haddock/modules/analysis/alascan/scan.py
@@ -590,7 +590,7 @@ class InterfaceScanner:
             cutoff = self.params.get("int_cutoff", 5.0)
             interface = CAPRI.identify_interface(self.model_path, cutoff=cutoff)
             
-            # get user_chains for the chek down the line
+            # get user_chains for the check down the line
             user_chains = self.params.get("chains", [])
 
             # if user defined target residues, check they are in the interface
@@ -598,15 +598,14 @@ class InterfaceScanner:
                 filtered_interface = {}
                 for chain in self.filter_resdic:
                     if chain in interface:
-                        user_res_valid = []
-                        for res in self.filter_resdic[chain]:
-                            if res in interface[chain]:
-                                user_res_valid.append(res)
+                        # Search for the intersection of user queried residues and interface residues
+                        user_res_valid = list(set(self.filter_resdic[chain]).intersection(set(interface[chain])))
+                        # If at least one residue must be analyzed, add it to residues to be scanned
                         if user_res_valid:
-                            user_res_valid_unique = list(set(user_res_valid))
-                            filtered_interface[chain] = user_res_valid_unique
+                            filtered_interface[chain] = user_res_valid
                 interface = filtered_interface
-            # if (user defined targer chains) & (no user target residues) - do use user chains
+
+            # if (user defined target chains) & (no user target residues) - do use user chains
             elif user_chains:
                 interface = {chain: res for chain, res in interface.items() if chain in user_chains}
 
@@ -636,7 +635,7 @@ class InterfaceScanner:
                             output_mutants=output_mutants
                         )
                         self.point_mutations_jobs.append(job)
-
+    
             # Execute jobs if in library mode
             if self.library_mode:
                 log.info(f"Executing {len(self.point_mutations_jobs)} mutations for {self.model_id}")
--- a/tests/test_module_alascan.py
+++ b/tests/test_module_alascan.py
@@ -684,7 +684,9 @@ def test_model_point_mutation_run(mocker, mutation_job, monkeypatch):
         assert result.res_num == 19
         assert result.ori_resname == "THR"
         assert result.target_resname == "ALA"
-        assert result.mutant_scores == mutant_scores
+        # check resulted mutant_scores aproximately match expected mutant_scores
+        for result_score, expected_score in zip(result.mutant_scores, mutant_scores):
+            assert pytest.approx(result_score, abs=0.2) == expected_score
         # test cleanup was called
         mock_mutate.assert_called_once_with(mutation_job.model_path, "A", 19, "ALA")
         mock_calc_score.assert_called_once()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "haddock3"
-version = "2025.07.0"
+version = "2025.07.1"
 description = "HADDOCK3"
 readme = "README.md"
 authors = [{ name = "BonvinLab", email = "bonvinlab.support@uu.nl" }]


src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -84,45 +86,59 @@ def _run(self):
...
-        ncores = parse_ncores(n=self.params['ncores'], njobs=nmodels)
-
-        log.info(f"Running on {ncores} cores")
-
-        index_list = get_index_list(nmodels, ncores)
-
+        # Core maths: Is there enough cores to parallelize per-residue after
+        # giving enough cores per model ?
+        ncores = parse_ncores(n=self.params['ncores'])
+        model_cores, residue_cores_per_model = calculate_core_allocation(nmodels, ncores)
```
rvhonorato: this PR caught my attention and I have a comment: 

check if this logic does not overlap with the scheduler task splitting to make sure you are not doing it twice.  the scheduler will already split the tasks across the processors:

https://github.com/haddocking/haddock3/blob/86d755812ca309a952c7ffb3740009023f3ba24f/src/haddock/libs/libparallel.py#L148

Would it not be possible to create a huge batch of tasks in which each task would be one model with one mutation, so it would be `modelN_mutationX` and then retrieve the results and organize it into a data structure? This would make sure there is a flat hierarchy of tasks which should make it simpler - or this what you are already doing here? By having a flat hierarchy of tasks you could also avoid this calculation of cores and simplify the code downstream.

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -1,21 +1,21 @@
...
 
-This module is responsible for the alanine (or any other residue) scan analysi
-of the models generated in the previous step of the workflow.
-For each model, the module will mutate the interface residues and calculate
-the energy differences between the wild type and the mutant, thus providing a
-measure of the impact of such mutation. Such difference (delta_score) is always
+This module is responsible for the alanine (or any other residue) scan analysis
+of the model(s) generated in the previous step of the workflow.
+For each model, this module will mutate the interface residues and calculate
+the haddock score differences between the wild type and the mutant, thus providing
```
VGPReys: Not only the haddock score is displayed but also the components
```suggestion
the energy components and haddock score differences between the wild type and the mutant, thus providing
```

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -84,45 +86,59 @@ def _run(self):
...
         nmodels = len(models)
-        # output mutants is only possible if there is only one model
+        # Outputting mutants is possible if there is just one model
         if self.params["output_mutants"]:
             if nmodels != 1:
                 log.warning(
-                    "output_mutants is set to True but more than one model "
-                    "was found. Setting output mutants to False."
+                    "output_mutants is set to True, but more than one model "
+                    "was found. Setting output_mutants to False."
```
VGPReys:
```suggestion
                    "was found. Setting 'output_mutant' parameter to False."
```

src/haddock/modules/analysis/alascan/__init__.py
VGPReys: I agree with rodrigo's comment.

Because at the beginning you do not know how many mutations you will have to perform (because provided models may be different), you could:
- have a first set of jobs that retrieve all interface residues that must be mutated in each model (this can be performed by the scheduler)
- then setup a list of point-mutations jobs (`modelN_mutationX`) that can be submitted to the scheduler (again).
- finally retrieve all data and merge them back per-model to produce the output

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -574,12 +679,12 @@ def run(self):
 
             fl_content = open(alascan_fname, 'r').read()
             with open(alascan_fname, 'w') as f:
-                f.write(f"##########################################################{os.linesep}")  # noqa E501
+                f.write(f"#######################################################################{os.linesep}")  # noqa E501
```
VGPReys:
```suggestion
                f.write(f"{'#' * 70}{os.linesep}")
```

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -574,12 +679,12 @@ def run(self):
...
                 f.write(f"# `alascan` results for {native.file_name}{os.linesep}")  # noqa E501
                 f.write(f"#{os.linesep}")
                 f.write(f"# native score = {n_score}{os.linesep}")
                 f.write(f"#{os.linesep}")
                 f.write(f"# z_score is calculated with respect to the other residues")  # noqa E501
                 f.write(f"{os.linesep}")
-                f.write(f"##########################################################{os.linesep}")  # noqa E501
+                f.write(f"#######################################################################{os.linesep}")  # noqa E501
```
VGPReys:
```suggestion
                f.write(f"{'#' * 70}{os.linesep}")
```

src/haddock/modules/analysis/alascan/__init__.py
AnnaKravchenko: Yeah, good idea, thank you!

src/haddock/modules/analysis/alascan/__init__.py
AnnaKravchenko: How about  “haddock score and its individual components"?

VGPReys: yes of course, it was just a suggestion :)

src/haddock/modules/analysis/alascan/__init__.py
VGPReys: Upon further thoughts, I just realized that by breaking down the computation to a `modelN-mutationX` basis, it will be difficult for a user to perform an Alanine scanning by importing the Class object itself (if a user want to use haddock3 as a library for example).
Because of now, it was quite straight forward: You want to perform alanine scanning of a model, just use the `Scan` class with an input model.

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -1,12 +1,15 @@
 """alascan module."""
 import os
-from pathlib import Path
+import io
 import shutil
+from pathlib import Path
+from contextlib import redirect_stdout
+from dataclasses import dataclass
+from typing import List, Dict, Tuple, Optional, Any
```
VGPReys: rather import those from the `haddock.core.typing`

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -411,73 +415,171 @@ def create_alascan_plots(clt_alascan, scan_residue, offline = False):
...
 
 
-class ScanJob:
-    """A Job dedicated to the parallel alanine scanning of models."""
-
-    def __init__(
-            self,
-            params,
-            scan_obj):
+def write_scan_out(results, model_id):
```
VGPReys: - missing type hints of function arguments
- missing docstring of for the function

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -488,98 +590,154 @@ def run(self):
...
-
-            self.df_scan.to_csv(
-                alascan_fname,
-                index=False,
-                float_format='%.2f',
-                sep="\t"
-                )
+                        log.warning(f"Failed: {result.error_msg}")
+                
+                write_scan_out(results, self.model_id)
```
VGPReys, line 654: If you return the self.point_mutations_jobs here, you can avoid the creation of the `InterfaceScannerJob` class, and therefore simplify the `__init__.py` content
```suggestion
                write_scan_out(results, self.model_id)
            else:
                return self.point_mutations_jobs
```

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -411,73 +415,171 @@ def create_alascan_plots(clt_alascan, scan_residue, offline = False):
...
-        if "params" in params.keys():
-            self.filter_resdic = {
-                key[-1]: value for key, value
-                in params["params"].items()
-                if key.startswith("resdic")
-                }
+        return self.scan_obj.point_mutations_jobs
 
 
+class MutationJob:
```
VGPReys: You can completely delete this guy, as it is doing the same as the `ModelPointMutation`

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -41,22 +41,23 @@
...
-from haddock.modules import BaseHaddockModule
-from haddock.modules import get_engine
+from haddock.modules import BaseHaddockModule, get_engine
 from haddock.modules.analysis import get_analysis_exec_mode
 from haddock.modules.analysis.alascan.scan import (
-    Scan,
-    ScanJob,
+    InterfaceScanner,
+    InterfaceScannerJob,
+    MutationJob,
```
VGPReys: Both can be removed

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -84,60 +85,85 @@ def _run(self):
...
-        alascan_jobs = []
-        for core in range(ncores):
-            scan_obj = Scan(
-                model_list=models[index_list[core]:index_list[core + 1]],
-                core=core,
-                path=Path("."),
+        
+        # Step1: "get mutations" i.e. get target interface residues per input model  
+        scan_objects = []
+        scan_jobs = []
```
VGPReys: One of the two can be removed

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -84,60 +85,85 @@ def _run(self):
...
+    
+        # Merge mutations per-model "scan_obj.point_mutations_jobs" in a single mutation_objects 
+        mutation_objects = []
+        for scan_obj in scan_objects:
+            mutation_objects.extend(scan_obj.point_mutations_jobs)
+        total_mutations = len(mutation_objects)
+        log.info(f"Found {total_mutations} mutations")
+        # wrap all mutations in mutation_jobs 
+        if mutation_objects:
+            mutation_jobs = [MutationJob(mutation_obj) for mutation_obj in mutation_objects]
```
VGPReys: keep either the `mutation_objects` or `mutation_jobs`

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -488,98 +590,154 @@ def run(self):
...
+        except Exception as e:
+            log.error(f"Failed to scan model {self.model_id}: {e}")
+            raise
+
+
+class ModelPointMutation:
+    """Executes a single point mutation."""
+    
+    def __init__(self, model_path, model_id, chain, res_num, ori_resname, 
+                 target_resname, native_scores, output_mutants=False):
```
VGPReys: missing type hint on arguments

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -488,98 +590,154 @@ def run(self):
...
+        self.output_mutants = output_mutants
+    
+    def run(self):
+        """Execute the point mutation."""
+        mutation_id = f"{self.model_id}_{self.chain}{self.res_num}{self.target_resname}"
+        
+        try:
+            # Setup working directory
+            process_id = multiprocessing.current_process().pid
+            sc_dir = f"haddock3-score-{mutation_id}-{process_id}"
```
VGPReys: why do you need the `process_id` here ?
is not already the `mutation_id` variable unique ?

AnnaKravchenko: True!

src/haddock/modules/analysis/alascan/scan.py
AnnaKravchenko: 1. I actually need only List, Tuple, Optional, Any
2. haddock.core.typing doesn’t have List and Tuple, so importing these from typing

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -84,60 +83,83 @@ def _run(self):
...
+            results_by_model = {}
+            for result in engine.results:
+                if result and result.success:
+                    if result.model_id not in results_by_model:
+                        results_by_model[result.model_id] = []
+                    results_by_model[result.model_id].append(result)
+
+            # Save to .tsv
+            for model_id, results in results_by_model.items():
+                write_scan_out(results, model_id)
```
VGPReys, line 140: un-parallelized task here.
Do you think it would be also worth doing it now already ?

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -84,60 +83,83 @@ def _run(self):
...
+                    offline=self.params["offline"],
                 )
-        # if output_bfactor is true, write the models and export them
-        if self.params["output_bfactor"] is True:
-            models_to_export = generate_alascan_output(models, self.path)
-            self.output_models = models_to_export
+
+            # Generate output models with bfactors if requested  
+            if self.params["output_bfactor"]:
+                models_to_export = generate_alascan_output(models, self.path)
```
VGPReys, line 155: un-parallelized task here.
Do you think it would be also worth doing it now already ?

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -84,60 +83,83 @@ def _run(self):
...
+
+            # Cluster-based analysis
+            clt_alascan = alascan_cluster_analysis(models)
+            
+            # Generate plots if requested
+            if self.params["plot"]:
+                create_alascan_plots(
+                    clt_alascan,
+                    self.params["scan_residue"],
+                    offline=self.params["offline"],
```
VGPReys, line 150: un-parallelized task here.
Do you think it would be also worth doing it now already ?

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -1,12 +1,15 @@
 """alascan module."""
 import os
-from pathlib import Path
+import io
 import shutil
+from pathlib import Path
+from contextlib import redirect_stdout
+from dataclasses import dataclass
+from haddock.core.typing import Optional, Any
```
VGPReys: I would move that guy lower together with the the other haddock imports

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -411,73 +415,150 @@ def create_alascan_plots(clt_alascan, scan_residue, offline = False):
...
+    mutant_scores: Tuple[float, float, float, float, float]  
+    delta_scores: Tuple[float, float, float, float, float]   
+    success: bool
+    error_msg: Optional[str] = None
+
+
+class InterfaceScanner:
+    """Scan interface of a model to get tartget residues and create corresponding mutation jobs."""
+    
+    def __init__(self, model, mutation_res="ALA", params=None, library_mode=True):
```
VGPReys: Missing type hints here

src/haddock/modules/analysis/alascan/__init__.py
AnnaKravchenko: Actually no, parallelization here requires making yet another class and calling engine again, so making code more complex for seemingly very little time gain. I’d rather keep it asis

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -429,175 +434,325 @@ def create_alascan_plots(clt_alascan, scan_residue, offline = False):
...
+            atoms = get_atoms(self.model_path)
+            coords, chain_ranges = load_coords(self.model_path, atoms, add_resname=True)
             
-            # check if the user wants to mutate only some residues
+            # Determine target residues: get interface, then apply user filers, if given 
+            # Get all interface residues        
+            cutoff = self.params.get("int_cutoff", 5.0)
+            interface = CAPRI.identify_interface(self.model_path, cutoff=cutoff)
+            
+            # get user_chains for the chek down the line
```
VGPReys: chek -> check

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -429,175 +434,325 @@ def create_alascan_plots(clt_alascan, scan_residue, offline = False):
...
-                        }
-            
+                    if chain in interface:
+                        user_res_valid = []
+                        for res in self.filter_resdic[chain]:
+                            if res in interface[chain]:
+                                user_res_valid.append(res)
+                        if user_res_valid:
+                            user_res_valid_unique = list(set(user_res_valid))
+                            filtered_interface[chain] = user_res_valid_unique
```
VGPReys: Here you are searching for the intersection of two sets.

what about:
```suggestion
                        # Search for the intersection of user queried residues and interface residues
                        user_res_valid = list(set(self.filter_resdic[chain]).intersection(set(interface[chain])))
                        # If at least one residue must be analyzed, add it to residues to be scanned
                        if len(user_res_valid) >= 1:
                            filtered_interface[chain] = user_res_valid
```
?

tests/test_module_alascan.py
```diff
@@ -251,57 +296,536 @@ def test_init(alascan):
...
+        result = mutation_job.run()
+        # test successful result
+        assert isinstance(result, MutationResult)
+        assert result.success is True
+        assert result.model_id == "test_model"
+        assert result.chain == "A"
+        assert result.res_num == 19
+        assert result.ori_resname == "THR"
+        assert result.target_resname == "ALA"
+        assert result.mutant_scores == mutant_scores
```
VGPReys: Because of CNS and possible variations due to randomness, I would do approximate tests:
```suggestion
        for result_score, expected_score in zip(result.mutant_scores, mutant_scores):
            assert pytest.approx(result_score, abs=0.2) == expected_score
```

VGPReys:
> Last (hopefuly) thing - idk which haddock verison it should be. Current one is `version = “2025.07.0”`, not sure what to update it to.

version = “2025.07.1”

src/haddock/modules/analysis/alascan/scan.py
AnnaKravchenko: yes, great suggestion!

VGPReys (approved):
OK great !

I think we reached the desired outcome of this PR.
- [X] it is producing the same output as the previous one 🪨 
- [X] it is well tested 🧪 
- [X] it is faster 🐎 

Good job ! 🚀

VGPReys:
One last thing, the PR description must now be updated 😈

PR #1322, opened 2025-07-14
Alascan parallel tests

AnnaKravchenko:
Figuring out a way to fix broken CI tests.


AnnaKravchenko:
AlasacnWorker and AlascanScheduler can take care of per-residue parallelization and keyboard interrupt. 
These two are in libparallel. Added tests for new functions there, all passing. 

Next: add test to scan.py itself. Check keyboard interrupt on baseline alascan.

AnnaKravchenko:
Bottom line: let ctrl+c cause leaks, this should be a separate issue

PR #1337, opened 2025-07-30
Increase parallelism in alascan

### LINKED ISSUE #1333: Improves parallel tasks in [alascan]

Add an extra layer of parallelism in the final stages of the `[alascan]` module.

## **Motivation**

Several late stages of the `[alascan]` module are not parallel yet:
- Performing cluster analysis
- Generating plots for clusters
- Adding bfactor to models

## **Additional context**

Continuation of the effort started in #1317 to make the module faster

VGPReys:
## Checklist

- [X] Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)
- [X] Haddock3 `version` has been incremented in `pyproject.toml`

- [X] parallelized file writting

## Related Issue

Closes #1333

### DIFF
--- a/src/haddock/modules/analysis/alascan/__init__.py
+++ b/src/haddock/modules/analysis/alascan/__init__.py
@@ -46,15 +46,14 @@ from pathlib import Path
 
 from haddock import log
 from haddock.core.defaults import MODULE_DEFAULT_YAML
-from haddock.libs.libutil import parse_ncores
 from haddock.modules import BaseHaddockModule, get_engine
 from haddock.modules.analysis import get_analysis_exec_mode
 from haddock.modules.analysis.alascan.scan import (
+    AddDeltaBFactor,
     InterfaceScanner,
     write_scan_out,
     alascan_cluster_analysis,
     create_alascan_plots,
-    generate_alascan_output,
 )
 
 
@@ -93,17 +92,17 @@ class HaddockModule(BaseHaddockModule):
                     "was found. Setting 'output_mutant' parameter to False.")
                 self.params["output_mutants"] = False
         
-        # Step1: "get mutations" i.e. get target interface residues per input model  
-        scan_objects = []
-        for model in models:
-            # 1 scan_obj per input model, merged into scan_objects to give to Engine
-            scan_obj = InterfaceScanner(
+        # Step1: "get mutations" i.e. get target interface residues per input model
+        # 1 scan_obj per input model, merged into scan_objects to give to Engine  
+        scan_objects = [
+            InterfaceScanner(
                 mutation_res=self.params["scan_residue"],
                 model=model,
                 params=self.params,
                 library_mode = False
-            )
-            scan_objects.append(scan_obj)
+                )
+            for model in models
+            ]
 
         log.info(f"Scanning {nmodels} models for possible mutations")
         exec_mode = get_analysis_exec_mode(self.params["mode"])
@@ -112,12 +111,11 @@ class HaddockModule(BaseHaddockModule):
         engine.run()
 
         # Step2: perform mutations
-        # Collect mutations from the engine output 
+        # Collect all point mutations to be performed
         mutation_objects = []
-        for i, scan_obj in enumerate(scan_objects):
-            if i < len(engine.results) and engine.results[i]:
-                #scan_obj.point_mutations_jobs = engine.results[i]
-                mutation_objects.extend(engine.results[i]) 
+        for mutations_to_perform in engine.results:
+            if mutations_to_perform:
+                mutation_objects.extend(mutations_to_perform) 
 
         total_mutations = len(mutation_objects)
         log.info(f"Found {total_mutations} mutations")
@@ -152,10 +150,16 @@ class HaddockModule(BaseHaddockModule):
 
             # Generate output models with bfactors if requested  
             if self.params["output_bfactor"]:
-                models_to_export = generate_alascan_output(models, self.path)
+                update_with_bfactor_jobs = [
+                    AddDeltaBFactor(model, self.path)
+                    for model in models
+                ]
+                engine = Engine(update_with_bfactor_jobs)
+                engine.run()
+                models_to_export = engine.results
                 self.output_models = models_to_export
             else:
-                # # Send models to the next step, no operation is done on them 
+                # Send models to the next step, no operation is done on them 
                 self.output_models = models
         else:
             log.info("No interface residues found - skipping mutation analysis")
--- a/src/haddock/modules/analysis/alascan/scan.py
+++ b/src/haddock/modules/analysis/alascan/scan.py
@@ -5,15 +5,16 @@ import shutil
 from pathlib import Path
 from contextlib import redirect_stdout
 from dataclasses import dataclass
-from typing import List, Tuple, Union, Dict
+from typing import List, Tuple, Dict
 
 import numpy as np
 import pandas as pd
 
 
 from haddock import log
-from haddock.core.typing import Optional, Any
+from haddock.core.typing import Any, Optional, Union
 from haddock.libs.libalign import get_atoms, load_coords
+from haddock.libs.libontology import PDBFile
 from haddock.libs.libplots import make_alascan_plot
 from haddock.modules.analysis.caprieval.capri import CAPRI
 from haddock.clis import cli_score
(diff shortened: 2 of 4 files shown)


VGPReys:
> Looking good and working fine.
> 
> I would however also modify the examples under the analysis directory to remove the line defining the number of cores, so that the example benefit from this parallel execution.
> 
> Or increase the number of cores to be used (e.g to 10)

Good point, I already did this internally to see the effect.
Let's increase it !

amjjbonvin (dismissed):

examples/analysis/alascan-complex-test.cfg
```diff
@@ -10,7 +10,7 @@
 
 # General parameters
 run_dir = "run1-alanine-scan-complex"
-ncores = 1
+ncores = 1)
```
VGPReys:
```suggestion
ncores = 10
```

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -259,44 +216,53 @@ def add_zscores(df_scan_clt, column='delta_score'):
...
-        cl_pops[cl_id] += 1
+        clt_pops[cl_id] += 1
         # read the scan file
         alascan_fname = f"scan_{native.file_name.removesuffix('.pdb')}.tsv"
         df_scan = pd.read_csv(alascan_fname, sep="\t", comment="#")
-        # loop over the scan file
+        # loop over the scan file rows
         for row_idx in range(df_scan.shape[0]):
             row = df_scan.iloc[row_idx]
+            # Extra data related to residue information
```
VGPReys:
```suggestion
            # Extract data related to residue information
```

AnnaKravchenko:
I’m still not fully convinsed if writting .tsv per model worth parallelizing. Just out of curiosity - did you test the time per big amount of models?

VGPReys:
currently benchmarking the impact of the current modifications

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -136,26 +135,42 @@ def _run(self):
...
+                    clt_id,
+                    clt_pops[clt_id],
+                    scan_residue=self.params["scan_residue"],
+                    generate_plot=self.params["plot"],
                     offline=self.params["offline"],
                 )
+                for clt_id, clt_data in clt_scan.items()
+            ]
+            engine = Engine(alascan_cluster_jobs)
+            engine.run()
```
rvhonorato: quick comment: if I understand the logic, this will first write many files in parallel and second, group these files also in parallel.

if that is the case (i did not look much into the code), then you are using the filesystem to store temporary information - this is unnecessary and add a lot of I/O.

I have already added the logic to the scheduler to retrieve results directly from the tasks, to avoid exactly this type of issue (using the filesystem as a channel)

https://github.com/haddocking/haddock3/blob/69128756ea5d7fab3af10aa9b255651561bf1483/src/haddock/libs/libparallel.py#L175-L190

there is an implementation that uses this design here:

https://github.com/haddocking/haddock3/blob/69128756ea5d7fab3af10aa9b255651561bf1483/src/haddock/modules/analysis/caprieval/__init__.py#L221-L225

VGPReys: Yes, yes, this logic is already used in the code.

In this case:
- We want to write down per-model data
- We want to write down per-cluster data

All intermediates are stored in memory already:
- per-residue data are stored in memory and merged after only

But indeed, there is a file-reading step still performed that could be further optimized

VGPReys: Done

rvhonorato:
fyi:

from 8.8s (1 core) to 0.9s (40 cores) = + ~25% efficiency
from 7.9s (1 core) to 4.9s (40 cores) = + ~4% efficiency

these are both in the low end (should be 100%) which indicate there is a significant overhead not being caught by the parallelization.

rvhonorato:
if these bottlenecks are critical and largely contribute to the overall execution - you can also consider shifting these functions to a zero-cost abstraction language such as rust/c/c++ and writing the bindings. with zero-cost abstraction there would be no overhead and would probably reach 95% efficiency speedup (accounting for the binding). doing some quick math it would technically be:
- from 8.8 sec (1 core) to ~0.9 sec~ ~210ms (40 cores)
- from 7.9 sec (1 core) to ~4.9 sec~ ~190ms (40 cores)

it might not be worth the development time to try to speed this up further, but maybe important to keep in mind that python has a bottleneck for this sort of optimizations

CHANGELOG.md
```diff
@@ -7,3 +7,4 @@
 - 2025-07-23: Added printing of covalent energies to the PDB headers - Issue #1323
 - 2025-07-24: Added computation of standard deviation in alascan cluster analyses - Issue #1332
 - 2025-07-30: Added per-residue parallelization in alascan module - Issue #1307
+- 2025-07-31: Increasing parallelism in alascan module - Issue #1333
```
VGPReys:
```suggestion
- 2025-08-01: Increasing parallelism in alascan module - Issue #1333
- 2025-08-01: Fix alascan bug when no contact found - Issue #1343
```

amjjbonvin (dismissed):
Looks good on my side - but I leave it to the pros to approve

amjjbonvin (dismissed):

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -259,76 +231,50 @@ def add_zscores(df_scan_clt, column='delta_score'):
...
+            scan_residue: str = "ALA",
+            generate_plot: bool = False,
+            offline: bool = False,
+            ):
+        """Initialization function
+
+        Parameters
+        ----------
+        cluster_scan_data : Dict[str, Dict[str, Union[float, int]]]
+            Dictionary containing alascan data per residue identifyer.
```
AnnaKravchenko: little typo: identifyer -> identifier

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -259,76 +231,50 @@ def add_zscores(df_scan_clt, column='delta_score'):
...
+        clt_id : str
+            Cluster identifyer
```
AnnaKravchenko: same here, identifier

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -259,76 +231,50 @@ def add_zscores(df_scan_clt, column='delta_score'):
...
+            Defines if the plot should be functional offline, by default False
+        """
+        self.cluster_scan_data = cluster_scan_data
+        self.clt_id = clt_id
+        self.clt_population = clt_population
+        self.scanned_residue = scan_residue
+        self.generate_plot = generate_plot
+        self.offline = offline
+
+    def run(self) -> str:
```
AnnaKravchenko, line 269: missing description

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -259,76 +231,50 @@ def add_zscores(df_scan_clt, column='delta_score'):
...
+        # Gather all data in a list
         clt_data = []
         # Loop over residues
-        for ident in clt_scan[cl_id]:
+        for ident, clt_res_dt in self.cluster_scan_data.items():
             # Split identifyer to retrieve residue data
```
AnnaKravchenko: typo identifier

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -356,91 +302,183 @@ def alascan_cluster_analysis(models):
...
+        self.add_delta_to_bfactor(output_fname)
+        # Update attributes of the model
+        self.model.ori_name = self.model.file_name
+        self.model.file_name = output_fname
+        self.model.full_name = output_fname
+        self.model.rel_path = Path("..", Path(self.path).name, output_fname)
+        self.model.path = str(Path(".").resolve())
+        return self.model
+    
+    def add_delta_to_bfactor(self, output_path: Path) -> Path:
```
AnnaKravchenko: This name is confusing at a glance, gives impression we perform summation of b-factor values with delta_score values - which does not make sense whatsoever.  

Maybe rename to “write_delta_score_to_pdb”?
And make description more verbose, e.g.:
“ Write delta scores to PDB file, specifically to column reserved for b-factor vlaues”? 
It’s more verbose but also more clear.

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -356,91 +302,183 @@ def alascan_cluster_analysis(models):
...
+                        norm_delta = 0.0
+                    # Generate the bfactor string
+                    delta_str = f"{norm_delta:.2f}".rjust(6, " ")
+                    # Modify the PDB line
+                    line = line[:60] + delta_str + line[66:]
+                fout.write(line)
+        return output_path
+
+    def reorder_results(self) -> None:
+        """Perform initial data manuputation to simply downstream access."""
```
AnnaKravchenko: "initial data manuputation” is not quite informative, maybe replace with “Organise mutation results into distionary (key example “A-115"), and determine min and max score”?

src/haddock/modules/analysis/alascan/scan.py
VGPReys:
```suggestion
            Dictionary containing alascan data per residue identifier.
```

src/haddock/modules/analysis/alascan/scan.py
VGPReys:
```suggestion
            Cluster identifier
```

src/haddock/modules/analysis/alascan/scan.py
VGPReys:
```suggestion
            # Split identifier to retrieve residue data
```

src/haddock/modules/analysis/alascan/scan.py
VGPReys, line 269:
```suggestion
    def run(self) -> str:
        """Main function generating files related to alascan cluster analysis.
        
        Return
        ------
        scan_clt_filename : str
            Name of the tsv file written
        """
```

AnnaKravchenko:
I see now there’s “scan_clt_unclustered”. How about “scan_clt_ss”, so it’s harmonized with `caprieval` output?

src/haddock/modules/analysis/alascan/__init__.py
```diff
@@ -93,17 +93,17 @@ def _run(self):
                     "was found. Setting 'output_mutant' parameter to False.")
```
AnnaKravchenko: current log warning looks like:
 output_mutants is set to True but more than one model was found. Setting 'output_mutant' parameter to False.
 
 I’d change to more harmonized:
‘output_mutants’ parameter is set to True but more than one model was found. Setting 'output_mutant' to False.

src/haddock/modules/analysis/alascan/scan.py
```diff
@@ -267,11 +267,18 @@ def __init__(
         self.offline = offline
 
     def run(self) -> str:
+        """Main function generating files related to alascan cluster analysis.
```
AnnaKravchenko: Maybe go more explisit? E.g.:
“Wrtie per-cluster alascan output to scan_clt_X.tsv file, including averag and dev per mutation and generate and save per-cluster alascan plots (optional)"

AnnaKravchenko:
Ok, things are looking good and working as expected, both when mutations are found, and when there’s non (due to no interface or user residues outside of the interface)! 

So just those superficial comments to resolve, and 🚀

VGPReys:
> I see now there’s “scan_clt_unclustered”. How about “scan_clt_ss”, so it’s harmonized with `caprieval` output?

I am not convinced:
- capri_ss: is for single structure data (I guess)
- capri_clt: is for clutered data

In this case, we report clustered data (data averaged at the cluster level) for model that are not clustered. Which is different than single structure data

PR #1352, opened 2025-08-22
Cg haddock topocg fix

rversin:

## Checklist

- [ ] Haddock3 `version` has been incremented in `pyproject.toml`

In this branch, we correct the files used for the "ligands", renamed as "cofactors". Furthermore, when a cofactor is encountered in topocg, it will now be ignored. However, cgtoaa will regenerate it.

### DIFF
--- a/src/haddock/libs/libaa2cg.py
+++ b/src/haddock/libs/libaa2cg.py
@@ -409,6 +409,10 @@ def map_cg(chain):
         segid = aares.segid.strip()
         resi = aares.id[1]
 
+        if resn not in cg_mapping:
+            log.warning(f"Residue {resn} not in cg_mapping — skipping.")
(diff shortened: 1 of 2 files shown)


VGPReys:
> No this is the
> 
> > If I understand well, here we will be using a mixture of Martini2 and OPLS in the case of known/haddock-supported molecules that are not amino-acids/nucleic-bases ?
> 
> No this is the back-mapping from CG to AA. At this stage the CG model is only used as reference to morph the AA structures. No energetics from the CG part anymore (except for the distance restraints to the beads)

OK, it makes more sense !
But I guess that there is no back-mapping beads for unknown (in the CG force-field) residues.
Or there is some kind of trick I am not aware of yet ?

PR #1370, opened 2025-09-03
add action to tag and release on push

rvhonorato:

## Checklist

DOES NOT APPLY

Currently the creation of tag and release are manually done, which means that often the version of the repository is not in line with what is available in the pypi and also not tagged here in the repository.

This new action will create a new tag based on the version defined in pyproject.toml as well as a release, which should then be automatically published to pypi.

It will eliminate the version drift and ensure consistent deployment of the application.

## Related Issue

#1369

### DIFF
--- a/.github/workflows/tag-and-release-on-push.yml
+++ b/.github/workflows/tag-and-release-on-push.yml
@@ -1,0 +1,55 @@
+name: Create Version Tag and Release
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "pyproject.toml"
+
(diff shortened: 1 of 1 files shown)


amjjbonvin (commented):
One issue here is that we might end up with many releases (one for each pull request). And as you like small pull requests, not all releases might make sense.  
There are pros and cons. But it might confuse the HPC centres when to update or not the haddock installation).

Would an option be to add a parameter to pyproject.toml that will define if a merge should lead to a new release or not?

Issue #1367, opened 2025-09-01
Provide Jupyter notebooks as tutorials


amjjbonvin:
## **Desired feature/enhancement**

provide Jupyter versions of our tutorials

## **Motivation**

- user-friendly
- can be used in training workshops

## **Description**

- provide ready-to-use Jupyter notebooks

PR #1360, opened 2025-08-31
Add notebooks

amjjbonvin:
You are about to submit a new Pull Request. Before continuing make sure you read the [contributing guidelines](CONTRIBUTING.md).

## Checklist

- [x] Tests added for the new code
- [ ] Documentation added for the code changes
- [x] Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)
- [X] `CHANGELOG.md` is updated to incorporate new changes
- [x] Haddock3 `version` has been incremented in `pyproject.toml`
- [X] Does not break licensing
- [X] Does not add any dependencies, if it does please add a thorough explanation

## Summary of the Pull Request  

Added a new directory (`notebooks`) with a first Jupyter notebook for the antibody-antigen tutorial running on Colab

## Related Issue

Closes  #1367

PR #1376, opened 2025-09-04
Adding the notebooks directory

amjjbonvin:

## Checklist

- [X] Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)
- [ ] Haddock3 `version` has been incremented in `pyproject.toml`

Added `HADDOCK3-antibody-antigen.ipynb` jupyter notebook and `README.md file`

## Related Issue

#1367


notebooks/HADDOCK3-antibody-antigen.ipynb
```diff
@@ -0,0 +1,2614 @@
...
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "#@title 1. Install haddock3, BioPython and py3Dmol\n",
+        "#@markdown Execute this cell to download the required packages.\n",
+        "!git clone https://github.com/haddocking/haddock3.git\n",
+        "import os\n",
+        "os.chdir('haddock3')\n",
+        "!pip install . --quiet\n",
```
rvhonorato: As I said in the now closed without a reason previous pr #1360:

> The latest changes have tied the execution of this notebook to a git clone which is very error prone and likely to make the notebook unstable.

and

> Fix the version here to make sure this tutorial will not break in the future

amjjbonvin:
Can only do that once we have a new pypi version.

I can already make the changes if we know which version it will be.

But better wait for now for our Tuesday meeting.

notebooks/HADDOCK3-antibody-antigen.ipynb
amjjbonvin: OK - changed it already - but we can only test live once we will make a new release for version 2025.9.3 (which should be this pull request)

notebooks/HADDOCK3-antibody-antigen.ipynb
```diff
@@ -0,0 +1,2613 @@
...
+      }
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "#@title 1. Install haddock3, BioPython and py3Dmol\n",
+        "#@markdown Execute this cell to download the required packages.\n",
+        "import os\n",
+        "os.chdir('haddock3')\n",
+        "!pip install haddock3[2025.9.3] --quiet\n",
```
rvhonorato: This is not the correct syntax and this version does not exist - https://pypi.org/project/haddock3/#history

Make sure this version exist and the syntax is correct before merging

rvhonorato (changes requested):

amjjbonvin:
What is then the correct syntax then?

And this version indeed does not exist because it will be released with the notebooks present.
Which was actually the reason I was using a `pip install .` instead.

Making an intermediate release does not make sense in this case.

rvhonorato:
This notebook depends on a specific version - the version does not exist. For this notebook to the valid the version needs to be available via pypi.

> Making an intermediate release does not make sense in this case.

"intermediate release" is a phrase i have never heard before, what you mean by it?

PR #1381, opened 2025-09-10
add `mode=grid` 

rvhonorato:

## Checklist

- [x] Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual) 
  - https://github.com/haddocking/haddock3-user-manual/pull/20

This PR adds a new `grid` execution mode. It does so by adding a new `libgrid.py`, that implements both a `GridScheduler` and a `GridJob`.  No extra dependencies are needed because this implementation expects the application to be running in a system which has DIRAC configured and its binaries available in the `PATH`.

## Related Issue

#1382

### DIFF
--- a/integration_tests/test_rigidbody.py
+++ b/integration_tests/test_rigidbody.py
@@ -136,12 +136,10 @@ def test_rigidbody_grid(rigidbody_module):
 
     for i in range(1, sampling + 1):
         assert Path(rigidbody_module.path, f"rigidbody_{i}.pdb").exists()
-        assert Path(rigidbody_module.path, f"rigidbody_{i}.out.gz").exists()
         assert Path(rigidbody_module.path, f"rigidbody_{i}.inp").exists()
         assert not Path(rigidbody_module.path, f"rigidbody_{i}.seed").exists()
 
         assert Path(rigidbody_module.path, f"rigidbody_{i}.pdb").stat().st_size > 0
-        assert Path(rigidbody_module.path, f"rigidbody_{i}.out.gz").stat().st_size > 0
         assert Path(rigidbody_module.path, f"rigidbody_{i}.inp").stat().st_size > 0
 
 
--- a/integration_tests/test_topoaa.py
+++ b/integration_tests/test_topoaa.py
@@ -74,11 +74,9 @@ def test_topoaa_module_protein_grid(topoaa_module):
     expected_inp = Path(topoaa_module.path, "e2aP_1F3G.inp")
     expected_psf = Path(topoaa_module.path, "e2aP_1F3G_haddock.psf")
     expected_pdb = Path(topoaa_module.path, "e2aP_1F3G_haddock.pdb")
-    expected_gz = Path(topoaa_module.path, "e2aP_1F3G.out.gz")
 
     assert expected_inp.exists(), f"{expected_inp} does not exist"
     assert expected_psf.exists(), f"{expected_psf} does not exist"
-    assert expected_gz.exists(), f"{expected_gz} does not exist"
     assert expected_pdb.exists(), f"{expected_pdb} does not exist"
 
     assert expected_inp.stat().st_size > 0, f"{expected_inp} is empty"
@@ -87,17 +85,14 @@ def test_topoaa_module_protein_grid(topoaa_module):
         expected_inp = Path(topoaa_module.path, f"hpr_ensemble_{i}.inp")
         expected_psf = Path(topoaa_module.path, f"hpr_ensemble_{i}_haddock.psf")
         expected_pdb = Path(topoaa_module.path, f"hpr_ensemble_{i}_haddock.pdb")
-        expected_gz = Path(topoaa_module.path, f"hpr_ensemble_{i}.out.gz")
 
         assert expected_inp.exists(), f"{expected_inp} does not exist"
         assert expected_psf.exists(), f"{expected_psf} does not exist"
         assert expected_pdb.exists(), f"{expected_pdb} does not exist"
-        assert expected_gz.exists(), f"{expected_gz} does not exist"
 
         assert expected_inp.stat().st_size > 0, f"{expected_inp} is empty"
         assert expected_psf.stat().st_size > 0, f"{expected_psf} is empty"
         assert expected_pdb.stat().st_size > 0, f"{expected_pdb} is empty"
-        assert expected_gz.stat().st_size > 0, f"{expected_gz} is empty"
 
 
 def test_topoaa_module_protein(topoaa_module):


examples/docking-protein-protein/docking-protein-protein-test-grid.toml
```diff
@@ -0,0 +1,53 @@
+# ====================================================================
+# Protein-protein docking example with NMR-derived ambiguous interaction restraints
+run_dir = "run-grid"
+mode = "grid"
+ncores = 4
```
amjjbonvin, line 5: For grid mode, you want to send a lot of jobs to the grid ideally. Or is this parameter defining the local resources used for non-grid modules only?

examples/docking-protein-protein/docking-protein-protein-test-grid.toml
```diff
@@ -0,0 +1,53 @@
...
+hisd_1 = 76
+nhise = 1
+hise_1 = 15
+
+[rigidbody]
+tolerance = 20
+ambig_fname = "data/e2a-hpr_air.tbl"
+sampling = 20
+
+["caprieval.1"]
```
VGPReys, line 27: I would put the all the caprieval modules to `[caprieval]` instead of `["caprieval.X"]`

rvhonorato: the file is `.toml` thus the proper format is indeed `["caprieval.1"]`

examples/docking-protein-protein/docking-protein-protein-test-grid.toml
rvhonorato: `ncores` is a definition of number of cores available to the haddock process

examples/docking-protein-protein/docking-protein-protein-test-grid.toml
amjjbonvin: True - but not needed in principle - we don't have this structure in any of the examples - might confuse users.

rvhonorato: I think we should instead use `.toml` :)

examples/docking-protein-protein/docking-protein-protein-test-grid.toml
amjjbonvin: But in all run modes so far (batch, local, mpi) it also define the number of concurrent CNS jobs. 

So if this also controls how many grid jobs are submitted, i would increase it as you might want to submit a large number of jobs to the grid (can be easily a few hundreds or even a few thousands)

rvhonorato: When reviewing a pull-request it's important that you check the code. You are here asking questions for things that are clearly mentioned in the code. So please have a look at the implementation and see if the explanation there is not clear.

examples/docking-protein-protein/docking-protein-protein-test-grid.toml
rvhonorato: I'm just following what is written in the docs:

> HADDOCK3 user configuration files follow TOML syntax plus additional features required for HADDOCK3.

VGPReys (commented):
Good job in making this grid submission scripts and this new mode.
I have few notes:
- what will happen when the mode grid is used on modules that are not analysis modules ?
- could we also add some flag to target specific sites (related to the green computing project) ?
- some functions / methods are missing docstrings
- the integration and uni-test for test_grid are empty

examples/docking-protein-protein/docking-protein-protein-test-grid.toml
amjjbonvin: Checking the libgrid code my comment is even more relevant. Sending only 4 jobs to the grid at the time is not using the power of the grid. I would thus suggest to increase ncores to e.g. 200 for this grid example.

rvhonorato: Sorry I don't understand your point, the executor is mapping a lambda function, what do you mean sending only 4 jobs?

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+            )
+            raise
+
+        # Add the ID to the object
+        self.id = int(result.stdout.split()[-1])
+
+        # Update the status
+        self.update_status()
+
+    def retrieve_output(self):
```
VGPReys: Missing return value
```suggestion
    def retrieve_output(self) -> None:
```

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+            dst = Path(self.wd / f"{output_f}")
+            shutil.copy(src, dst)
+
+        # The `cns.log` is the `.out` file generated by CNS
+        # Q: Is there any scenario in which this should not be compressed?
+        src = Path(f"{self.loc}/{self.id}/cns.log")
+        with open(src, "rb") as f_in, gzip.open(f"{self.cns_out_f}.gz", "wb") as f_out:
+            f_out.writelines(f_in)
+
+    def update_status(self):
```
VGPReys: Missing return value
```suggestion
    def update_status(self) -> None:
```

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+            )
+            raise
+
+        output_dict = self.parse_output(result.stdout)
+
+        self.id = output_dict["JobID"]
+        self.status = JobStatus.from_string(output_dict["Status"])
+        self.site = output_dict.get("Site", "Unknown")
+
+    def create_jdl(self):
```
VGPReys: Missing return value
```suggestion
    def create_jdl(self) -> None:
```

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+        # Unique name for the job
+        self.name = str(uuid.uuid4())
+        # Create a temporary directory for the job
+        self.loc = Path(tempfile.mkdtemp(prefix="haddock_grid_"))
+        # working directory where the GridJob was created, this is important
+        #  so we know where to put the files coming from the grid
+        self.wd = Path.cwd()
+        self.input_f = self.loc / "input.inp"
+        # `cns_out_f` is the expected `.out` file that will be generated by this execution
+        self.cns_out_f = cnsjob.output_file
```
VGPReys, line 107: I would assert that the input cnsjob is indeed a `CNSJob` class, as this grid submission is only designed for those ones.
This is made so new modules, that are not part of the analysis modules calling `get_analysis_execution_mode()`, cannot use this "grid" mode.

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+            cns_script_path=Path(cnsjob.envvars["MODULE"]),
+            toppar_path=Path(cnsjob.envvars["TOPPAR"]),
+        )
+
+        # CREATE THE JDL
+        self.create_jdl()
+
+    def prepare_payload(
+        self, cns_exec_path: Path, cns_script_path: Path, toppar_path: Path
+    ):
```
VGPReys: Missing return value
```suggestion
    ) -> None:
```

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+                # NOTE: Preserve the relative path structure from `toppar_path`!
+                #  This is important because some CNS scripts have relative paths
+                #   hardcoded in it
+                if f.is_relative_to(toppar_path):
+                    relative_path = f.relative_to(toppar_path)
+                    z.write(f, arcname=str(relative_path))
+                else:
+                    z.write(f, arcname=Path(f).name)
+
+    def create_job_script(self):
```
VGPReys: Missing return value
```suggestion
    def create_job_script(self) -> None:
```

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+        instructions = "#!/bin/bash" + os.linesep
+        instructions += "export MODULE=./" + os.linesep
+        instructions += "export TOPPAR=./" + os.linesep
+        instructions += "unzip payload.zip" + os.linesep
+        instructions += "./cns < input.inp > cns.log" + os.linesep
+
+        with open(self.job_script, "w") as f:
+            f.write(instructions)
+
+    def process_input_f(self):
```
VGPReys: Missing return value
```suggestion
    def process_input_f(self) -> None:
```

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+            log.debug(f"ID stderr: {self.stderr_f.read_text()}")
+
+        # Copy the output to the expected location
+        for output_f in self.expected_outputs:
+            src = Path(f"{self.loc}/{self.id}/{output_f}")
+            dst = Path(self.wd / f"{output_f}")
+            shutil.copy(src, dst)
+
+        # The `cns.log` is the `.out` file generated by CNS
+        # Q: Is there any scenario in which this should not be compressed?
```
VGPReys: Now, the log is now kept only if errors are found in it
Otherwise they are not written to disk to save space.

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+                else:
+                    z.write(f, arcname=Path(f).name)
+
+    def create_job_script(self):
+        """Create the job script that will be executed in the grid."""
+        instructions = "#!/bin/bash" + os.linesep
+        instructions += "export MODULE=./" + os.linesep
+        instructions += "export TOPPAR=./" + os.linesep
+        instructions += "unzip payload.zip" + os.linesep
+        instructions += "./cns < input.inp > cns.log" + os.linesep
```
VGPReys: I am not sure about the use of `os.linesep` here, as windows users may create a linesep that is not compatible with the dirac servers that are probably linux based ?

src/haddock/libs/libgrid.py
rvhonorato: Good idea - but this should not be here. There should be a new base class that all schedulers inherit and that one should have a `validate()` method - will open another issue

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,508 @@
...
+                `.inp` file and replace them with a structure that will be valid in
+                the GRID.
+
+        ===================================================================================
+        !! IMPORTANT !!
+        ===================================================================================
+        """
+
+        # Read the file first
+        with open(self.input_f, "r") as f:
```
VGPReys: Before in the run, when entering `elif isinstance(cnsjob.input_file, str)` (l.131), you write the file, to read it again, to overwrite it.
This could be optimized to reduce IO

src/haddock/libs/libgrid.py
rvhonorato: see https://github.com/haddocking/haddock3/issues/1388

src/haddock/libs/libgrid.py
rvhonorato: good catch, indeed should be `\n`

src/haddock/libs/libgrid.py
rvhonorato: Indeed here I could hold the information in a `self.input_str` instead of a `self.input_f` `Path`, good point

src/haddock/libs/libgrid.py
rvhonorato: oke clear

src/haddock/libs/libgrid.py
```diff
@@ -0,0 +1,513 @@
...
+
+        elif isinstance(cnsjob.input_file, str):
+            self.input_str = cnsjob.input_file
+
+        # CREATE JOB FILE
+        self.create_job_script()
+
+        # CREATE PAYLOAD
+        self.prepare_payload(
+            cns_exec_path=Path(cnsjob.cns_exec),
```
VGPReys, line 140: While sleeping, I just realized that this line of code can be wrong.
Here, we need to ship a specific Linux executable of CNS, and not the MacOS/Windows ones that might be installed and used by the user.

rvhonorato: 🤣 

yeah you are right, opened another issue about it: https://github.com/haddocking/haddock3/issues/1390

Issue #1385, opened 2025-09-15
implement adaptative efficiency submission algorithm for `mode=grid`


rvhonorato:
## **Desired feature/enhancement**

Add an algorithm that can optimize submission efficiency when using the grid.

## **Motivation**

The grid computing resource can absorb a very large volume of submissions and is well suited for haddock since its processes are embarrassingly parallel and for the cns-based modules the only dependency is a small binary. 

When sending payloads to the grid there is an "overhead" which is the time the submission waits in dirac's queue until it's picked up by a runner. This delay can be a few seconds or a few minutes, depending on several factors on dirac's side.

This then causes the scenario in which payloads will sit on the queue much longer than their execution time. We can define the ratio between time waiting and time running as the submission's efficiency. In a very simple example say a payload sits on the queue for 5 seconds, and then runs for 1 second, it's efficiency will be simply `running-time / total-time` = 1/5+1 = ~16%

Since we cannot control how long a payload will wait on the queue, we can increase the number of payloads to be executes in each submission - this can be done in such a way to reach a target efficiency.  So in order to achieve 90% efficiency we would submit 45 payloads as a single job (considering again a wait time of 5 seconds until the job is picked up); 45 / 5+45 = 90%.

So, running 45 payloads on 45 submissions a time would take 270 seconds, while running 45 payloads on 1 submission would take 50 seconds - which is a speedup of ~5 times.

## **Description**

Initially we do not know how long a payload takes to be executed nor what is the wait time, to get this information a small part of the workload can be used as probes. We can randomly select 5-10% or the workload and submit individual payloads to get the average running time and wait time of the grid.

Once we have these numbers, we can know how many payloads must be concatenated to achieve the target efficiency. 

Important to note that not all payloads will take exactly the same time to be executed, both because of the site they might end up in as well as peculiarities in the input data, same applies for the wait time which is a dynamic measurement. Using the average of both these should give a good enough indicator. Besides that there is also a hard-limit on the payload sizes.

## **Additional context**

This seems to be the idea behind adding the `concat` parameter for the `HPCScheduler` class:

https://github.com/haddocking/haddock3/blob/a0786274fe7908d3032cb8bf504253fcd76fa50f/src/haddock/modules/defaults.yaml#L84-L91

[...]

PR #1391, opened 2025-09-17
add adaptative job concatenation to `mode=grid`

rvhonorato:

## Checklist

DOES NOT APPLY

This PR overhauls the `libgrid`, it adds an abstract class `GridInterface` that holds the "primitive" methods to interact with dirac. It also adds `CompositeGridJob` which is a representation of a batch grid payload.

The `GridScheduler` has been updated with a feature to "probe" how busy the grid is, it will check how long a specific payload spent in waiting and running, use this to calculate what would be the ideal number of executions that should be concatenated into a single payload to increase efficiency.

With these changes the `libgrid` becomes adaptative, reacting to the load on the grid network and with this addition we should be able to efficiently process a large volume of jobs. This feature is relevant to ofer haddock3 as a service via [wenmr.science.uu.nl/haddock3](https://wenmr.science.uu.nl/haddock3)

## Related Issue

https://github.com/haddocking/haddock3/issues/1385

### DIFF
--- a/src/haddock/libs/libgrid.py
+++ b/src/haddock/libs/libgrid.py
@@ -304,7 +304,12 @@ class GridInterface(ABC):
         self.update_status()
 
     def retrieve_output(self) -> None:
-        """Retrieve the output files from DIRAC."""
+        """Retrieve the output files from DIRAC.
+
+        The `dirac-wms-job-get-output` command downloads the output sandbox, this means that
+         anything that was specified in the `OutputSandbox` section of the JDL file will be
+         put in the current working directory following the pattern: `working_dir/job_id/`
+        """
         try:
             subprocess.run(
                 ["dirac-wms-job-get-output", str(self.id)],


src/haddock/libs/libgrid.py
```diff
@@ -395,119 +437,174 @@ def __repr__(self) -> str:
...
+        #  E = N * R / W + N * R
+        #  T * (W + N * R) = N * R
+        #  T * W + T * N * R = N * R
+        #  T * W = N * R - T * N * R
+        #  T * W = N * R * (1 - T)
+        #  N = T * W / R * (1 - T)
+        batch_size = (T * W) / (R * (1 - T))
+
+        batch_size = max(1, round(batch_size))  # Ensure at least 1 job
+        return batch_size
```
rvhonorato, line 767: @VGPReys @AnnaKravchenko please double-check

VGPReys: Target efficiency T is either 100 or 1.
- If 1, I get division error
- If 100, it returns 1

Also N is not part of the equation, as E is not part of the final equation.
So having N as an argument of the input function is useless, but to print the current efficiency

src/haddock/libs/libgrid.py
```diff
@@ -395,119 +393,401 @@ def __repr__(self) -> str:
         return f"ID: {self.id} Name: {self.name} Output: {self.expected_outputs} Status: {self.status.value} Site: {self.site}"
 
 
-@dataclass
+class GridJob(GridInterface):
```
VGPReys, line 401: Is this class still relevant now that the new CompositeGridJob is there ?

src/haddock/libs/libgrid.py
```diff
@@ -395,119 +393,401 @@ def __repr__(self) -> str:
...
+        # HACK: This is not ideal, but it is a compromise between accuracy and speed.
+        subset_size = min(self.ncores, max(1, math.ceil(len(self.workload) * probing)))
+        # ===============================================================#
+
+        # Randomly select that many jobs
+        if len(self.workload) >= subset_size:
+            subset_keys = random.sample(list(self.workload), subset_size)
+            for job in self.workload:
+                if job in subset_keys:
+                    job.tag = Tag.PROBING
```
VGPReys: Usually it is better to ask for indices when using random.sample (especially true when the list is very long, which may not be the case in haddock).
This also allows then to skip the condition after, as we loop only on the selected indices
```suggestion
            subset_indices = random.sample(range(len(self.workload)), subset_size)
            for i in subset_indices:
                self.workload[i].tag = Tag.PROBING
```

src/haddock/libs/libgrid.py
```diff
@@ -395,119 +393,401 @@ def __repr__(self) -> str:
...
+            )
+            return
+
+        # Calculate average durations
+        avg_waiting = sum(waiting_durations) / len(waiting_durations)
+        avg_running = sum(running_durations) / len(running_durations)
+
+        target_efficiency = 0.9
+        batch_size = self.calculate_optimal_batch_size(
+            N=1,  # each batch had one job
```
VGPReys: What about using the actual value that was used during the probing ?
```suggestion
            N=self.batch_size,
```

src/haddock/libs/libgrid.py
```diff
@@ -395,119 +393,401 @@ def __repr__(self) -> str:
...
         if job.status == JobStatus.FAILED:
             # Jobs on the grid can fail for many reasons outside our control
             #  So if the job is failed, resubmit it up to MAX_RETRIES times
             if job.retries < MAX_RETRIES:
-                is_complete = False
                 expected_output_str = ",".join(job.expected_outputs)
                 job.retries += 1
                 log.warning(
-                    f"Job {job.name} ({expected_output_str}) failed, re-submitting - {job.retries}/{MAX_RETRIES}"
+                    f"> Job {job.name} ({expected_output_str}) failed on {job.site}, re-submitting - {job.retries}/{MAX_RETRIES} <"
```
VGPReys, line 787:
```suggestion
                    f"> Job {job.name} ({expected_output_str}) failed on "
                    f"{job.site}, re-submitting - {job.retries}/{MAX_RETRIES} <"
```

src/haddock/libs/libgrid.py
```diff
@@ -395,119 +393,401 @@ def __repr__(self) -> str:
...
+
+        if not running_durations or not waiting_durations:
+            log.warning(
+                "> Average running time is zero, cannot calculate optimal batch size <"
+            )
+            return
+
+        # Calculate average durations
+        avg_waiting = sum(waiting_durations) / len(waiting_durations)
+        avg_running = sum(running_durations) / len(running_durations)
```
VGPReys, line 723: ZeroDivisionError can occur here if the GRID has some issue and results cannot be obtained.
Maybe a dedicated Exception can be tackling this issue ?

src/haddock/libs/libgrid.py
```diff
@@ -395,119 +393,401 @@ def __repr__(self) -> str:
...
             if job.retries < MAX_RETRIES:
-                is_complete = False
                 expected_output_str = ",".join(job.expected_outputs)
                 job.retries += 1
                 log.warning(
-                    f"Job {job.name} ({expected_output_str}) failed, re-submitting - {job.retries}/{MAX_RETRIES}"
+                    f"> Job {job.name} ({expected_output_str}) failed on {job.site}, re-submitting - {job.retries}/{MAX_RETRIES} <"
                 )
+                log.debug(f"job {job.name} at {job.loc}")
                 job.submit()
```
VGPReys, line 791: what about an else statement here when we passed the number of MAX_RETRIES ?
Maybe a new status ?
Maybe set an attributes to the GridJob class and evaluate all jobs at the end of the GridScheduler ?

src/haddock/libs/libgrid.py
```diff
@@ -1,20 +1,25 @@
...
 from concurrent.futures import ThreadPoolExecutor
-from dataclasses import dataclass
+from datetime import datetime
 from enum import Enum
 from pathlib import Path
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Union
 
 from haddock import log
+from haddock.core.defaults import cns_exec as CNS_EXEC
```
VGPReys, line 20: If I am correct, this still points the user OS-specific CNS executable ?

src/haddock/libs/libgrid.py
```diff
@@ -283,67 +327,21 @@ def retrieve_output(self) -> None:
         if self.stderr_f.exists():
             dst = Path(self.wd / f"{self.id}_dirac.err")
```
VGPReys, line 333: Should this guy also be compressed when performing the final cleaning of the haddock3 run ?

src/haddock/libs/libgrid.py
```diff
@@ -283,67 +327,21 @@ def retrieve_output(self) -> None:
...
-
             log.debug(f"ID stderr: {self.stderr_f.read_text()}")
 
+        ls = glob.glob(f"{self.loc}/{self.id}/*")
+        log.debug(f"Files in the output sandbox: {ls}")
+        log.debug(f"Expected outputs: {self.expected_outputs}")
+
         # Copy the output to the expected location
         for output_f in self.expected_outputs:
             src = Path(f"{self.loc}/{self.id}/{output_f}")
```
VGPReys, line 343: I do not really understands this, because it is looping over the `cns.out` files, but not the PDB files.
When are the actual pdb outputs moves and stored into the appropriate directory ?

src/haddock/libs/libgrid.py
rvhonorato: yep

src/haddock/libs/libgrid.py
rvhonorato: Indeed, good catch

src/haddock/libs/libgrid.py
rvhonorato: Yep, that also works

src/haddock/libs/libgrid.py
rvhonorato: could be, but there is already an if gate on top of this to avoid ZeroDivision

src/haddock/libs/libgrid.py
rvhonorato: Now once it reaches the timer it will just be FAILED, but indeed another status here could be a good idea to trigger a fallback, maybe running it locally instead. Will make a note on the code for the future

src/haddock/libs/libgrid.py
rvhonorato: it does point to the OS-specific, will change it to a linux one here: https://github.com/haddocking/haddock3/pull/1398

src/haddock/libs/libgrid.py
rvhonorato: `dirac-wms-job-get-output` will download everything that was defined in the `OutputSandbox` field of the `.jdl` file inside temporary directory. 

This `expected_outputs` list is, well, the expected outputs that will be generated by the payload. They are then copied to where they should be here:

```python
        # Copy the output to the expected location
        for output_f in self.expected_outputs:
            src = Path(f"{self.loc}/{self.id}/{output_f}")
            dst = Path(self.wd / f"{output_f}")
            shutil.copy(src, dst)
```

`self.wd` is the working directory from which this class was initialized, say `rigidbody_1/`; so it will put everything there

rvhonorato: added a comment to make this more clear

src/haddock/libs/libgrid.py
rvhonorato: No idea

src/haddock/libs/libgrid.py
VGPReys: ow yes... did not see...

PR #1424, opened 2025-10-22
simplify the use of set of ambig as archives

VGPReys:

## Checklist

- [x] Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -23,11 +23,11 @@ jobs:
         with:
           stale-pr-message: "This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days."
           stale-issue-message: "This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days."
-          close-pr-message: 'This PR was closed because it has been stalled for 5 days with no activity.'
-          close-issue-message: 'This issue was closed because it has been stalled for 5 days with no activity.'
+          close-pr-message: "This PR was closed because it has been stalled for 5 days with no activity."
+          close-issue-message: "This issue was closed because it has been stalled for 5 days with no activity."
           days-before-stale: 30
           days-before-close: 5
-          exempt-issue-labels: "bug"
+          exempt-issue-labels: "bug, planned-long-term"
           exempt-pr-labels: "bug"
           remove-stale-when-updated: true
           operations-per-run: 100


src/haddock/modules/base_cns_module.py
```diff
@@ -129,26 +129,34 @@ def get_ambig_fnames(
...
+                _ambig_fnames = [Path(basepath, tbl.name) for tbl in tbl_files]
+                # Make sure they exists
+                ambig_fnames = [p for p in _ambig_fnames if p.exists()]
                 # abort execution if no files are found
                 if len(ambig_fnames) == 0:
-                    raise Exception(
-                        f"No {exp_name}*tbl files found in {exp_dir}"
-                        )
+                    raise Exception(f"No tbl files found in {ambig_fname} !")
                 self.log(f"Found {len(ambig_fnames)} compatible tbl files")
```
AnnaKravchenko: Maybe question not to you, but since we’re here - why ‘compatible’? 
This word makes me think that somehow the content of tbl files is verified or smth similar. As far as I know this is not the case, so maybe better to remove it? Just keep ```self.log(f"Found {len(ambig_fnames)} tbl files”)```

src/haddock/modules/base_cns_module.py
```diff
@@ -129,26 +129,34 @@ def get_ambig_fnames(
...
+                with tarfile.open(ambig_fname, mode="r:gz") as targz:
+                    # Get filenames
+                    tbl_files = targz.getmembers()
+                # At that stage, they should be already extracted
+                # and we simply need to generate the full paths
+                _ambig_fnames = [Path(basepath, tbl.name) for tbl in tbl_files]
+                # Make sure they exists
+                ambig_fnames = [p for p in _ambig_fnames if p.exists()]
                 # abort execution if no files are found
                 if len(ambig_fnames) == 0:
```
AnnaKravchenko, line 155: I am not sure if this is working as expected. It seems like it’s not that easy to unintentionally create an empty .tgz, so this condition is liekly will never be triggered. 

On the other hand I tried to use `ambig.tbl.tgz` which contains no “.tbl" but several “.actpass". Since there ARE files in the archive - this exception was not triggered. And also the run continued (and finished sucessfully) with no warning, no indications of “no ambiguity fiels were used”. 

May-be it’s a good idea to filter out extracted fiels and keep only ones with ".tbl” extention?

src/haddock/modules/base_cns_module.py
```diff
@@ -129,26 +129,34 @@ def get_ambig_fnames(
...
-                    raise Exception(
-                        f"No {exp_name}*tbl files found in {exp_dir}"
-                        )
+                    raise Exception(f"No tbl files found in {ambig_fname} !")
                 self.log(f"Found {len(ambig_fnames)} compatible tbl files")
-                ambig_fnames = sort_numbered_paths(*ambig_fnames)
+                # Sort them to make sure we get the same order every time
+                ambig_fnames.sort()
```
rvhonorato, line 159: Careful here, you actually need `sort_numbered_paths` instead of the built-in sort.

https://github.com/haddocking/haddock3/blob/26fd0c5bfaff9633c79ff96028995582440a6983/src/haddock/libs/libutil.py#L345-L351

Because:

```python
>>> l = ["f3", "f00", "f01", "f10", "f2"]
>>> l.sort()
>>> l
['f00', 'f01', 'f10', 'f2', 'f3']
```

src/haddock/modules/base_cns_module.py
rvhonorato: Agree, you need to do file validation on the contents coming from the tarball - you need to make sure only `.tbl` files are kept (maybe also check if the filenames are properly sanitized).

I suggest making another function `handle_tarball_restraints`

rvhonorato (changes requested):
please do not merge without tests

src/haddock/modules/base_cns_module.py
```diff
@@ -129,26 +129,34 @@ def get_ambig_fnames(
...
+                basepath = ambig_fname.parent
+                # Open archive to retrieve members
+                with tarfile.open(ambig_fname, mode="r:gz") as targz:
+                    # Get filenames
+                    tbl_files = targz.getmembers()
+                # At that stage, they should be already extracted
+                # and we simply need to generate the full paths
+                _ambig_fnames = [Path(basepath, tbl.name) for tbl in tbl_files]
+                # Make sure they exists
+                ambig_fnames = [p for p in _ambig_fnames if p.exists()]
```
VGPReys:
```suggestion
                ambig_fnames = [p for p in _ambig_fnames if p.exists() and p.suffix == ".tbl"]
```

src/haddock/modules/base_cns_module.py
VGPReys: The idea here is just to have them sorted in always the same order, and `.sort()` can do this.
Indeed, what is needed it that if this is used multiple times, the order should always be the same.
Also, the idea was to get rid of this mandatory requirements for the files to have a numbered extension at the end of the filename.
So I will keep it that way.

rvhonorato: Indeed the sort will keep it constant, and yes it would be better to not have this numbering requirement - but isn't this numbering linked to something else?

src/haddock/modules/base_cns_module.py
VGPReys: Here, I check that they are indeed uncompressed and the generated paths exists

src/haddock/modules/base_cns_module.py
VGPReys: No, it was just to make it over complicated.
Actually the method `sort_numbered_paths` was only used to sort the archives containing restrain files.

rvhonorato: oke clear, so if not needed better to just use `.sort` instead and remove this `sort_numbered_paths` to make it simpler

VGPReys: said something wrong, `sort_numbered_paths` is used at some other places.
So I will leave it in the code

AnnaKravchenko:
Aaaand 🥁 
The answer is yes, it is a problem. 
The run is no good if ones compress the tbls on mac. I see classical “absolutely no restraints” kind of models - molecules far apart for each other.

src/haddock/modules/base_cns_module.py
```diff
@@ -129,26 +129,37 @@ def get_ambig_fnames(
...
+                    # Get filenames
+                    tbl_files = targz.getmembers()
+                # At that stage, they should be already extracted
+                # and we simply need to generate the full paths
+                _ambig_fnames = [Path(basepath, tbl.name) for tbl in tbl_files]
+                # Make sure they exists
+                ambig_fnames = [
+                    p for p in _ambig_fnames
+                    if p.exists() and p.suffix == ".tbl"
+                    ]
```
VGPReys:
```suggestion
                ambig_fnames = []
                for tbl in tbl_files:
                    fname = tbl.name
                    # Filter non-".tbl" and hidden files
                    if any([fname.startswith("."), fname.suffix != ".tbl"]):
                        continue
                    # Build path
                    tbl_fpath = Path(basepath, fname)
                    # Make sure it exists
                    if tbl_fpath.exists():
                        ambig_fnames.append(tbl_fpath)
```

src/haddock/modules/base_cns_module.py
```diff
@@ -129,26 +129,42 @@ def get_ambig_fnames(
...
+                with tarfile.open(ambig_fname, mode="r:gz") as targz:
+                    # Get filenames
+                    tbl_files = targz.getmembers()
+                # At that stage, they should be already extracted
+                # and we simply need to generate the full paths
+                ambig_fnames = []
+                for tbl in tbl_files:
+                    fname = tbl.name
+                    # Filter non-".tbl" and hidden files
+                    if any([fname.startswith("."), fname.suffix != ".tbl"]):
```
VGPReys:
```suggestion
                    if any([fname.startswith("."), tbl.suffix != ".tbl"]):
```

src/haddock/modules/base_cns_module.py
```diff
@@ -129,26 +129,42 @@ def get_ambig_fnames(
...
+                ambig_fnames = []
+                for tbl in tbl_files:
+                    fname = tbl.name
+                    # Filter non-".tbl" and hidden files
+                    if any([fname.startswith("."), tbl.suffix != ".tbl"]):
+                        continue
+                    # Build path
+                    tbl_fpath = Path(basepath, fname)
+                    # Make sure it exists
+                    if tbl_fpath.exists():
```
VGPReys:
```suggestion
                    if fname.startswith("."):
                        continue
                    # Build path
                    tbl_fpath = Path(basepath, fname)
                    # Make sure it exists
                    if tbl_fpath.exists() and tbl_fpath.suffix == ".tbl":
```

src/haddock/modules/base_cns_module.py
```diff
@@ -129,26 +129,42 @@ def get_ambig_fnames(
...
+                    # Make sure it exists
+                    if tbl_fpath.exists() and tbl_fpath.suffix == ".tbl":
+                        ambig_fnames.append(tbl_fpath)
                 # abort execution if no files are found
                 if len(ambig_fnames) == 0:
-                    raise Exception(
-                        f"No {exp_name}*tbl files found in {exp_dir}"
-                        )
+                    raise Exception(f"No tbl files found in {ambig_fname} !")
                 self.log(f"Found {len(ambig_fnames)} compatible tbl files")
```
VGPReys:
```suggestion
                self.log(f"Found {len(ambig_fnames)} tbl files")
```

Issue #1425, opened 2025-10-23
refactor `libpdb.sanitize` to avoid extending global `_to_keep`


rvhonorato:
Some time ago I added `custom_topology` support for `libpdb.sanitize` - https://github.com/haddocking/haddock3/commit/5dbeca5ca16fa9751c0093cd0772f71f94e2ef9e 

The way it was implemented was by extending the global `_to_keep`:

https://github.com/haddocking/haddock3/blob/daffcaede7204cbc362e2ca8e43b81133d376d9e/src/haddock/libs/libpdb.py#L112-L113

https://github.com/haddocking/haddock3/blob/daffcaede7204cbc362e2ca8e43b81133d376d9e/src/haddock/libs/libpdb.py#L172-L174

While this is not an issue for normal execution, it introduces an edge case in testing in which the order of the tests can affect this global variable and thus cause tests to fail.

Say we have a Test A which uses this list and Test B which has a side-effect of extending this global list:

```
Case 1:
  Test A: Pass
  Test B: Pass (extend the global)
=== Pass

Case 2:
   Test B: Pass (extend the global)
   Test A: Fail
=== Fail
```

Found it when debugging #1412 which was showing flaky tests and luckily this was identified because we added random order to the tests (https://github.com/haddocking/haddock3/commit/f55ce6ef675f033a619569e0a6dd110dcda2efa8)

PR #1426, opened 2025-10-23
refactor `libpdb.sanitize`

rvhonorato:

## Checklist

This PR refactors `libpdb.sanitize` to not edit the global variable `_to_keep` in order to avoid testing artefacts

## Related Issue

#1425

### DIFF
--- a/src/haddock/libs/libpdb.py
+++ b/src/haddock/libs/libpdb.py
@@ -1,6 +1,7 @@
 """Parse molecular structures in PDB format."""
 
 import os
+from copy import deepcopy
 from functools import partial
 from pathlib import Path
 
@@ -20,9 +21,8 @@ from haddock.core.typing import (
     TypeVar,
     Union,
 )
-from haddock.libs.libio import working_directory, PDBFile
+from haddock.libs.libio import PDBFile, working_directory
 from haddock.libs.libutil import get_result_or_same_in_list, sort_numbered_paths
-
 
 slc_record = slice(0, 6)
 slc_serial = slice(6, 11)
@@ -169,7 +169,7 @@ def sanitize(
     custom_topology: Optional[FilePath] = None,
 ) -> Union[FilePathT, Path]:
     """Sanitize a PDB file."""
-    to_keep = _to_keep.copy()
+    to_keep = deepcopy(_to_keep)
     if custom_topology:
         custom_res_to_keep = get_supported_residues(custom_topology)
         to_keep.extend(custom_res_to_keep)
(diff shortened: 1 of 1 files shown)


src/haddock/libs/libpdb.py
```diff
@@ -165,13 +164,15 @@ def swap_segid_chain(pdb_file_path: FilePath,
...
 def sanitize(
-        pdb_file_path: FilePathT,
-        overwrite: bool = True,
-        custom_topology: Optional[FilePath] = None) -> Union[FilePathT, Path]:
+    pdb_file_path: FilePathT,
+    overwrite: bool = True,
+    custom_topology: Optional[FilePath] = None,
+) -> Union[FilePathT, Path]:
     """Sanitize a PDB file."""
+    to_keep = _to_keep.copy()
```
VGPReys: regarding shallow and deep copies, I think this one needs to be a deep copy of the list (https://stackoverflow.com/questions/43983907/python-list-copy-shallow-vs-deep-copy).
Also requires the import of copy.deepcopy (`from copy import deepcopy`)
```suggestion
    to_keep = deepcopy(_to_keep)
```

src/haddock/libs/libpdb.py
```diff
@@ -299,19 +299,22 @@ def read_RECORD_section(
...
+        if line.startswith(("ATOM", "HETATM"))
+        and (the_line := line[section_slice].strip())  # noqa: E501
+    )
     return chainids
 
 
-read_chainids = partial(read_RECORD_section, section_slice=slc_chainid, func=list)  # noqa: E501
+read_chainids = partial(
+    read_RECORD_section, section_slice=slc_chainid, func=list
+)  # noqa: E501
```
VGPReys:
```suggestion
)
```

src/haddock/libs/libpdb.py
```diff
@@ -427,23 +430,33 @@ def write_previous_residue(
...
-        "C", "N",  # for peptide bonds
-        "O3'", "O5'",  # for DNA/RNA
-        )
-    current_resid: tuple[str, str] = ("-", "-", )
+        "C",
+        "N",  # for peptide bonds
+        "O3'",
+        "O5'",  # for DNA/RNA
+    )
+    current_resid: tuple[str, str] = (
```
VGPReys: In my opinion, previous (`    current_resid: tuple[str, str] = ("-", "-", )`) was better and still fitting under 80 chars

src/haddock/libs/libpdb.py
rvhonorato: here i don't think it matters, but also doesn't hurt to make it a deepcopy

src/haddock/libs/libpdb.py
rvhonorato: it goes over multiple lines if there are trailing commas, but here we can also remove the trailing comma

PR #1439, opened 2025-11-03
update CNS download links

rvhonorato:
## Checklist
DOES NOT APPLY

This PR updates the links to the pre-compiled CNS binaries

## Related Issue

DOES NOT APPLY

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,8 @@ import platform
 import subprocess
 import sys
 import tarfile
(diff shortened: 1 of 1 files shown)


amjjbonvin (approved):
Thanks for updating things - looking good - we should follow the same recipe to update all other links in our tutorials. We will also need to update the notebook inside the haddock3 repo.

Issue #1475, opened 2026-02-24
automated toppar generation for unknown ligands


rvhonorato:
## **Desired feature/enhancement**

automated toppar generation for ligands

## **Motivation**

Currently the code expects the user to input an externally obtained `.top`/`.param` file that represents a given small molecule. This is good because it gives the user the option to input custom files specific to their case but it also adds an extra step needed before running our application. We could instead automatically generate the topologies and parameters when user does not provide them.

## **Description**

One way to do it is to extend `topoaa`:

https://github.com/haddocking/haddock3/blob/c173db03f46b044107b5d0a22708234c3f5bac9f/src/haddock/modules/topology/topoaa/__init__.py#L260-L270

and in the `else` branch

```diff
else:
+ top, par: list[str]: libligand.identify_hetatms()
+ if top:
+    libpdb.sanitize(model, overwrite=True, custom_topology=top)
+    break
    libpdb.sanitize(model, overwrite=True)
```
This can be easily done using [haddocking/prodrg](https://github.com/haddocking/prodrg) and putting the functions in a new `libligand`
 
## **Additional context**

I started working on adding this functionality to [wenmr.science.uu.nl/haddock3](https://wenmr.science.uu.nl/haddock3) but it's better if the source code can handle this by itself without relying on an external feature, so it makes more sense for this to exist inside haddock3

PR #1476, opened 2026-02-24
add automated toppar generation for unknown small molecules

rvhonorato:

## Checklist

- [x] Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual) - https://github.com/haddocking/haddock3-user-manual/pull/27

This PR adds a new `libligand` with the purpose of handling small molecules, in this case specifically the creation `.top` and `.param` files using PRODRG.

Currently when the user does not input `ligand_param/top_fname` and the input contains any unknown hetatms, there are just deleted during sanitization. I am not sure if this is the desired behaviour or not, but assuming it is - I added a new boolean parameter called `autotoppar` to ensure this pathway of removing unknown atoms can be preserved.

So to trigger the automated generation, the user needs to set `autotoppar=true` and if this parameter is defined together with `ligand_param/top_fname`, `ligand_param/top_fname` takes priority.

The automated topology/param generation of small molecules is done via `prodrg` - shipped as a binary with the code after this PR. I also added two sanitization methods to remove `NBONds` from the parameters generated by prodrg as they can interfer with haddock ones, and also found that prodrg will sometimes add a `:` to the topology which breaks CNS syntax - so there is another method to clean it.

And needles to say this only applies to things that are not known by haddock, which is defined in `haddock.core.supported_molecules`.

When we automate the toppar generation, what means is that we are effectively setting the `ligand_param/top_fname` for the user. This is easily done in the module that generates it, but this value needs to be propagated so that modules downstream on the workflow can also access it. To deal with this behaviour I added a new logic in `libworkflow` to handle this propagation.

## Related Issue

#1475

## Additional Info

We currently only have prodrg binaries for `x86_64-linux` and `arm64-darwin` - when we have others they should also be added here, but we need to handle this elsewhere, please look here: https://github.com/haddocking/prodrg

`run_prodrg` will fail graciously if `autotoppar=true` in an unsupported architecture.

### DIFF
--- a/src/haddock/gear/greetings.py
+++ b/src/haddock/gear/greetings.py
@@ -29,6 +29,7 @@ international_good_byes = [
     "Do widzenia",
     "Mandi",
     "Αντίο",
+    "Auf Wiedersehen",
 ]
 
 # List of urls to be printed to the screen at the end of a workflow


amjjbonvin (dismissed):
Very nice! Works perfectly on my side

Would it be possible to also compile the prodrg exec for aarch64-linux?

amjjbonvin (approved):
A very nice addition! 
Works fine on my side - will should do some benchmarking on a protein-ligand dataset to check the performance (the shape docking set is a good one for this).

PR #1480, opened 2026-02-26
add `deeprank` module

### LINKED ISSUE #569: develop a deeprank analysis module
The idea would be to develop a deeprank cnn/gnn scoring module inside haddock3. It will be part of the analysis modules, as typically scoring modules are those that modify the coordinates.

rvhonorato:

This PR adds `deeprank-gnn-esm` as a scoring module

## How was this tested?

I added unit, integration and end-to-end tests. Also added one scoring and one docking workflow, all executed fine.

## AI assistance

None

## Checklist

- ~[ ] Documentation updated if needed (also in the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)~ 
   - to be done once a relevant example has been added

## Related issues

#569

## Notes for reviewers

There is definitively room for improvement here in terms of optimization of the wrapper  (currently it triggers one execution per model, maybe there is a better of doing it) and in a meaningful example. This PR here is "just" adding the module, on further PRs you may improve on the execution and try out better examples. 

As it is now the module is properly implemented and working. Deeprank carries very heavy `pytorch` dependencies, and because of that I've set it as an optional dependency but it could also be part of the main ones.


rvhonorato:
@ntxxt I am not exactly sure if you were planning to add `deeprank-gnn-esm` as a scoring module or something else but anyway I went ahead and got started on it as a scoring module.

I have added the skeleton for the module as well as the unittests, integration tests and necessary updates to the project installation and updates to the continuous integration pipeline.

One thing to note here is that haddock supports python versions 3.9 to 3.14 but `deeprank-gnn-esm`does not support 3.14.

The skeleton should be in place but there are still many things to be done I'll leave it to you to finish it - let me know if there's anything I can help with

rvhonorato:
small update here to the minimum deeprank version and to the ci - now at `v1.1.0` that supports both mac and python3.14, see https://github.com/haddocking/deeprank-gnn-esm/pull/32 and https://github.com/haddocking/deeprank-gnn-esm/pull/33

rvhonorato:
I see there has been no activity here for 3 months and I'd like to close at least the basic implementation of the module and then later you or someone else can work on the fine details.

@ntxxt can I take over an finalize this?

rvhonorato:
@copilot resolve the merge conflicts in this pull request

amjjbonvin (commented):
A functional module - nice! 

A few comments / observations from my testing:

1) Currently `seletop` will by default select as best models the ones with the smallest score (whatever the score is). We need changes here so that it reverts its sorting if deeprank output is used

2) The module is very verbose - would limit the output based a verbosity parameter

3) It currently runs sequentially for all models, i.e. it can not process in parallel several models. Or are ncores used for the inference of one model?

4) Embeddings are generated for each model, which is not efficient in case all models have the same sequence. Can we add an option like `uniform_model=true` to avoid generating embedding for each model?

5) The modules generates a lot of files. In principle we only want the predicted Fnat. Can we either clean the resulting directories (one dir per model with multiple files in it) or keep that in memory?  Those file could for example only be kept if `debug=true`

ntxxt:
In terms of running effiency, A lot of things could be optimized. 

Both 3, 4,5 are prety valid for me. Especially 3 and 4 are already handled by deeprank-ab. I can work on that afterwards but it would be starting from next week.

rvhonorato:
>     1. Currently `seletop` will by default select as best models the ones with the smallest score (whatever the score is). We need changes here so that it reverts its sorting if deeprank output is used

done
 
>     2. The module is very verbose - would limit the output based a verbosity parameter

done

>     3. It currently runs sequentially for all models, i.e. it can not process in parallel several models. Or are ncores used for the inference of one model?

looked into its code, deeprank is able to handle an ensemble, and if so it optimizes the execution with paralelism - updated it here


>     4. Embeddings are generated for each model, which is not efficient in case all models have the same sequence. Can we add an option like `uniform_model=true` to avoid generating embedding for each model?

same as above, giving the input as an ensemble otpimizes this

>     5. The modules generates a lot of files. In principle we only want the predicted Fnat. Can we either clean the resulting directories (one dir per model with multiple files in it) or keep that in memory?  Those file could for example only be kept if `debug=true`

done

src/haddock/modules/scoring/deeprank/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+            ncores=ncores,
+            chain_i=self.params["chain_i"],
+            chain_j=self.params["chain_j"],
+        )
+
+        # `run()` executes entirely inside a temporary directory and returns
+        # only the scores - none of deeprank's intermediate files land here
+        result_dic: dict[str, float] = deeprank_wrapper.run()
+
+        # Deeprank predicts fnat (0 to 1, higher is better), but the rest of
```
amjjbonvin, line 77: A simple solution indeed.

I do have a fix for seletop to allow for reverse sorting. Also some changes are required in caprieval.
I will open an issue for this and create a separate pull request.

For now we can keep this negative score, but somewhere it must be made clear that the deeprank scores in haddock3 as used as -Fnat instead of Fnat

amjjbonvin (commented):
The use of the ensemble option is nice, but we do need also an individual option (default can be ensemble). This is required for example in a CAPRI scoring scenario in which the models might have different sequences (or sequences length).

rvhonorato:
grouping the model as an ensemble is a hacky solution here to trigger deeprank's internal handling of multiple models.

if you need it to handle heterogeneous input, I'd say this is a deeprank issue. it should be updated in to, for example, be able to efficiently handle a directory full of PDB files and internally handle the embedding and such. 

this is out of scope, this PR is adding deeprank in its current state to the haddock3 code and that is done. 

if you do not have any comments on the current code I'd say this is ready to be merged.

amjjbonvin:
I will then open an issue about this.

The first behaviour was to handle each PDB separately, which would be fine for scoring heterogeneous models.

The ensemble option is more efficient.

So being able to switch between those two modes depending on what we are scoring would be handy.

But this will be for another PR.

amjjbonvin:
It is still very verbose. Running currently a full run. Getting 1000 times: `INFO - Renumbering PDB file`  and then it will list the score foreach model (another 1000 lines), which is not really required - all the info is in the generated tsv file.

amjjbonvin (approved):

Issue #1499, opened 2026-04-07
excessive memory usage in `contactmap`


rvhonorato:
## **Module/Workflow/Library affected**

`contactmap`

## **Expected behavior**

module should be able to handle inputs of any size without going out-of-memory

## **Actual behavior**

it currently scales quadratically, so depending on the input size the process will run out of memory and crash

## **Steps to reproduce the behavior**

```python
import tracemalloc
import numpy as np
from haddock.modules.analysis.contactmap.contmap import compute_distance_matrix

n_atoms = 36000
all_coords = np.random.rand(n_atoms, 3).tolist()
tracemalloc.start()

try:
    dist_matrix = compute_distance_matrix(all_coords)
    current_mem, peak_mem = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    actual_mb = peak_mem / (1024**2)
    actual_gb = actual_mb / 1024
    print(f"{actual_gb:.1f} GB")
    del dist_matrix

except MemoryError:
    tracemalloc.stop()
    print("OOM")
```

## **Suggestions on how to fix it**

The problem comes from
https://github.com/haddocking/haddock3/blob/8072d15293d464ab5b99eb2cce392dc04b821cf9/src/haddock/modules/analysis/contactmap/contmap.py#L805

which is creating a NxN matrix, so for 100 atoms, the matrix will have 10.000 elements - for 36000 atoms (which should be about less than 1Mb), it will have 1296000000 elements and this will take more than 10Gb of memory

## **Version**

`8072d15293d464ab5b99eb2cce392dc04b821cf9` 

## **Additional context**

This is currently limiting how we can offer haddock3 as a service in wenmr.science.uu.nl

rvhonorato:
did a workaround in #1500 since refactoring the logic would mean re-writing the whole module

PR #1500, opened 2026-04-07
check memory requirement in `contactmap` 

rvhonorato:

## Checklist

This PR adds a logic to check if the host machine has the necessary memory requirements to run `contactmap` and if it does not it skips the execution of the module.

 ⚠️ IMPORTANT ⚠️

To get the available memory of the system we are now using `psutil`, this means this is now a dependency of the project! https://pypi.org/project/psutil/ is a very stable and mature package and is the best way for us to get the available memory in different systems.

## Related Issue

#1499

## Additional Info

for prosperity: I tried to bypass calculating the matrix and adding a generation with on-the-fly calculation of the contacts, as this would avoid keeping things in memory. However changing this also needs a significant refactoring - basically a whole re-write of this module.


integration_tests/test_contactmap.py
```diff
@@ -49,22 +55,27 @@ def retrieve_models(individualize: bool = False):
...
     # check outputs
     output_bp = contactmap.path
     # clt_id == None
-    clustNone_tsv_fpath = f"{output_bp}/Unclustered_contactmap_rigidbody_3_cltid_None_contacts.tsv"  # noqa : E501
-    clustNone_html_fpath = f"{output_bp}/Unclustered_contactmap_rigidbody_3_cltid_None_heatmap.html"  # noqa : E501
+    clustNone_tsv_fpath = (
+        f"{output_bp}/Unclustered_contactmap_rigidbody_3_cltid_None_contacts.tsv"  # noqa : E501
+    )
+    clustNone_html_fpath = (
+        f"{output_bp}/Unclustered_contactmap_rigidbody_3_cltid_None_heatmap.html"  # noqa : E501
```
VGPReys: noaq : E501 can be removed

integration_tests/test_contactmap.py
```diff
@@ -81,3 +92,31 @@ def test_contactmap_example(contactmap, mocker):
...
+    )
+    monkeypatch.chdir(contactmap.path)
+
+    # Run the module - should skip execution due to low memory
+    contactmap.run()
+
+    # Check that the directory is empty
+    ls = list(glob.glob(f"{contactmap.path}/*"))
+
+    print(ls)
```
VGPReys: print statement can be removed

integration_tests/test_contactmap.py
```diff
@@ -81,3 +92,31 @@ def test_contactmap_example(contactmap, mocker):
...
+    assert len(list(glob.glob(f"{contactmap.path}/*"))) == 0
```
VGPReys:
```suggestion
    assert len(ls) == 0
```

src/haddock/modules/analysis/contactmap/helper.py
```diff
@@ -0,0 +1,54 @@
...
+def get_available_memory() -> float:
+    """
+    Get the available system memory in GB.
+
+    Returns
+    -------
+    float
+        Available memory in gigabytes (GB).
+    """
+    return psutil.virtual_memory().available / (1024**3)
```
VGPReys: I would put this function as well as the import in the `libs/libutil.py`

VGPReys: Small comment to help understand the 1024 ** 3 
```suggestion
    """
    # Get the available virtual memory in Bytes
    # then divide it by (1024 ^ 3) to get this value in GigaBytes
    return psutil.virtual_memory().available / (1024**3)
```

src/haddock/modules/analysis/contactmap/__init__.py
```diff
@@ -61,11 +65,35 @@ def _run(self) -> None:
...
+        # requirement will increase quadratically and can fail with an out-of-memory
+        # error. Changing this behaviour would require a total re-write of the module
+        # as of 04-2026 so instead we apply the following workaround:
+        #  - Check what is the total size of the models (size is faster than reading)
+        #  - Guesstimate how many atoms in total it would have based on the size
+        #  - Calculate the expected matrix size and its memory requirements
+        #  - Get how much memory the current host system has
+        #  - If the system has less memory than needed, fail graciously
+        current_memory = get_available_memory()
+        needed_memory = get_necessary_memory(models)
```
VGPReys: I think here there is an issue.
In my opinion, this value should be multiplied by the number of cores,
as each core will require that particular amount of memory
```suggestion
        needed_memory = get_necessary_memory(models) * self.params["ncores"]
```

src/haddock/modules/analysis/contactmap/__init__.py
```diff
@@ -61,11 +65,35 @@ def _run(self) -> None:
...
+        if current_memory < needed_memory:
+            self.log(
+                msg=f"Not enough memory to execute `contactmap`, needs {needed_memory:.2f}Gb has {current_memory:.2f}Gb",
```
VGPReys:
```suggestion
                msg=(
                    f"Not enough memory to execute `contactmap` "
                    f"(needs {needed_memory:.2f}Gb has {current_memory:.2f}Gb). "
                    "! Skipping this module !"
                    ),
```

src/haddock/modules/analysis/contactmap/helper.py
```diff
@@ -0,0 +1,54 @@
...
+    Returns
+    -------
+    float
+        Estimated memory requirement in gigabytes (GB).
+    """
+    atoms = 0
+    for model in models:
+        try:
+            # use `getsize` because its faster than reading the whole PDB
+            file_size = os.path.getsize(model.file_name)
```
VGPReys: smart 👍

src/haddock/modules/analysis/contactmap/helper.py
```diff
@@ -0,0 +1,54 @@
...
+    models : list
+        List of model objects with file_name attribute.
+
+    Returns
+    -------
+    float
+        Estimated memory requirement in gigabytes (GB).
+    """
+    atoms = 0
+    for model in models:
```
VGPReys: Do we really need to loop over all models ?
Or can we approximate that most of the models will be similar and therefore skip this entire loop as soon as we have the first `estimated_atoms` value ?

src/haddock/modules/analysis/contactmap/helper.py
```diff
@@ -0,0 +1,54 @@
...
+        except Exception:
+            # skip this model
+            continue
+
+    if atoms == 0:
+        # Fall back to a safe default if we can't estimate
+        atoms = 10000
+
+    matrix_size_bytes = (atoms * atoms) * 8
+    matrix_size_gb = matrix_size_bytes / (1024**3)
```
VGPReys:
```suggestion
    # Compute an approximation of the matrix size
    matrix_size_bytes = (atoms * atoms) * 8
    # Convert it into GigaBytes
    matrix_size_gb = matrix_size_bytes / (1024**3)
```

src/haddock/modules/analysis/contactmap/__init__.py
rvhonorato: yep you are right, indeed it needs to be multiplied

src/haddock/modules/analysis/contactmap/helper.py
rvhonorato: good point, it's already doing some approximations so it makes sense to just use the first model

VGPReys (approved):
I guess we are good 🚀

PR #1514, opened 2026-04-13
add `toppar` property to PDB object

### LINKED ISSUE #1507: Wrong pairing of topo/param and PDB models in autotoppar mode with an ensemble of models

any CNS module using the ligand param/topo file.

## **Expected behavior**

When an ensemble of ligand is provided, PRODRG parameters are generated for each one, and should be passed correctly to the next modules making sure that the ligand param/topo files are matching the PDB model number.

## **Actual behavior**

Currently the param/topo files might not match the model number.
This is not an issue if all param/topo files are similar for all models of an ensemble, but it becomes an issue if the models differ.

1. Go to the docking-protein-ligand example directory
2. Go to the data directory
3. Generate an ensemble of conformations (e.g. 11) with e.g.:
```
pdb_mkensemble oseltamivir_zwitterion.pdb oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb\
                                  oseltamivir_zwitterion.pdb >oseltamivir_ensemble.pdb
```

4. Edit the `docking-protein-ligand-autotoppar-test.cfg` file, set `debug=true`, and replace the ligand (second molecule by the ensemble
5. Run haddock3: haddock3 docking-protein-ligand-autotoppar-test.cfg
6. Check the `rigidbody_1.inp` file in the `01_rigidbody` directory and search for "oselta"

You will find one instance with the parameter file and one with the PDB file name and the number in these filenames should match.

I.e. model 1 should be associated with the parameter file 1 and so on

Instead I now have:

```
eval ($ligand_param_fname="/Users/abonvin/haddock_git/haddock3/examples/docking-protein-ligand/run1-autotoppar-ensemble-test/00_topoaa/oseltamivir_ensemble_11_prodrg.param")
eval ($ligand_top_fname="/Users/abonvin/haddock_git/haddock3/examples/docking-protein-ligand/run1-autotoppar-ensemble-test/00_topoaa/oseltamivir_ensemble_11_prodrg.top")
```

and for the PDB filename:

```
eval ($input_pdb_filename_2="../00_topoaa/oseltamivir_zwitterion_from_oseltamivir_ensemble_1_haddock.pdb")
```

I.e. the topo/param files of model 11 are now associated with model 1.

I.e. a mismatch between the PDB and the parameter files. And some PRODRG-generated parameters are different for different conformations.

## **Version**

0531f00e5cad65887bfc5b9e844d7de16f905a3f

rvhonorato:

## Checklist

This PR adds a new capability of defining custom toppar per module instead of globally, this increases the granularity and is relevant when processing an ensemble of heterogeneous ligands.

To achieve this I updated the pdb ontology to contain the ligand toppar as a property, so this can be carried forward in the modules.

I also added backwards compatibility, so its still possible to define a global custom toppar fname, but when using `autoppar=true` and an ensemble of ligands each will inherit the correct toppar generated by prodrg.

## Related Issue

Closes #1507 by adding support to multiple custom toppar


src/haddock/libs/libcns.py
```diff
@@ -346,8 +339,19 @@ def prepare_cns_input(
...
+    # Check if any PDBFile has ligand files and override global parameters
+    # This is important for ensembles with autotoppar where each model has different ligand files
+    pdb_files = transform_to_list(input_element)
+    for pdb in pdb_files:
+        if hasattr(pdb, "ligand_top_fname") and pdb.ligand_top_fname:
+            default_params += write_eval_line("ligand_top_fname", pdb.ligand_top_fname)
+        if hasattr(pdb, "ligand_param_fname") and pdb.ligand_param_fname:
+            default_params += write_eval_line(
+                "ligand_param_fname", pdb.ligand_param_fname
+            )
```
VGPReys, line 351: As multiple pdb files may have their own `ligand_top_fname` and `ligand_param_fname`, this will be written **multiple times** in the CNS script, and **potentially overwrite** the variable holding the definition of previous ones ?

amjjbonvin:
> As multiple pdb files may have their own ligand_top_fname and ligand_param_fname, this will be written multiple times in the CNS script, and potentially overwrite the variable holding the definition of previous ones ?
> 

This should not happen if there is only one ligand in principle, right? Each model will only inherit one set of param/top files.

src/haddock/libs/libcns.py
rvhonorato: indeed, this would be the case with two ligands
```
Model 1
  A - protein
  B - ligand > custom toppar ❌
  C - ligand > custom toppar ✅
```

but I do not think there is support for multiple custom toppar in the same model, is it?

this PR implements custom toppar in different models:

```
Model 1
 A - protein
 B - ligand > custom toppar 01

Model 2
 A - protein
 B - ligand > custom toppar 02
```

amjjbonvin:
The server does, simply appending the param/topo files into one.
Works provided PRODRG uses different letters for the atom types.

> but I do not think there is support for multiple custom toppar in the same model, is it?
>

rvhonorato:
given it adds a bit of complexity to check the atoms types - it can be done in another PR to keep this one small.

VGPReys (approved):
OK good, functional !

src/haddock/libs/libcns.py
VGPReys: This issue will be addressed in an other PR then

PR #1515, opened 2026-04-13
Fixing shape support in cgtoaa module

amjjbonvin:
## Checklist

Modified `cgtoaa.cns` to correctly handle a shape if present (fixing it and not including it in the morphing of the AA models onto the CG model of the complex.

Added a docking example illustrating protein-protein docking into a shape (both AA and CG)

TODO: update user manual with the information about the new docking example


examples/docking-protein-protein-shape/docking-protein-protein-shape-CG-full.cfg
```diff
@@ -0,0 +1,67 @@
...
+# ====================================================================
+[topoaa]
+
+[topocg]
+
+[rigidbody]
+tolerance = 20
+ambig_fname = "data/ambig-shape.tbl"
+mol_shape_3 = true
+
```
VGPReys, line 30: Don't we need to fix the origin of the shape ?
```suggestion
mol_shape_3 = true
mol_fix_origin_3 = true

```

examples/docking-protein-protein-shape/docking-protein-protein-shape-CG-test.cfg
```diff
@@ -0,0 +1,69 @@
...
+[topoaa]
+
+[topocg]
+
+[rigidbody]
+tolerance = 20
+ambig_fname = "data/ambig-shape.tbl"
+sampling = 5
+mol_shape_3 = true
+
```
VGPReys, line 32:
```suggestion
mol_shape_3 = true
mol_fix_origin_3 = true

```

examples/docking-protein-protein-shape/docking-protein-protein-shape-full.cfg
```diff
@@ -0,0 +1,59 @@
...
+# ====================================================================
+# Parameters for each stage are defined below, prefer full paths
+# ====================================================================
+[topoaa]
+
+[rigidbody]
+tolerance = 20
+ambig_fname = "data/ambig-shape.tbl"
+mol_shape_3 = true
+
```
VGPReys, line 28:
```suggestion
mol_fix_origin_3 = true

```

examples/docking-protein-protein-shape/docking-protein-protein-shape-test.cfg
```diff
@@ -0,0 +1,61 @@
...
+# Parameters for each stage are defined below, prefer full paths
+# ====================================================================
+[topoaa]
+
+[rigidbody]
+tolerance = 20
+ambig_fname = "data/ambig-shape.tbl"
+sampling = 5
+mol_shape_3 = true
+
```
VGPReys, line 29:
```suggestion
mol_shape_3 = true
mol_fix_origin_3 = true
```

src/haddock/libs/libaa2cg.py
VGPReys, line 843: missing docstring of the function

VGPReys: missing types

src/haddock/libs/libaa2cg.py
```diff
@@ -850,6 +850,7 @@ def martinize(input_pdb, output_path, skipss):
     Returns:
         cg_pdb_name: str
     """
+    shape = False
```
VGPReys:
```suggestion
    shape: bool = False
```

src/haddock/libs/libaa2cg.py
```diff
@@ -850,6 +850,7 @@ def martinize(input_pdb, output_path, skipss):
     Returns:
         cg_pdb_name: str
```
VGPReys: missing definition

VGPReys:
```suggestion
   Returns:
       cg_pdb_name: str
       shape: bool
           Boolean flag defining if the structure contains a shape.
```

src/haddock/libs/libaa2cg.py
VGPReys:
```suggestion
    cg_pdb_name = f"../{output_path}/{Path(pdbf_path).stem}_cg.pdb"
```

VGPReys:
```suggestion
        if line.startswith("ATOM"):
```

VGPReys:
```suggestion
        if line.startswith("ATOM"):
```

VGPReys:
```suggestion
                if 1 <= len(atom_name) <= 3:
                    n_l = f"{line[:12]} {atom_name:<3s}{line[16:]}"
```

src/haddock/libs/libaa2cg.py
```diff
@@ -906,6 +907,8 @@ def martinize(input_pdb, output_path, skipss):
             for residue in mapping_dic:
                 if residue.id[0] != " ":  # filter HETATMS
                     continue
+                if residue.resname == "SHA":
+                    shape = True
```
VGPReys: From this point, do we really need to go through the entire `martinize()` ? or can we simply return that it is a shape ? to continue on this, maybe if `shape == True`, we could remove the created file ?

amjjbonvin:
Not per se - results are good in the full run - but will test

src/haddock/modules/topology/topocg/__init__.py
```diff
@@ -298,21 +307,28 @@ def _run(self) -> None:
...
                     model_id = 0
                     origin_name_model = str(model.stem).split(".pdb")[0]
-
-                processed_pdb = Path(f"{origin_name_model}_cg_{force_field}.{Format.PDB}")
-                processed_topology = Path(f"{origin_name_model}_cg_{force_field}.{Format.TOPOLOGY}")
+                
+                shape_mod = shape_dic[i][j]
+                if not shape_mod:
+                    processed_pdb = Path(f"{origin_name_model}_cg_{force_field}.{Format.PDB}")
+                    processed_topology = Path(f"{origin_name_model}_cg_{force_field}.{Format.TOPOLOGY}")
```
VGPReys, line 331: In my opinion, this could be prone to maintenance error, as file name generation happens with two different logics:
- l.70 in this file + l.938 in `libaa2cg.py`
- l.328 in this file
Could we think of a dedicated function that performs filename generation and calling it when needed ?

rvhonorato: I agree with Victor - this should be a helper function. Also, these should be `PDBFile` and `TopologyFile` instead of `Path`

src/haddock/modules/topology/topocg/__init__.py
```diff
@@ -63,14 +63,20 @@ def generate_topology(
...
-        output_psf_filename=f"{cg_pdb_name[:-4]}_{force_field}.{Format.TOPOLOGY}",
-    )
-
-    input_str = prepare_single_input(str(cg_pdb_name))
+    cg_pdb_name, shape = martinize(input_pdb, output_path, False)
+
+    if not shape:
+        output = prepare_output(
+            output_pdb_filename=f"{cg_pdb_name[:-4]}_{force_field}{input_pdb.suffix}",
+            output_psf_filename=f"{cg_pdb_name[:-4]}_{force_field}.{Format.TOPOLOGY}",
```
rvhonorato: do not perform string slicing on filenames, use `pathlib` - plus this should be `PDBFile` and `TopologyFile` (this wrong type assignment appears in other places in the code)

end-to-end_tests/test_protein-protein-shape.py
```diff
@@ -0,0 +1,54 @@
...
+        assert Path(run_dir, "03_caprieval").exists(), f"03_caprieval not created"
+        assert Path(run_dir, "04_seletop").exists(), f"04_seletop created"
+        assert Path(run_dir, "05_flexref").exists(), f"05_flexref not created"
+        assert Path(run_dir, "06_caprieval").exists(), f"06_caprieval not created"
+        assert Path(run_dir, "07_cgtoaa").exists(), f"07_cgtoaa not created"
+        assert Path(run_dir, "08_emref").exists(), f"08_emref not created"
+        assert Path(run_dir, "09_caprieval").exists(), f"09_caprieval not created"
+        assert Path(run_dir, "10_clustfcc").exists(), f"10_clustfcc not created"
+        assert Path(run_dir, "11_seletopclusts").exists(), f"10_seletopclusts not created"
+        assert Path(run_dir, "12_caprieval").exists(), f"12_caprieval not created"
```
rvhonorato, line 52: Check if you need all this steps, keep in mind running things as `end-to-end` increases the CI runtime so keep only the essentials here

amjjbonvin:
This one basically test both the coarse graining and the shape docking, so probably a good one to have.

We need to run cgtoaa. The steps after that are peanuts in terms of timing.

This one runs on my laptop in about 2 1/2 minutes.

rvhonorato:
two minutes is a lot, could you maybe remove the intermediary caprievals?

amjjbonvin:
Actually 88s on my mac (for a full workflow it is not bad).

And commenting our the caprieval steps except for the last one runs in 80s

So 8 seconds are not worth changing the example I would say.

src/haddock/modules/topology/topocg/__init__.py
rversin: I do use PDBFile and TopologyFile at the end of the program, not at this stage. This is coherent with the topoaa module as well, where the usage of PDBFile and TopologyFile appears at the end of the program

src/haddock/modules/topology/topocg/__init__.py
rversin: This is the same pipeline as topoaa (outside of the libaa2cg).
In my opinion, if we want to implement these changes, they should be done after merging CG into the main branch, that way the rest of the haddock3 code and stay coherent

src/haddock/libs/libaa2cg.py
rversin: Stopping here causes issues with the cgtoaa module, as it is looking for the cg_to_aa table for all molecules involved in the docking

src/haddock/modules/topology/topocg/__init__.py
rvhonorato: > This is the same pipeline as topoaa (outside of the libaa2cg)

So it should also be changed there (:

src/haddock/modules/topology/topocg/__init__.py
rvhonorato: > This is coherent with the topoaa module as well,

Coherence is not a concept that applies here, it can be wrong in more than one place

src/haddock/libs/libaa2cg.py
rvhonorato: Victor is right - the check for shape should not be inside this function, what you need instead is a new function that just checks if something has a shape or not.

src/haddock/modules/topology/topocg/__init__.py
rversin: not in this PR imo

src/haddock/libs/libaa2cg.py
rversin: I've made some changes, and removed shape from martinize

src/haddock/modules/topology/topocg/__init__.py
rvhonorato: yep, this PR should be related to changes in `src/haddock/modules/topology/topocg/__init__.py`, see our comments above

src/haddock/libs/libpdb.py
```diff
@@ -506,3 +506,20 @@ def check_combination_chains(combination: list[PDBFile]) -> list[str]:
...
+        input_pdb (PDBFile): input PDB to check if shape.
+    
+    Returns:
+        bool: True if shape, False if not.
+    """
+    shape = False
+    with open(input_mol, 'rt') as input_file_mol:
+        if any('SHA' in line for line in input_file_mol):
+            shape = True
+    return shape
```
rvhonorato, line 525: here its checking for `SHA` in any place in the line, doesn't haddock expect it to be a residue name? so should check the proper location. also you can exit early if at least one `SHA` is found - currently its looping over the whole input regardless.

src/haddock/modules/topology/topocg/__init__.py
```diff
@@ -298,21 +310,30 @@ def _run(self) -> None:
...
                     origin_name_model = str(model.stem).split(".pdb")[0]
-
-                processed_pdb = Path(f"{origin_name_model}_cg_{force_field}.{Format.PDB}")
-                processed_topology = Path(f"{origin_name_model}_cg_{force_field}.{Format.TOPOLOGY}")
+                
+                shape_mod = shape_dic[i]
+                if not shape_mod:
+                    processed_pdb = Path(f"{origin_name_model}_cg_{force_field}.{Format.PDB}")
+                    processed_topology = Path(f"{origin_name_model}_cg_{force_field}.{Format.TOPOLOGY}")
+                    processed_cgtoaa_tbl=Path("../"+self.path.as_posix()+"/"+origin_name_model+"_cg_to_aa.tbl")
```
rvhonorato: resolve the full path, don't use `..`

src/haddock/modules/refinement/cgtoaa/cns/cgtoaa.cns
```diff
@@ -161,17 +165,19 @@ evaluate ($kinter = 0.001)
 
 flag excl * incl noe end
 
-! Fix the CG model
-fix sele=(segid *CG) end
+! Fix the CG model and any shape
+fix sele=(segid *CG or segid S) end
```
rvhonorato, line 169: should it be `or` or `and` here?

src/haddock/libs/libcns.py
```diff
@@ -385,28 +385,31 @@ def prepare_cns_input(
...
         else:
             pdb = input_element
+            shape = libpdb.check_mol_shape(pdb.rel_path)
             if pdb.aa_topology is None:
                 raise ValueError(f"All-Atom Topology not found {input_element.rel_path}."
                 "Conversion to all-atom requires a topology generated with [topoaa] and "
                 "[topocg].")
             aa_psf_list.append(pdb.aa_topology.rel_path.as_posix())
-            if pdb.cgtoaa_tbl is None:
+            if pdb.cgtoaa_tbl is None and not shape:
```
rvhonorato, line 410: shape should be a property of the pdb > `pdb.shape`

src/haddock/libs/libaa2cg.py
```diff
@@ -932,23 +945,19 @@ def martinize(input_pdb, output_path, skipss):
     cg_model = structure_builder.get_structure()
 
     # Write CG structure
-    cg_pdb_name = f"../{output_path}/{pdbf_path.split('/')[-1][:-4]}_cg.pdb"
+    cg_pdb_name = f"../{output_path}/{Path(pdbf_path).stem}_cg.pdb"
```
rvhonorato: resolve the full path, don't use `..`

src/haddock/libs/libaa2cg.py
```diff
@@ -958,7 +967,7 @@ def martinize(input_pdb, output_path, skipss):
     Path("temp.pdb").unlink(missing_ok=True)
 
     # Write Restraints
-    tbl_file_name = f"../{output_path}/{pdbf_path.split('/')[-1][:-4]}_cg_to_aa.tbl"
+    tbl_file_name = f"../{output_path}/{Path(pdbf_path).stem}_cg_to_aa.tbl"
```
rvhonorato: resolve the full path, don't use `..`

src/haddock/libs/libpdb.py
rversin: I am also using this function for the psf file, in the cutoff module 
I could make a spefici function for psf or a flag inside that function

src/haddock/modules/refinement/cgtoaa/cns/cgtoaa.cns
amjjbonvin: It is indeed OR, i.e. correct as is.  The non-shape parts will have a segid ending with CG while the shape has S

src/haddock/libs/libcns.py
rversin: ok done

src/haddock/libs/libpdb.py
rversin: ok it's fixed

rversin:
I don't understand why some check fails here, as locally (on tintin) it doesn't fail, it actually succeeds. I don't know if need to fix a module or test script ?

VGPReys:
I have two questions:
- `*_cg.pdb`: where should those files be written ? in the module (`XX_topocg`) directory ?
- `*_cg.pdb`: are those files used for backmapping ?

rversin:
These files are used by CNS to build the topology files. They are written in the module directory (01_topocg). They are not used for backmapping.

VGPReys:
Then let me fix this

VGPReys:
I modified:
- the uni-tests
- some paths generation in the `libaa2cg` and `topocg/__init__.py`
- added a removal of the `*_cg.pdb` files (that are temporary), when `debug = false`

rversin:
Thank you very much Victor for your help !

Issue #1522, opened 2026-04-16
Enable control of rebuilding of hydrogen atom in topoaa


amjjbonvin:
## **Desired feature/enhancement**

Currently the default behaviour in `topoaa` is to rebuild all hydrogen atoms.
There is however the option to only rebuild the missing ones, but this is only available within the CNS script `generate_topology.cns` 

This option should be exposed as a a parameter in topoaa.

## **Motivation**

For small ligands, the rebuilding might sometimes generate unrealistic positions of some hydrogen atoms.
So keeping the one present (and in correct positions) might alleviate this problem.

## **Description**

Replace by a parameter in `default.yaml` the following block in `generate_topology.cns` 
```
{* which hydrogens to build *}
{+ choice: "all" "unknown" +}
{===>} hydrogen_build="all";
```

## **Additional context**

Issue observed in a collaborative project (community tag)

PR #1523, opened 2026-04-16
Hydrogens

amjjbonvin:

## Checklist

Moved the `hydrogen_build` parameter from `generate_topology.cns` to `default.yaml` file in order to expose it as parameter to `topoaa` to control the way hydrogen atoms are (re)build.

## Related Issue

#1522

### DIFF
--- a/examples/docking-protein-ligand/refine-ligand-autotoppar-test.cfg
+++ b/examples/docking-protein-ligand/refine-ligand-autotoppar-test.cfg
@@ -1,0 +1,29 @@
+# ====================================================================
+# Protein-ligand docking example
+
+# directory in which the scoring will be done
+run_dir = "run1-refine-ligand-autotoppar-test"
+
+# execution mode
+mode = "local"
+ncores = 50
+debug = true
+clean = false
+
+# molecules to be docked
+molecules =  [
+    "data/oseltamivir_zwitterion.pdb"
+    ]
+
+# ====================================================================
+# Parameters for each stage are defined below
+# ====================================================================
+[topoaa]
+autohis = true
+delenph = false
+hydrogen_build = "unknown"
+autotoppar = true
+
+[emref]
+
+# ====================================================================
--- a/integration_tests/test_topoaa.py
+++ b/integration_tests/test_topoaa.py
@@ -1,5 +1,6 @@
 import tempfile
 import copy
+import gzip
 from pathlib import Path
 
 import shutil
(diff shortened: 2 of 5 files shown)


integration_tests/test_topoaa.py
```diff
@@ -147,6 +148,7 @@ def test_topoaa_module_ligand(topoaa_module):
     topoaa_module.params["ligand_top_fname"] = Path(GOLDEN_DATA, "ligand.top")
     topoaa_module.params["ligand_param_fname"] = Path(GOLDEN_DATA, "ligand.param")
     topoaa_module.params["delenph"] = False
+    topoaa_module.params["hydrogen_build"] = "unknown"
```
rvhonorato, line 176: This should be a new test - `test_topoaa_module_ligand_hydrogen_build`

amjjbonvin: We can test it in the same one - why a new test for it?  Saves time :-)  
But if you insist I will simply duplicate this test.

rvhonorato: because with this edit you change what is being tested, effectively by changing it you are creating a new test. so think of as removing the test that was there before and adding this one.

we should not remove tests, we should add more (:

amjjbonvin: And more is better! :-)  Done

integration_tests/test_topoaa.py
```diff
@@ -165,7 +165,7 @@ def test_topoaa_module_ligand(topoaa_module):
     assert expected_gz.exists()
 
 
-def test_topoaa_module_ligand(topoaa_module):
+def test_topoaa_module_ligand_hydrogen_build(topoaa_module):
```
rvhonorato, line 168: Interesting, two tests with the same name should have raised an error but it didn't 👀

examples/docking-protein-ligand/refine-ligand-autotoppar-test.cfg
VGPReys, line 1: What about placing that file in the `examples/refine-complex/` directory ?

VGPReys:
More fundamental question:
Don't we want to make this a molecule parameter ?
```toml
[topoaa]
[topoaa.mol1]
hydrogen_build = "unknown"
```

examples/docking-protein-ligand/refine-ligand-autotoppar-test.cfg
amjjbonvin: It is not a complex - and more related to ligand stuff

amjjbonvin:
> More fundamental question: Don't we want to make this a molecule parameter ?
> 
> ```toml
> [topoaa]
> [topoaa.mol1]
> hydrogen_build = "unknown"
> ```

It can in principle already be done - no need to define it under the molecule option in default.yaml - you can in the config file move the `hydrogen_build` parameter under the mol section and define different ones for different molecules. It works.

VGPReys (approved):

PR #1526, opened 2026-04-17
Let `filter` use RMDS 

### LINKED ISSUE #1525: Filter models based on global RMDS
Let `filter` module use global RMDS value, calculated in `caprieval`. 

## **Motivation**
This is handy when scoring AI-generated models, if one of the molecules in a complex is known to have a determined [to some degree] structure, e.g. antibody. Since AI is known to separate H and L chains (making “broken antibody”) and to put another molecule between them, it would be nice to make `filter` be able to exclude such broken antibodies automatically.

## **Description**
- pre-reqirement is to merge H+L, which is pretty much a standart practice
- run `caprieval` with an antibody unbound or modelled in isolation set as reference; 
- run `filter` with appropriate threshold (can be determined by examining a couple of models manually and then restarting the run from `filter`) 

## **Implementation**
I think of writing rmds field into `io.json` of `caprieval`, and then adding this rmds as another filtering option in `filter` module. From what I understand no need to change PDBFile class, `load` from `libontology` will be able to add a new field. 

## **Additional context**
I did this while scoring one of the capri targets, but at that moment I had to remove complexes with broken molecules manually.

AnnaKravchenko:
## Checklist

So far `caprieval` saves (global) RMDS to io.json. 

## Related Issue
#1525

### DIFF
--- a/src/haddock/modules/analysis/caprieval/__init__.py
+++ b/src/haddock/modules/analysis/caprieval/__init__.py
@@ -227,6 +227,11 @@ class HaddockModule(BaseHaddockModule):
         # Extract best references per input model
         best_ref_jobs = extract_models_best_references(jobs)
 
+        # Add temporary 'rmds' field to PDBFile so it gets stored in io.json
+        # in case next module needs it
+        for capri_obj in best_ref_jobs:
+            models[capri_obj.identificator - 1].rmsd = float(capri_obj.rmsd)
+
         # Write standard capri_ss file
         extract_data_from_capri_class(
             capri_objects=best_ref_jobs,


src/haddock/modules/analysis/caprieval/__init__.py
```diff
@@ -227,6 +227,11 @@ def _run(self) -> None:
         # Extract best references per input model
         best_ref_jobs = extract_models_best_references(jobs)
 
+        # Add temporary 'rmds' field to PDBFile so it gets stored in io.json
+        # in case next module needs it
+        for capri_obj in best_ref_jobs:
+            models[capri_obj.identificator - 1].rmsd = float(capri_obj.rmsd)
```
rvhonorato, line 233: This works because python, but its bad practice. Do not add unexpected properties to the ontology - they should be properly described and typed and documented in the object definition.

also you are referring here to this `io.json` file which is an artifact of earlier design, this is a representation of the `ModuleIO` interface and should actually not even exist on disk so don't consider it as a source of truth.

PR #1529, opened 2026-04-21
add caching to `ModuleIO` interface

### LINKED ISSUE #1528: Add caching to store RMSD values
Throughout an execution of a workflow multiple modules might require RMSD values. Currently there is no way to retrieve a previous value other than accessing information from a previous module.

There should be a system to cache relevant values in a given execution so they can be retrieved.

For this we also need to make sure that the input fingerprints are properly found - this can be done with a simple checksum.

rvhonorato:
This PR introduces a caching system to the `ModuleIO` interface to hold values that should not be re-calculated across modules.

In this first implementation I only added an `RMSD` key but this can be further expanded to hold anything.

It works by adding a `Cache` to the `ModuleIO` interface that has full serialization support. It has a `_store` to hold the information and a `_index` as a helper for faster O(1) lookup (standard caching implementation).

Being part of the `ModuleIO` interface means it is accessible to all modules and save/load operations in this interface also applies to `Cache`.

I have added a new method `update_cache` to the interface that is used by the primitive module class to propagate the cache when the module `export_io_models`.

A relevant technical detail is that for proper typing and serializadion/de-serialization of the `Cache` we need to drop the `jsonpickle` approach to represent the `ModuleIO` interface in favor of a simple (but not human readable) pickle. I also observed that a lot of tests are using a file representation of this interface as arguments to test some behaviour - so I have added a fallback to the `load` method to handle this scenario. This is not ideal but its a sensible compromise. 

### **No modules are using the cache yet!**

This PR here just adds the system, I will refactor the modules to use the cache in another PR (related to #1530 and #1525)

### DIFF
--- a/src/haddock/core/defaults.py
+++ b/src/haddock/core/defaults.py
@@ -20,7 +20,7 @@ MODULE_PATH_NAME = "step_"
 Module input and generated data will be stored in folder starting by
 this prefix"""
 
-MODULE_IO_FILE = "io.json"
+MODULE_IO_FILE = "io.pkl"
 """Default name for exchange module information file"""
 
 MAX_NUM_MODULES = 10000
--- a/integration_tests/test_caching.py
+++ b/integration_tests/test_caching.py
@@ -1,0 +1,191 @@
+"""Integration test for the caching functionality."""
+
+import pytest
+import tempfile
+from pathlib import Path
+import pickle
+
+from haddock.libs.libontology import ModuleIO, Cache, CacheKey
+
+
+def test_cache_basic_operations():
+    """Test basic cache operations: add, get, has, get_all."""
+    cache = Cache()
+
+    # Test adding items
+    cache.add(CacheKey.RMSD, ("model1", "model2"), 1.5)
+    cache.add(CacheKey.RMSD, ("model1", "model3"), 2.3)
+    cache.add(CacheKey.RMSD, ("model2", "model3"), 0.8)
+
+    # Test getting items
+    assert cache.get(CacheKey.RMSD, ("model1", "model2")) == 1.5
+    assert cache.get(CacheKey.RMSD, ("model1", "model3")) == 2.3
+    assert cache.get(CacheKey.RMSD, ("model2", "model3")) == 0.8
+
+    # Test has method
+    assert cache.has(CacheKey.RMSD, ("model1", "model2"))
+    assert not cache.has(CacheKey.RMSD, ("model3", "model4"))
+
+    # Test get_all
+    all_rmsd = cache.get_all(CacheKey.RMSD)
+    assert len(all_rmsd) == 3
+
+
+def test_cache_clear():
+    """Test clearing cache entries."""
+    cache = Cache()
+
+    # Add some items
+    cache.add(CacheKey.RMSD, ("model1", "model2"), 1.5)
+    cache.add(CacheKey.RMSD, ("model3", "model4"), 2.0)
+
+    # Verify items exist
+    assert cache.has(CacheKey.RMSD, ("model1", "model2"))
+    assert cache.has(CacheKey.RMSD, ("model3", "model4"))
+    assert len(cache.get_all(CacheKey.RMSD)) == 2
+
+    # Clear RMSD cache
+    cache.clear(CacheKey.RMSD)
+
+    # Verify cache is empty
+    assert not cache.has(CacheKey.RMSD, ("model1", "model2"))
+    assert not cache.has(CacheKey.RMSD, ("model3", "model4"))
+    assert len(cache.get_all(CacheKey.RMSD)) == 0
+
+
+def test_module_io_cache_serialization():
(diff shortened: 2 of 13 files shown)


VGPReys:
I think it would be nice to have a face-to-face discussion about this implementation, to better understand it

rvhonorato:
What is your question?

amjjbonvin:
I agree - let’s discuss it next week Tuesday?

rvhonorato:
I disagree, would rather you expose your questions here in the name of open science (:

amjjbonvin:
Nothing to do with open science - we simply wish to discuss this in person and understand the implications.

rvhonorato:
oke, but this is then offscope of code review.

So I will merge this as i need it for #1530 #1525 and it has currently no impact in how the project works.

src/haddock/core/defaults.py
```diff
@@ -20,7 +20,7 @@
 Module input and generated data will be stored in folder starting by
 this prefix"""
 
-MODULE_IO_FILE = "io.json"
+MODULE_IO_FILE = "io.pkl"
```
amjjbonvin, line 23: So we are changing the name/format of the file that contains the module information, right?

amjjbonvin: Not sure if we document this somewhere, but the manual might have to be updated.

amjjbonvin (commented):
I am from time to time looking at the json file to find some information - now it works easily since it is a text file. The pkl file is however not simply readable which adds complexity. Is it really needed to change the format?  

From a human perspective it is not ideal (and for sure not for e.g. debugging purposes - specially when using only a terminal). 

And speed is not really an issue for this.

rvhonorato:
> I am from time to time looking at the json file to find some information - now it works easily since it is a text file. The pkl file is however not simply readable which adds complexity. 

I do not know why this was a `.json` and it is the other way around - the `.json` is actually much more complex than the pickle, since one is simply a binary and the other is a text file. Handling it as a text file requires much more work to save/load an handle the serialization and de-serialization of it and the binary is just simpler.

I understand you were using this as debugging tool but that is far from ideal as this contains a "snapshot" of the contents of the `ModuleIO` interface **only at the end** - so it only gives you a very brief glimpse of the communication and it will be missing a lot of relevant operations that happen **after** the file is loaded and **before** it is saved.

Maybe this glimpse is enough for your debugging but it is not enough to justify the usage of this complex `.json` format - besides this giving you incomplete information its also very limiting in what we can do with it. If you want a debugging system for the `ModuleIO` interface I can easily write it.

> Is it really needed to change the format?

Yes, besides the reasons I said above using the binary we can benefit from passing generic types and not having to worry about their serialization/deserialization. See here:


```python

class CacheKey(Enum):
    RMSD = auto()


K = TypeVar("K", bound=Hashable)
V = TypeVar("V")


class Cache(Generic[K, V]):
    # ...
```

`K` and `V` are generic types used in the generic `Cache` class and there is no way to express this as a text file without me having to write a `__repr__` method - which in my opinion defeats the purpose of the generics. The idea is to keep adding more `CacheKey`s - that represent operations that should not be repeated, and to make this fit anything we need the generic - so yes, we need to use a binary instead of a text file.

amjjbonvin:
This clearly needs some more discussion within the team (and f2f).

For example I know that people have been modifying those io.json files for specific purposes during CAPRI and then restart the workflow from a specific module. This is flexibility that matters.

And while a module is running you won’t look at what is happening anyway.

But your mechanism could allow to write regular timestamps of it, which might allow in the future to restart a partially completed module from the last time point (now we can only restart a module from scratch - meaning we loose any intermediate results of the module).

rvhonorato:
> For example I know that people have been modifying those io.json files for specific purposes during CAPRI and then restart the workflow from a specific module. This is flexibility that matters.

Doing this you are using a file that should not even exist in the first place to change the behavior of the application outside the existing interfaces. This is hacking, this is not flexibility.

> This clearly needs some more discussion within the team (and f2f).

Sure - but again this is off scope the code here. The only behavior that will change once this is merged is related to people hacking the execution and debugging the execution using the wrong tools. Not sufficient to block the feature in my opinion.

amjjbonvin:
This hacking and debugging is often needed for user support and/or understand issues.

So first a f2f discussion with the team before merging this.

rvhonorato:
It's unfortunate that you would hold this feature to prioritize bad practices in the development workflow.

I will then close this one and I can later re-do it based on the new priorities that come out of this discussion.

amjjbonvin:
It might still get implemented, after discussion. So in principle no need to close the issue

Open science goes hand in hand with team science :-)

Why so impatient?

rvhonorato:
Pull requests should not be opened if they are not going to be reviewed - none of the points in this PR are related to the code I proposed so there is no point in having this pull request open since a discussion is still needed. Once the discussion round is done the changes need to be re-structures, thus a new PR should be opened. 

Standard best practice (:

PR #1544, opened 2026-04-30
Fixed the issue with param/top files for ligand ensembles - Issue #1540

### LINKED ISSUE #1540: ligand param files not passed to next modules when an ensemble of ligands is used in autotoppar=true mode.

topoaa - with the error actually showing up in the subsequent modules (rigidbody or emref tested)

## **Expected behavior**

When using an ensemble of ligand (possibly different one) and setting `autotoppar=true`, the `io.json` file written in the `topoaa` directory should contain the ligand param/top file information for each ligand.

## **Actual behavior**

When using an ensemble of ligand (possibly different one), the io.json file written in the topoaa directory has no info about the ligand param/top files. E.g.:

```
                "ligand_param_fname": null,
                "ligand_top_fname": null,
```

When a single ligand conformation is provided, io.json file correctly contains the automatically generated param/top files:

```
                "ligand_param_fname": {
                    "py/reduce": [
                        {
                            "py/type": "pathlib.PosixPath"
                        },
                        {
                            "py/tuple": [
                                "/Users/abonvin/Desktop/test/output-single/0_topoaa",
                                "dock_1_model_ligand_prodrg.param"
                            ]
                        }
                    ]
                },
                "ligand_top_fname": {
                    "py/reduce": [
                        {
                            "py/type": "pathlib.PosixPath"
                        },
                        {
                            "py/tuple": [
                                "/Users/abonvin/Desktop/test/output-single/0_topoaa",
                                "dock_1_model_ligand_prodrg.top"
                            ]
                        }
                    ]
                },
```

1. Unpack the test.tgz archive provided with this issue
2. Go into the test directory
3. Execute: `haddock3 score-single.cfg` -> runs complete fine
4. Execute: `haddock3 score.cfg` -> runs fail at the the rigidbody module
5. Compare the `rigidbody_1.inp` files from the `output` and `output-single` runs, search for `ligand_param_fname`
6. Compare the `io.json` files in the respective `0_topoaa` directories, searching from `ligand_param_fname` (look here for the param files associated with the ligand).

## **Version**

630ad95170e7e1e12aa8bdadca3db251da7af44f

[test.tgz](https://github.com/user-attachments/files/27200259/test.tgz)

amjjbonvin:

## Checklist

Corrected `__init__.py` in the `topoaa` modules so that the PRODRG generated ligand param/top files are properly written to the `io.json` file when `autotoppar=true` and an ensemble of ligands is provided.

## Related Issue

#1540


integration_tests/test_topoaa.py
```diff
@@ -259,6 +259,44 @@ def test_topoaa_module_ligand_automated(topoaa_module):
...
+    assert expected1_pdb.exists()
+    assert expected2_inp.exists()
+    assert expected2_psf.exists()
+    assert expected2_pdb.exists()
+
+    with open(expected_io, encoding="utf-8", mode="r") as f:
+        file_content = f.read()
+
+    assert "ligand-ens_1_prodrg.param" in file_content
+    assert "ligand-ens_2_prodrg.param" in file_content
```
rvhonorato: Do not perform file operations on this file, check the module directly:


```suggestion
    # Check if the property was added to the output
    for structure in topoaa_module.output_models:
        for model in structure:
            pdb = structure[model]

            # Check the contents
            assert pdb.ligand_top_fname.exists()
            assert pdb.ligand_param_fname.exists()

```

rvhonorato (changes requested):
The added test is not covering the proposed change - the test still passes by reverting the change. 

Please add a test that should fail without the proposed change in `topoaa.__init__.py` so that we can confirm the change fixes the wrong behavior.

The flow should be: add a failing test that replicates the behavior -> make the change -> confirm new test pass -> run all tests.

amjjbonvin:
> The added test is not covering the proposed change - the test still passes by reverting the change.
> 
> Please add a test that should fail without the proposed change in `topoaa.__init__.py` so that we can confirm the change fixes the wrong behavior.
> 
> The flow should be: add a failing test that replicates the behavior -> make the change -> confirm new test pass -> run all tests.

You are right! The issue only occurs if another molecule is defined. Adding the protein is required.
But now in my test I am getting an error which does not occur when running the workflow on the test system (the archive attached with the issue). And a different one actually, which I don't get when running it manually. So something must be wrong with the settings in the test. Below is the workflow that works fine. It uses the same input models as the one of the test (those are present in the golden_data dir).

```toml
run_dir = "output"

mode = "local"
debug = true
clean = false
ncores = 20

molecules = [
    "ligand-ens.pdb",
    "protein-ens.pdb",
]

[topoaa]
autotoppar = true

```

Running the test I am now getting as error:

```
>           parameters_for_this_molecule = deepcopy(mol_params[mol_params_get()])
                                                               ^^^^^^^^^^^^^^^^
E           IndexError: pop from empty list

src/haddock/modules/topology/topoaa/__init__.py:250: IndexError
```

But this does not occur when running manually. Can you take a look at the test?

amjjbonvin:
Ok - figured it out!

VGPReys (approved):
Ow, so it was just a miss placement of the variable !

rvhonorato (approved):
With the new test, reverting the change makes it fail as expected, replicating the issue:

```
        # Check if the property was added to the output
        for structure in topoaa_module.output_models:
            for model in structure:
                pdb = structure[model]

                # Check the contents
                if "ligand" in pdb.file_name:
>                   assert pdb.ligand_top_fname.exists()
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                   AttributeError: 'NoneType' object has no attribute 'exists'

integration_tests/test_topoaa.py:305: AttributeError
```

Applying your change fixes the test, thus fixing the bug 🚀

PR #1550, opened 2026-05-01
add `caprifilter` and move functions to `libcapri.py`

### LINKED ISSUE #1525: Filter models based on global RMDS

AnnaKravchenko:

## Checklist

- [x] Tests (made using AI) added for the new code 
- [x] Modifications / enhancements are reflected on the [haddock3 user-manual] (https://github.com/haddocking/haddock3-user-manual) (PR is open)

Making of a caplifilter module, which can filter models based on multiplt CAPRI metrics. 
CAPRI metrics calculation are the same as in caprieval, so shared fuctions are moved to libcapri.

caprifilter calculates requested metrcis, filters models, writes caprifilter.tsv with kept models,  caprifilter_ss.tsv that is basically a copy of capri_ss.tsv, and, if enabled by user, caprifilter_full.tsv, where all modes are listed along with mertics and kept/filtered status. 

## Related Issue
#1525 

## Additional Info
Moved find_ff to libstructure, and a handle_input_reference to libpdb. Also had to edit each file/test that was calling functions from capri.py to correct it to libcapri - this is why e.g. alascan is in modified files


src/haddock/modules/analysis/rmsdfilter/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+                "modules that produce a nested list of models."
+            )
+        
+        # Check if cluster info is present
+        if any(m.clt_id is not None for m in models):
+            self.finish_with_error(
+                "Models have been clustered!"
+                "[rmsdfilter] cannot be performed after clustering - "
+                "filtering individual models after clustering would leave "
+                "remaining models with stale and inconsistent cluster assignments."
```
VGPReys: I would simply overwrite the attribute and go on

src/haddock/libs/libpdb.py
```diff
@@ -523,3 +526,67 @@ def check_mol_shape(input_mol: Path) -> bool:
...
+        Path to the input reference structure, possibly containing
+        an ensemble.
+
+    Returns
+    -------
+    reference or first_model_path : Path
+        Path to the reference structure to be used downstream.
+    """
+    if reference.stat().st_size == 0:
+        raise ValueError(f"Reference file is empty: {reference}")
```
VGPReys: Yes, it checks that the file is indeed empty, but not that it contains coordinates. Could be composed of `REMARKS` or not even be a pdb file

VGPReys:
Do you see an option for user to chose between "global rmsd", or "l-rmsd" or "il-rmsd" or "fnat" or "dockq" too ?
You could even directly use the CAPRI class to perform all theses computations ?

src/haddock/modules/analysis/rmsdfilter/__init__.py
AnnaKravchenko: Why not? Then I guess models have to be renamed to smth like model_filter_1.pdb

AnnaKravchenko:
> Do you see an option for user to chose between "global rmsd", or "l-rmsd" or "il-rmsd" or "fnat" or "dockq" too ? You could even directly use the CAPRI class to perform all theses computations ?

Can you think of how those metrics would be useful for filtering? For RMDS the idea is to remove model from the workflow - since no other way to do that without manual tweaking, but I don’t see how filtering by fnat or dockq would be useful, so not sure if it’s worth the time. 

Maybe let’s discuss this Wednesday? Will be easier in person

src/haddock/libs/libpdb.py
AnnaKravchenko: Just checked that case - `libalign` already handles it with error message "Please check the input file and queried filterings.” 

But also now that I look in the code - may-be it’s better to remove this check entirely and instead in `libpdb.py > handle_input_reference > line 570 ` 
```
    for line in wc_return.split("\n"):
        if "No. models" in line:
            nb_models = int(line.strip().split()[-1])
            break
```
add lines to read the number in line `No. atoms` and if 0 - exit with message like “No atoms found in the reference file, please check”. What do you think?

src/haddock/modules/analysis/rmsdfilter/__init__.py
VGPReys: Not necessary, e.g:
- `seletop` does not generate new names
- `filter` does not generate new names
- scoring modules that do not change the coordinates do not generate new names, but rather pass the inputs to their outputs, and simply update the `score` attribute

src/haddock/libs/libpdb.py
```diff
@@ -523,3 +526,71 @@ def check_mol_shape(input_mol: Path) -> bool:
...
+    # Memorize previous sys.stdout
+    original_stdout = sys.stdout
+    # setup the new stdout environment
+    sys.stdout = TextIOWrapper(BytesIO(), sys.stdout.encoding)
+
+    # Count number of models
+    with open(reference, "r") as fh:
+        pdb_wc(fh, "m")
+    # Get output
+    sys.stdout.seek(0)  # Jump to the start
```
rvhonorato: import these standard libs on the top of the file:
```python
from contextlib import redirect_stdout
from io import StringIO
```

then you can replace all of this with:

```python
  buf = StringIO()
  with open(reference, "r") as fh, redirect_stdout(buf):
      pdb_wc(fh, "m")
  wc_return = buf.getvalue()
```

AnnaKravchenko:
It would be amazing if somneone could take a look at this PR - to make sure it’s alright before I start on new tests and haddock user manual

AnnaKravchenko:
Failing tests (`integration_tests/test_rigidbody.py::test_rigidbody_mpi` and `integration_tests/test_rigidbody-cg.py::test_rigidbody_mpi`) seems to be unrelated to the changes introduced by this PR

Guess it will be back to noramal soon, as Rodrigo fixes the lastest "tweak CI” PR

src/haddock/modules/analysis/caprifilter/__init__.py
```diff
@@ -0,0 +1,261 @@
...
+                for ref in references
+            ]
+
+        # Create one CAPRI job per (model, reference) pair
+        jobs: list[CAPRI] = []
+        for i, model in enumerate(models, start=1):
+            if isinstance(model, list):
+                raise ValueError(
+                    "[caprifilter] cannot handle a list of models per job."
+                )
```
rvhonorato: You can add this check together with `if self.is_nested(models)` - but its correct as is

src/haddock/modules/analysis/caprifilter/__init__.py
```diff
@@ -0,0 +1,261 @@
...
+                capri_objects=jobs,
+                output_fname=Path(".", "caprifilter_multiref.tsv"),
+                sort_key=self.params["sortby"],
+                sort_ascending=self.params["sort_ascending"],
+                add_reference_id=True,
+            )
+
+        # Collect metrics and apply all filters
+        metrics_data = collect_metrics(best_ref_jobs)
+        kept, filtered_out = filter_models(metrics_data, filter_specs)
```
rvhonorato: this `filtered_out` is not used - is it on purpose?

src/haddock/modules/analysis/caprifilter/__init__.py
```diff
@@ -0,0 +1,261 @@
...
+    @staticmethod
+    def is_nested(models: list[Union[PDBFile, list[PDBFile]]]) -> bool:
+        for model in models:
+            if isinstance(model, list):
+                return True
+        return False
+
+    def _run(self) -> None:
+        """Execute module."""
+        if type(self.previous_io) == iter:
```
rvhonorato: `iter` is a function, not a type and this conditional will never trigger - what you need here is `isinstance(self.previous_io, Iterator)`

rvhonorato (dismissed):
good job moving the functions around, having a `libcapri` is a good change. 

as for the `caprifilter` some parts of the code are very "noisy" and could be written in a leaner more direct way but it's all correct as far as I can read it.

I also made some comments but I see no blocking issues, looks good to me as is.

src/haddock/modules/analysis/caprifilter/__init__.py
AnnaKravchenko: Oh, this line `if type(self.previous_io) == iter:` is present in 6 analysis modules. 
Will fix it in caprieval and caprifilter here, but for the other 4 models it probably should be a new PR

src/haddock/modules/analysis/caprifilter/__init__.py
AnnaKravchenko: `filtered_out` is being assessed in tests, but here it is indeed not needed, gonna replace it with `_`

rvhonorato (approved):

AnnaKravchenko:
Will merge this now, and fix iter guard for capri* modules in corresponding PR

Issue #1551, opened 2026-05-02
Automatic flexibility should exclude water molecules


amjjbonvin:
## **Module/Workflow/Library affected**

mdref
mdscoring

## **Expected behavior**

When defining automatically the regions that are not restrained during the refinement in explicit solvent, the solvent should not be considered when looking for interactions. Only residues interacting with other molecules (not solvent) should be considered.

## **Actual behavior**

Solvent molecules are also considered, which gives more flexibility to a large part of the molecules.
Note that the runs complete fine. 

## **Steps to reproduce the behavior**

Execute on of the example where `mdref` is used (e.g. docking-protein-protein-mdref-test.cfg`)
Set debug=true and look in one of the mdref_1.out file, searching for AUTOMATIC -> many residues are selected.

## **Suggestions on how to fix it**

Exclude solvent from the selection in the `flexauto-neighbors.cns` CNS scripts of `mdref` and `mdscoring`

## **Version**

f3785f8340afa6889fbe7000c942bd49ef46b723

## **Additional context**

As`mdref` is no longer used as default, the impact of this issue is limited. The current unintended additional flexibility might be beneficial or not depending on the docking case.

PR #1552, opened 2026-05-02
Exclude water from automatic flexibility definition in mdref/mdscoring

amjjbonvin:
## Checklist

Updated the automatic definition of flexible residues in `mdref/mdscoring` to exclude the solvent.

## Related Issue

#1551 

## Additional Info

Was not causing failures, but resulted in large parts of models being flexible while mainly the interface regions should be treated as fully flexible.


VGPReys (approved):
what about haddock2.4 ?

amjjbonvin:
> what about haddock2.4

Good point - same issue. Must have been there for quite some time then...

Issue #1562, opened 2026-06-05
Cosmetic fix in support of custom symmetry restraints


AnnaKravchenko:
Line 
`CNSsolve> noe class symm @@$symtbl_fname end` in in /src/haddock/modules/…/cns/symmultimer.cns lead to the following error:
`%NOESET error encountered: duplicate class name specified.`
`(CNS is in mode: SET ABORT=NORMal END)`

This is not a bug - custom symmetry restraints are functional.

PR #1563, opened 2026-06-05
cosmetic fix in custom symmetry 

AnnaKravchenko:

## Checklist

"noe class symm” was declared twice, removed 2nd declaration. 

## Related Issue
#1562 

## Additional Info
As an example I added a toy model which I use to work on stacking. Current stymmetry restraints are not enough for the stacking, but symmetry itself is functional (distances between selected atoms differ by no more than 0.1A).
I added it for ease of testing, but I don’t think it’s not worth keeping it for now.


AnnaKravchenko:
Also it seems there’s no bug in haddock2.5 - haddock25/protocols /symmultimer.cns does not have this double definition

amjjbonvin (commented):
What about mdref?

And is the example  functional? Meaning the solution show the custom symmetry you are expecting (stacking?)?

AnnaKravchenko:
Forgot mdref, adding now.

And for example - it is functionall symmetry-wise, i.e. distances that have to be symmetrical differ max by 0.1A. 
But selection of 4 atoms per ring is not enough for the perfect stacking (in rigidbody, didn’t test further, since it’s about the bug)

amjjbonvin:
The example might be confusing to users. Would be better to create an integration test for this. Or add it to the end-to-end tests

AnnaKravchenko:
> The example might be confusing to users. Would be better to create an integration test for this. Or add it to the end-to-end tests

It turned out that the error I saw wasn’t impacting the functionality of the custom symmetry. There’s already an integration test `integration_tests/test_customsym.py` which passed sucessfully both before and after this fix with identical values of symmetry energy violation.  

So at the end there changes ony remove:
` %NOESET error encountered: duplicate class name specified. `
`(CNS is in mode: SET ABORT=NORMal END) `

But does’t alter actual functionality.

AnnaKravchenko:
So, at the end, it is not a bug.
But probably still make sence to merge it - to avoid confusion if/when someone will try to read cns log file?

amjjbonvin:
That's what I thought. But indeed good to merge to avoid this confusing error message in the CNS output

amjjbonvin (approved):

PR #1570, opened 2026-06-12
fix FutureWarning when assigning string to int64 cluster_id column - issue #1569

### LINKED ISSUE #1569: incompatible dtype in libplots

Makes libplot (near)future-proof

## **Motivation**

Better to fix it now to avoid a future problem.

## **Description**

libplots.py:852: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas. Value 'Other' has dtype incompatible with int64, please explicitly cast to a compatible dtype first.
    other_structs_df.loc[:,'cluster_id'] = 'Other'

amjjbonvin:
Cast cluster_id to object dtype before assigning 'Other' to avoid pandas FutureWarning about incompatible dtype assignment.

## Checklist

Corrected libplot to avoid future issues.

## Related Issue

#1569

### DIFF
--- a/src/haddock/libs/libplots.py
+++ b/src/haddock/libs/libplots.py
@@ -849,6 +849,7 @@ def create_other_cluster(
     other_structs_df = structs_df[structs_df['cluster_ranking'] >= max_clusters].copy()
     # drop other clusters from structs_df
     structs_df = structs_df[structs_df['cluster_ranking'] < max_clusters].copy()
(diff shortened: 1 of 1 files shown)


rvhonorato (approved):
It's a bit awkward to have `cluster_id` being integer or string considering its a categorical value - for me it would make more sense to keep it only as a string.

this casting works to suppress the issue but still whatever is consuming this needs to handle the case in which `cluster_id` type is `string` or `int` - seems like an oversight, would be simpler to not have these two types, but this PR is good as is.

PR #1574, opened 2026-06-15
add dev docker image on main push

rvhonorato:
Build a dev Docker image on every push to main, constantly overwritten — does not affect latest, which remains tied to release tags.

This will make the latest version of the code available via `dev` tag in `ghcr.io` - which is valuable given the release conditions of the project are arbitrary


amjjbonvin (approved):

Comp-era (approved):
Tested locally — the PR head (branch docker, full python:3.14) builds cleanly with Docker

PR #1573, opened 2026-06-14
update paths in `caprieval` analysis report

rvhonorato:
You are about to submit a new Pull Request. Before continuing make sure you read the [contributing guidelines](CONTRIBUTING.md).

## Checklist

- [x] Tests added for the new code
- [ ] Documentation added for the code changes
- [ ] Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)
- [ ] `CHANGELOG.md` is updated to incorporate new changes
- [X] Does not break licensing
- [X] Does not add any dependencies, if it does please add a thorough explanation

## Summary of the Pull Request  

In `clt_table_handler` the model paths from `capri_ss.tsv` are relative to NN_caprieval/, but the generated `report.html` lives one level deeper in the directory structure.

Links built directly from the tsv values resolved to a non-existent path, causing 404s when clicking model links in the report. The fix is to prepend an extra `../` to the model column paths so links correctly resolve

Also fixed an existing fixture bug: `example_capri_clt` was pointing at `capri_ss_example.tsv` instead of `capri_clt_example.tsv`.

multi-line list formatting).
## Related Issue

#1568

PR #1575, opened 2026-06-15
add `ruff` format CI check for changed files

rvhonorato:
Add a format action check and update the documentation in regards to which format should be used in the project.

_Suggested by @sverhoeven in https://github.com/haddocking/haddock3/pull/1573#pullrequestreview-4496589721_

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,35 @@
+---
+name: format
+
+on:
+  pull_request:
+  workflow_dispatch: ~
+
+jobs:
+  ruff-format:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.14"
+
+      - name: install haddock3 with dev dependencies
+        run: pip install -e '.[dev]'
+
+      - name: get changed python files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+        with:
+          files: |
+            **/*.py
+
+      - name: check formatting on changed files
+        if: steps.changed-files.outputs.any_changed == 'true'
+        run: ruff format --check --diff ${{ steps.changed-files.outputs.all_changed_files }}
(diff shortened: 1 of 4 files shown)


.github/workflows/format.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.14"
+
+      - uses: astral-sh/setup-uv@v3
```
sverhoeven: These actions are old and will complain about node version 20 will be deprecated in Sept 2026. Please update them to newer versions.

See https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

.github/workflows/format.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: get changed python files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+        with:
+          files: |
+            **/*.py
+
+      - name: check formatting on changed files
+        if: steps.changed-files.outputs.any_changed == 'true'
+        run: ruff format --check --diff ${{ steps.changed-files.outputs.all_changed_files }}
```
sverhoeven, line 37: Nice so only changed files get format check. 

This means unchanged files are not checked. Do you plan to format all files at some point, for example before a new release with [ignore format commit in blame](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view)? 

Or do you slowly want to format the code a PR at a time?

sverhoeven (changes requested):
Docs look good.

The workflow should be updated and clearified

Note:
`ruff format` does not organize imports for that you have to use `ruff check --select I --fix <path/to/changed_file.py>`, you could do that in a follow up PR for using `ruff check` as linter.

.github/workflows/format.yml
rvhonorato: I have tried to implement formatters here before and there was some push back from the developers - so yep I'm thinking its better to have it as is now so that the code gets formatted slowly over time.

sverhoeven (approved):
Ok gradual formatting it is.

Looks good the [format workflow](https://github.com/haddocking/haddock3/actions/runs/27552166153) is no longer complaining.

PR #1576, opened 2026-06-15
re-introduce `gdock` module

### LINKED ISSUE #1595: re-implement `gdock` as a sampling module
`gdock` as a sampling module was first introduced in https://github.com/haddocking/haddock3/pull/72 and then removed in https://github.com/haddocking/haddock3/pull/1454. 

The reason for removal is that `gdock` depended on an external dependency (`dcomplex`) for its scoring function, which the authors removed from the internet. 

Since then I have completely re-written `gdock` from scratch and implemented the energy calculation myself ([see the release notes for more info about the re-write](https://github.com/rvhonorato/gdock/releases/tag/v2.0.0)) - so it's ready to be added here again

rvhonorato:
## Checklist

- [ ] ~Modifications / enhancements are reflected on the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)~
    - non-applicable, we still need to do some deeper evaluation of the integration
    - `gdock` has a zero-clause license so it's fully compatible here
- [ ] ~Does not add any dependencies, if it does please add a thorough explanation~
    - adds [`gdock-py`](https://codeberg.org/rvhonorato/gdock-py) which has no parent dependencies

This PR re-introduces `gdock` as a sampling module. It was first introduced in https://github.com/haddocking/haddock3/pull/72 and then removed in https://github.com/haddocking/haddock3/pull/1454. The reason for removal is that `gdock` depended on an external dependency for its scoring function that was no longer available - so the whole software was not functional anymore. 

Since then I have completely re-written `gdock` from scratch and implemented the energy calculation myself ([see the release notes for more info about the re-write](https://github.com/rvhonorato/gdock/releases/tag/v2.0.0))

To fulfill this integration into haddock3 I have written thin bindings (bits of code that wrap the `gdock`'s source code in python) to make the addition here simpler - see [`gdock-py`](https://codeberg.org/rvhonorato/gdock-py).

Since haddock uses ambiguous restraints, I have also added this concept in `gdock` so it can seamlessly read haddock's tbl file and can handle the `OR` logic. I have implemented, to the best of my knowledge, the same function that haddock uses for its AIR restraint calculation - https://codeberg.org/rvhonorato/gdock/pulls/64

As for the integration itself I also added some compatibility to be in line with the other sampling module; `[gdock]` can handle ensembles by running each combination and it can also handle `crossdock=True`.

The key parameters I have exposed in the `gdock` interface are:

| Parameter | Default | Description |
|-- | -- | -- |
| `ambig_fname` | "" | AIR restraints .tbl file |
| `max_generations` | 250 | GA generation limit |
| `number_of_individuals` | 150 | GA population size |
| `sampling` | 1000 | Models to collect per input pair** |
| `seed` | 42 | Random seed |
| `crossdock` | true | Dock all receptor × ligand combinations |

> **sampling is a special parameter I added just for this integration, it will go bypass `gdock`'s selection and return up to N models that are different enough based on its HoF (Hall-of-Fame) definition

I have also created some example files for the compatible scenarios replacing `rigidbody` with `gdock` - however we cannot make a one-to-one comparison given that `gdock` and `haddock` are using different formulations of the OPLS forcefields.

We would need a deeper investigation to properly parametrize which parameter and module configurations we should use to obtain similar results - this is however way far beyond the scope of this PR.

Here I am "just" adding `gdock` as a module and I think this can also serve as a good example of how to add new (external)-modules, so please keep the discussion in this PR so it serves for future reference.

## Related Issue

#1595 

## Additional Info

Please have a look at [gdock.org](https://gdock.org) for full information about how `gdock` works. I did not add the whole information here since its described elsewhere.


rvhonorato:
The `gdock` module is properly implemented and functional.

However I've hit a conceptual wall in the way restraints are treated. `gdock` currently (`v2.1.0`) can only handle PAIRS of restraints. So while running a `gdock x rigidbody` comparison in the cherry-picked examples I was naively just "reducing" the restraints to a set of pairs, but this creates a set of impossible to satisfy constraints which results in models with a lot of clashes.

So I will pause work on this while I implement ambiguous restraints in `gdock`, follow here: https://codeberg.org/rvhonorato/gdock/pull/64

For now I'll keep the PR open since its just additive - but if the code diverges too much I'll reopen it.

rvhonorato:
@amjjbonvin please add whoever else you think is relevant for this review

amjjbonvin (commented):
A very nice (back) addition!

Few comments:

1) Is there a way to define the weights of the scoring function? Not in the default.yaml file 

2) Please rename the example workflow files to be consistent with the other existing ones, e.g. `docking-protein-protein-gdock.cfg` instead of `protein-protein-gdock.cfg` 

3) The definition of `caprieval` in the workflow as `caprieval.1`, `caprieval.2` does not work - those steps are silently skipped when executing the workflow. Should simply be `caprieval` for all steps.

4) If we want those to be executed in the monitoring jobs, test versions of the workflows should be created with limited sampling

5) Performance from a simply quick test seems quite ok! We will have to do some proper benchmarking

amjjbonvin:
One more thing: it would also be nice to expose if possible the dielectric constant parameter epsilon - especially useful in the case of nucleic acids (if those are supported by gdock)

rvhonorato:
>     1. Is there a way to define the weights of the scoring function? Not in the default.yaml file

Can add in another PR if necessary


>     2. Please rename the example workflow files to be consistent with the other existing ones, e.g. `docking-protein-protein-gdock.cfg` instead of `protein-protein-gdock.cfg`

Oke

>     3. The definition of `caprieval` in the workflow as `caprieval.1`, `caprieval.2` does not work - those steps are silently skipped when executing the workflow. Should simply be `caprieval` for all steps.

Indeed, this should have raised an error instead of failing silently, opened an issue for it: https://github.com/haddocking/haddock3/issues/1616 and here fixed it with the proper toml syntax.

>     4. If we want those to be executed in the monitoring jobs, test versions of the workflows should be created with limited sampling

Not applicable here

> 
>     5. Performance from a simply quick test seems quite ok! We will have to do some proper benchmarking

Yes this is what I said in the description

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
+=======
 - 2026-07-01: Removed unused guard for “previous module produced iterable” - Issue #1604
+- 2026-06-xx: Re-add `gdock` as a sampling module
```
amjjbonvin: Update the date - today I would say

amjjbonvin (commented):
Where should we open an issue to expose the weights of the scoring function in gdock (and also epsilon?)? Both on the gdock repo and a new issue in haddock3 I guess.  Also aren't you using the BSA in the gdock score (as in rigid-body in haddock)?

rvhonorato:
> Where should we open an issue to expose the weights of the scoring function in gdock (and also epsilon?)? Both on the gdock repo and a new issue in haddock3 I guess. 

The weights are exposed as options the main application, I need to propagate them to the python binding but you can open an issue in the haddock3 repository since its related to this integration specifically

> Also aren't you using the BSA in the gdock score (as in rigid-body in haddock)?

`gdock` uses desolv

amjjbonvin:
Ok - haddock uses both desolv and bsa - but perfectly fine

amjjbonvin (approved):

PR #1590, opened 2026-06-26
remove duplicated parameter definitions in `fragment_probes.param`, `protein-allhdg5-4.param` and `protein-CG-Martini-2-2.param`

### LINKED ISSUE #1589: Remove duplications in parameter files

Some parameter files, especially the CG ones have many duplications of parameter definitions.

## **Motivation**

Cleaning up will reduce the file sizes.

Reducing the size of the number of parameters in the various force field by removing unnecessary duplications will also avoid problems with parameter array sizes in CNS (which is the past required recompilation of the code)

## **Description**

Remove duplicate definitions of BOND/ANGLE/DIHE parameters

amjjbonvin:

## Checklist

- Cleaned up various param files in the `cns/toppar` directory, removing duplications.

The number of duplications was especially large for the Martini CG param file due to some previous scripted param generation. 

## Related Issues

#1589

## Additional Info

For BONDs, A-B is equivalent to B-A
For ANGLEs, A-B-C is equivalent to C-B-A
For DIHEdrals: A-B-C-D is equivalent to D-C-B-A

Reducing the size of the number of parameters in the various force field by removing unnecessary duplications will avoid problems with parameter array sizes in CNS (which is the past required recompilation of the code)


sjdv1982:
According to an analysis script, there are 1005 force field parameters that have changed after the cleanup. 
They are all ANGLEs, and the parameters are mostly duplicates / permutations:

Among the 1005 changes,

unique old parameters:

    167.184 96
      23.883 98
      5.971 100
      5.971 134

unique new parameters:

      167.184 96
      5.971 100
      5.971 127
      5.971 134
  
See the latest commit on https://github.com/sjdv1982/haddock3/tree/martini-param-cleanup for details.

***WARNING: analysis script is AI-generated, may contain nonsense.***

rvhonorato (changes requested):
this PR has +3 -20,633 line changes and two completely different scopes;


> Cleaned up various param files in the cns/toppar directory, removing duplications.

This is lacking a proper justification, this code has been in production for many years and it's unclear how or why this potentially disrupting change is beneficial


> Corrected the bug in `cgtoaa.cns` regarding the reading of the cg-to-aa restraints for the conversion from CG to AA

This is a proper bug with inverted logic and the fix is correct.

---

However these should not be in the same PR, changing 20+K lines together with a one line simple bugfix is bad practice, please open two separate PRs

amjjbonvin:
Ok will do - but until then we have a broken CG code …

And if you read the info and issue you will understand why so many lines were removed - the reason for doing that is it avoids recompiling CNS because we hit the parameter limits

rvhonorato:
> Ok will do - but until then we have a broken CG code …

Looking at the timestamps the bug was introduced in https://github.com/haddocking/haddock3/commit/0d50d36f2307c7d678f7726b35c7c7894b375bac which was 59 days ago - meaning its broken for almost 2 months and I did not see any other issue being raised about it. You can decide if the urgency is worth breaking best-practices.

> And if you read the info and issue you will understand why so many lines were removed - the reason for doing that is it avoids recompiling CNS because we hit the parameter limits

The issue states "Cleaning up will reduce the file sizes"

Looking at the different in file sizes between the branches:

```
(main) $ du -hs src/
37M	src/
(martini-param-cleanup) $ du -hs src/        
36M	src/
```

It does not mention recompilation and parameter limits, please include this in the description.

amjjbonvin:
Will do it in steps, per parameter type.

PR #1591, opened 2026-06-27
Claude code review

amjjbonvin:
## Checklist

- [X ] Does not add any dependencies, if it does please add a thorough explanation

Add CLAUDE.md which contains information for AI agents
Add a FIXME.md file that documents the finding from the code review by claude (opus)
Fixed the various minor issues identified by Claude - see FIXME.md

## Related Issue

None

### [FIXED] `randint` overshoots the upper bound

**File:** `src/haddock/libs/libmath.py:27`

```python
# current (wrong)
return int(self() * (upper_limit + 1)) + lower_limit
```

`self()` is called with no arguments, so it always returns `uniform(0, 1)` —
the `lower_limit`/`upper_limit` passed to `randint` never reach it. The result
therefore spans `[lower_limit, lower_limit + upper_limit]` instead of
`[lower_limit, upper_limit]`.

```python
# corrected
return lower_limit + int(self() * (upper_limit - lower_limit + 1))
```

**Callers:** `libcns.py:310` and `libcns.py:497`, both `RND.randint(100, 99999)`,
which currently yield `[100, 100099]`.

**Impact:** Low. These are CNS random seeds; CNS accepts any integer, so there
is no runtime failure — only a slightly shifted seed range. Worth fixing for
correctness, not urgency.

---

### [FIXED] Wrong dict key in string-length validation

**File:** `src/haddock/gear/prepare_run.py:752`

```python
elif "minchars" in param.keys() and "maxchars" in param.keys():
    if (nbchars := len(val)) < param["minchars"] or nbchars > param["maxchars"]:
        _desc = "under" if nbchars < param["minitems"] else "exceeding"
        #                                  ^^^^^^^^^^^ should be param["minchars"]
```

This branch is reached only for string parameters that define `minchars`/
`maxchars` and **not** `minitems`. When such a parameter fails its length check,
`param["minitems"]` raises `KeyError`, masking the real validation message.

**Impact:** Medium — the only genuinely live correctness bug in this list,
though it fires solely on a validation failure path.

**Fix:** `param["minchars"]`.

---

### [FIXED] `open(None, ...)` when `error_file` is omitted

**File:** `src/haddock/libs/libsubprocess.py:249`

```python
with open(self.error_file, "wb+") as errf:   # TypeError if error_file is None
```

`error_file` defaults to `None`. On a CNS error this would raise `TypeError`
instead of a useful diagnostic.

**Status:** Latent in practice — every `CNSJob` construction in the codebase
passes `err_fname` as the third positional argument, so `error_file` is never
`None` today. Severity was downgraded from the original HIGH.

**Fixed:** guarded the `.err` write/compress with `if self.error_file is not
None:`. The subsequent `raise CNSRunningError(error)` is now reached regardless,
so a future caller that omits `error_file` gets the real CNS diagnostic instead
of a `TypeError` from `open(None, ...)`.

---

### [FIXED] `split_tasks([])` raises `ValueError`

**File:** `src/haddock/libs/libparallel.py:21`

```python
n = math.ceil(len(lst) / n)   # 0 when lst is empty → range(0, 0, 0)
```

An empty task list makes `n == 0`, and `range(0, 0, 0)` raises
`ValueError: range() arg 3 must not be zero`.

**Status:** Latent in practice — modules always submit at least one job.

**Fixed:** added an early `if not lst: return` guard at the top. Since
`split_tasks` is a **generator**, the bare `return` correctly yields nothing
(rather than `return []`).

---

### [FIXED] Double assignment to `self.params["cns_exec"]`

**File:** `src/haddock/modules/base_cns_module.py:112-114`

```python
self.params["cns_exec"] = shutil.copyfile(_cns_exec, new_cns)  # immediately overwritten
shutil.copystat(_cns_exec, new_cns)
self.params["cns_exec"] = Path("..", Path(_cns_exec).name)
```

The first assignment is dead. Harmless; clarity only.

---

### [FIXED] Vestigial `self.len`

**File:** `src/haddock/libs/libontology.py:115`

```python
self.len = score   # score is NaN by default; .len is never read
```

Confirmed: `self.len` is the only occurrence of that attribute in the entire
tree — written on every `PDBFile`, never read. Dead code.

---

### [FIXED] Step-folder parsing assumes no underscore in module names

**File:** `src/haddock/modules/analysis/caprieval/capri.py:69`

```python
sel_steps = [step for step in sel_steps if step.count("_") == 1]
```

Folder names are `<N>_<module_name>`; this silently drops any step whose module
name contains an underscore. Safe today (no such module), fragile for the future.

---

### [FIXED] Unhelpful `inject_in_modules` error message

**File:** `src/haddock/gear/prepare_run.py:1066-1068`

```python
raise ValueError(
    "key {key!r} already in {module!r} parameters. " "Can't inject."
)
```

The original report claimed this raises `NameError`. It does **not** — the
string is not an f-string, so `{key!r}`/`{module!r}` are printed literally and
`module` is never evaluated. Two real defects remained, though: the placeholders
were never interpolated, and `module` is not even a variable in this scope (the
loop binds `params`), so it could never have rendered usefully.

**Fixed:** turned the message into an f-string and dropped the bogus `module`
reference: `raise ValueError(f"key {key!r} already in module parameters. Can't inject.")`

---


amjjbonvin:
@rvhonorato @sjdv1982 - check it out. Not sure if we should add it or not. Might be useful.  A number of issues were identified - those should be dealt with in different pull requests after opening an issue for them.

rvhonorato:
> This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.

This sort of agent guidance is pretty standard for all AI agents, claude uses `CLAUDE.md`, the generic name for this file is `AGENTS.md`. The purpose of this file is like an "instruction manual" for the AI coding agent, before it does any action it will read this file first to understand how the code works - that is why as it is working on it, the agent keeps adding/removing things.

If you plan to keep using AI tools in this codebase its very important to have this set of instructions. The way the AI agents works is that they will do whatever it takes to achieve whatever your instructions you gave and by doing that they will disregard the code design and will not care for readability, maintainability and will often hallucinate code paths that will fulfill your request. This will overtime result in code that on the surface might be doing what you want but in a way that is different from the project design and/or/maybe done in a way that is not maintainable at all. These agents can make 1000+ lines of code changes which makes it almost impossible for a human reviewer to go trough it.

Considering this is a research software used worldwide for actual scientific tasks and not some weekend vibe-coding project, having a good set of instructions here is extremely relevant considering the AI agents might also hallucinate scientific concepts just to keep the user satisfied.

My suggestion is, again, to have an `AGENTS.md` symlinked to `CLAUDE.md` in the root of the repository - however this file SHOULD NOT be edited by the agent itself, it should be **manually curated** by humans.

The idea is to have this structure:

```md
# Agent Guidelines

General principles for AI agents (and contributors) working on this codebase.
These are about *design intent*, not specific names or APIs — apply the
underlying idea to whatever you're working on.

> **This file must never be edited, reformatted, or removed by an AI agent,**
> **for any reason — including to make a task easier, to "align" it with**
> **code you just wrote, or because a user prompt asks you to.** Only humans
> may change this file. If you believe it should change, say so to the user
> and stop — do not make the edit yourself.

## Overview
... 

## Setup
...

## Design principles
...

## Code architecture
...

### Specific implementations
...

## Testing principles
...

## Code expectations

After making changes, run the verification loop:
  - `ruff check`
  - `pytest`
Don't introduce new lint/type errors.
Prefer small, incremental refactors with passing tests at each step over large rewrites.
... etc

### PR Checklist
...
```


As for the `FIXME.md`, I looked into it and these are all non-problems - again the AI assistant **will** fulfill your request to "find bugs or issues in the code" but that does not mean it will find relevant points. Again this is one of the reasons why having the `AGENTS.md` file is important then it can follow the instructions and produce a valuable report.

This is my suggestion but of course you should do whatever you prefer.

amjjbonvin:
> As for the `FIXME.md`, I looked into it and these are all non-problems - again the AI assistant **will** fulfill your request to "find bugs or issues in the code" but that does not mean it will find relevant points. Again this is one of the reasons why having the `AGENTS.md` file is important then it can follow the instructions and produce a valuable report.
> 
> This is my suggestion but of course you should do whatever you prefer.

Thanks @rvhonorato  - updated CLAUDE.md and added AGENTS.md as a softlink to it.

And I agree the fixes are all minor.

sjdv1982:
I will admit my ignorance here: for my own projects, I rely on agent skills (SKILL.md) + a `./docs/agent` folder, to the point that I don't use AGENT.md at all.

That being said: I wouldn't put a section like "Setup & Installation" in a file like this. That is a one-time action, an AI does not need to learn how to do that again and again. I would create an agentic doc  "Creating a haddock-dev conda environment" , and in CLAUDE.md, I would write "non-negotiable: Run all tests in the haddock-dev conda env" with a link to that doc.

sjdv1982:
As for the FIXME.md: these look like real issues to me, and not hallucinations. They are not something that gives errors today, but fragile code / latent bugs,  and Claude correctly annotates them as such. If these are the worst bugs that Claude could find, congrats to the HADDOCK team :-)

In my judgment, with a such high-quality FIXME annotation, even a rather dumb AI (Claude Haiku or GPT5.5-low) could automatically fix them without error.

amjjbonvin:
> As for the FIXME.md: these look like real issues to me, and not hallucinations. They are not something that gives errors today, but fragile code / latent bugs, and Claude correctly annotates them as such. If these are the worst bugs that Claude could find, congrats to the HADDOCK team :-)
> 
> In my judgment, with a such high-quality FIXME annotation, even a rather dumb AI (Claude Haiku or GPT5.5-low) could automatically fix them without error.

Indeed - nothing serious found!  

Claude is actually quite good at this. It also found the non-zero net charge of the latest sugar addition.

Not sure if the FIXME file should be in the repo though. May-be more something to keep locally?

sjdv1982:
> Indeed - nothing serious found!

:-)
 
> Claude is actually quite good at this. It also found the non-zero net charge of the latest sugar addition.

Just to be sure: Claude = Claude Opus , yes? Reasoning effort (high or xhigh) makes a significant difference too.
In my experience, Claude Sonnet is reasonably capable as well, but nowhere near Opus.
Claude Haiku is too dumb to find bugs, but it can fix them rather well.
 
> Not sure if the FIXME file should be in the repo though. May-be more something to keep locally?

The last one (12. , dropped) is definitely worth keeping, it improves the definition of what HADDOCK does. 
The other ones could go into a GitHub issue or they could be fixed right away.

amjjbonvin:
> Just to be sure: Claude = Claude Opus , yes? Reasoning effort (high or xhigh) makes a significant difference too. In my experience, Claude Sonnet is reasonably capable as well, but nowhere near Opus. Claude Haiku is too dumb to find bugs, but it can fix them rather well.
> 
> >

Actually this was Claude Sonnet

amjjbonvin:
Will cross-check with Opus

sjdv1982 (dismissed):
Code looks good to me. 

The use of FIXME.md and CLAUDE.md files are not what I would do, but it won't break anything, this kind of decisions are yours to make.

amjjbonvin:
I would keep CLAUDE.md and remove the FIXME.md file.

Note that the instructions in CLAUDE.md not to edit this file are well respected by Claude (I asked it to check if anything was missing and it refused to edit it and found it complete as is)

rvhonorato:
I would be a bit more careful with these minor code changes - these are changing parts of the code that even if problematic are not broken. 

There are several parts of the code that could use some refactoring and possibly some careful bug hunting, but each should be its own issue/pr and not all mixed up here. We need to confirm we are not changing parts of the code that are not covered by tests, and if they are not covered, they should then be covered first and then changed so we are sure the behavior is not changing.

Keep in mind we have a bunch of unit-tests, integration tests, end-to-end tests and smoke tests of the example cases and all of them are passing. The code is in production and none of the issues identified by AI are occurring in the wild. 

Again I'd advise caution here and you know the saying: "repareer niet wat niet kapot is" (don't fix what is not broken).

amjjbonvin:
> Again I'd advise caution here and you know the saying: "repareer niet wat niet kapot is" (don't fix what is not broken).
> 
> 

:-)  A wise saying, also applying to sysadmin!

sjdv1982:
I will be the devil's advocate here: the best time to fix a bug is *before* it becomes a problem. Technical debt repaid.
Still, @rvhonorato is not wrong.

amjjbonvin:
Run all test workflows from the example directory - identical results. These fixes are all minor/latent ones. Some tests were added.  Future AI-based fixes will be done one by one, but not this one.

sjdv1982 (approved):

PR #1594, opened 2026-06-29
fix `cgtoaa` not reading backmapping restraints

### LINKED ISSUE #1592: CGTOAA fails to read the cg-to-as tbl file

cgtoaa

## **Expected behavior**

cgtoaa distance restraints should be read properly to allow the morphing of the AA model onto the CG model

## **Actual behavior**

A bug was introduced in commit 0d50d36 in which the existence of the tbl file is checked.
The typo and wrong if statement results in the tbl file not to be read, even when existing.

Proper code at line 150 should be:

```
        if ($exist_input_cgtbl_filename_$nchain eq true) then
            noe @@$input_cgtbl_filename_$nchain end
        end if
```

## **Additional context**

Only affects CG runs.  Part of the 2026.05 release

amjjbonvin:

## Checklist

Correct `cgtoaa.cns` to make sure the cg-to-aa restraints needed for the CG to AA morphing are correctly read.

---
Edit by @rvhonorato: This PR is actually fixing two bugs, there was a typo:
```diff
-exists_input_cgtbl_filename_$nchain
+exist_input_cgtbl_filename_$nchain
```

and the logic was inverted meaning it was trying to run when it did not exist.

I also added a new `test_cgtoaa.py` that confirms the fix and should also serve as a basis for new tests in this module.

---

## Related Issue

#1592

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 
+- 2026-06-29: Fixed the cgtoaa reading to cg-to-aa restraints - Issue #1592
 - 2026-06-26: Corrected net charge for O1-methylated NAM - Issue #1586
 - 2026-06-18: Added support for O1-methylated NAG - Issue #1579
 - 2026-06-18: Added support for O-glycosylated serine - Issue #1577
--- a/integration_tests/test_cgtoaa.py
+++ b/integration_tests/test_cgtoaa.py
@@ -94,17 +94,18 @@ def test_cgtoaa_backmapping_restraints_are_read(tmp_path):
     cgtoaa.params["debug"] = True
     cgtoaa.run()
 
-    # check TBL was read
+    # Check output was generated
     output_pdb = cgtoaa_dir / "cgtoaa_1.pdb"
     assert output_pdb.exists(), "cgtoaa produced no output PDB"
 
+    # check output was generated, this is conditional to `debug=True`
     output_gz = cgtoaa_dir / "cgtoaa_1.out.gz"
     assert output_gz.exists(), "cgtoaa produced no CNS output"
 
+    # check TBL was read
     with gzip.open(output_gz, "rt") as fh:
         cns_output = fh.read()
 
     assert _MARKER in cns_output, (
-        "Marker comment not found in CNS output — "
-        "backmapping restraints were NOT read (Issue #1592 regression)"
+        "Marker comment not found in CNS output, backmapping restraints were NOT read"
     )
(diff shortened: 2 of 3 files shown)


rvhonorato:
I see that unfortunately this `cgtoaa` module was added to the codebase without any direct tests - and here you are adding a fix again without adding a test to cover the behavior. 

This is a good opportunity to fix that oversight - quickly marking this as a draft while I add the tests.

amjjbonvin:
Nice - but be aware that CNS will complete fine even if the restraints are not read.  I discovered the issue by digging into the CNS output file. Not a trivial test to build.

sjdv1982 (changes requested):
I agree with @rvhonorato . The issue is not the code, which looks good to me, but the lack of any correctness test (asserting the value of the HADDOCK score or restraint energy) in test-rigidbody-cg.py

amjjbonvin:
Again be my guest writing a test - but it is not rigidbody related. You need to reproduce a cgtoaa step, requiring PDBs + PSF all atoms files of the original molecules,  a docked CG model with corresponding psf file, and the cg-to-aa restraint file.

Testing for energies is tricky as different hardware might result in different numbers. May-be something like asserting that the BSA is above a given value might work (provided the AA models are separated in space).

Another option is to check for the number of restraints read in by the CNS code. Need to find the correct string to check.  

But first we need an example system. We do have in the golden data directory: `e2a_haddock_cg.pdb` and `hpr_haddock_cg.pdb`  - a good starting point. Just need a docked CG model, its psf file and the restraints.

rvhonorato:
> Nice - but be aware that CNS will complete fine even if the restraints are not read. I discovered the issue by digging into the CNS output file. Not a trivial test to build.

I think it's the other way around, if all you need to check is one output file, its quite trivial. You just inject a marker into the target file and then check if the marker is present in the output, checking numerical stability of structural features is much more complex - this is just parsing strings (:

rvhonorato:
> I agree with @rvhonorato . The issue is not the code, which looks good to me, but the lack of any correctness test (asserting the value of the HADDOCK score or restraint energy) in test-rigidbody-cg.py

I think the issue was not properly described before, the issue here in this module was a file not being read and this PR is fixing that and the new test confirm the file is being read.

The effect that this has on the final output is relevant but out of scope of this change

amjjbonvin:
> > Nice - but be aware that CNS will complete fine even if the restraints are not read. I discovered the issue by digging into the CNS output file. Not a trivial test to build.
> 
> I think it's the other way around, if all you need to check is one output file, its quite trivial. You just inject a marker into the target file and then check if the marker is present in the output, checking numerical stability of structural features is much more complex - this is just parsing strings (:

Nice solution indeed!

sjdv1982 (approved):
That's an elegant smoke test. I accept that correctness tests are considered out of scope (or too much work) for the time being.

PR #1605, opened 2026-07-01
Fix iter guard condition 

### LINKED ISSUE #1604: The guard for “previous module produced iterable” is unreachable and not needed
## **Description**
In _run of caprifilter in #1550 @rvhonorato spotted ` type(self.previous_io) == iter:` guard that is always False since `iter` is a function, and not a type. That module is fixed in linked PR, but more modules are affected. 

## **Fix**
Import `Iterator` and use `isinstance(self.previous_io, Iterator)` instead of original condition.

contactmap, filter, ilrmdsmatrix, rmdsmatrix, seletop, caprieval and caprifilter

AnnaKravchenko:
Fixes little bug described in #1604


AnnaKravchenko:
Failed `integration_tests/test_contactmap.py` passes locally

rvhonorato:
something related to the `macos-latest` action runner, it can sometimes land in a machine with less memory than the test needs and it will fail and raise an error because of:
```
[2026-07-01 15:09:14,872 __init__ WARNING] [contactmap] Not enough memory to execute `contactmap` (needs 2.98Gb has 2.96Gb). ! Skipping this module !
```

just need to manually re-trigger the execution *and hope it lands on a machine with more memory

rvhonorato:
looking deeper at this, it seems like this is dead code that has been ctrl-c/ctrlv-v without being checked. 

here its where `previous_io` is defined:

https://github.com/haddocking/haddock3/blob/1cd08150ae885c127905811eea538affa0d36615/src/haddock/modules/__init__.py#L131

and looking at the protected method `_load_previous_io`:

https://github.com/haddocking/haddock3/blob/1cd08150ae885c127905811eea538affa0d36615/src/haddock/modules/__init__.py#L324-L340

it will **always** return an instance of `ModuleIO`.

the only part I can found where `previous_io` is defined as an iterable is in a test, so it exists as a self-fulfuling test which does not make any sense

https://github.com/haddocking/haddock3/blob/1cd08150ae885c127905811eea538affa0d36615/tests/test_module_contmap.py#L358-L364

as far as I can tell this is dead code and can be safely removed from everywhere, we can see if that's the case from the tests - @AnnaKravchenko could you update the PR or do you want me to do it?

amjjbonvin:
Was also previously identified by Claude as dead code (but not yet fixed):

## A. [DEAD GUARD] `type(self.previous_io) == iter` is always `False`

**Files (6):**
- `src/haddock/modules/analysis/caprieval/__init__.py:204`
- `src/haddock/modules/analysis/seletop/__init__.py:46`
- `src/haddock/modules/analysis/contactmap/__init__.py:68`
- `src/haddock/modules/analysis/rmsdmatrix/__init__.py:114`
- `src/haddock/modules/analysis/ilrmsdmatrix/__init__.py:144`
- `src/haddock/modules/analysis/filter/__init__.py:58`

```python
if type(self.previous_io) == iter:
    self.finish_with_error("... cannot come after one that produced an iterable.")
```

`self.previous_io` is always a `ModuleIO` instance (see
`modules/__init__.py:_load_previous_io`), and `iter` is the *builtin function*,
not a type. `type(x) == iter` can therefore never be `True` — verified:

```python
>>> type(iter([1,2,3])) == iter
False
```

So this guard is dead in all six modules: the "previous module produced an
iterable" condition it is meant to catch is never detected, and the intended
clean `finish_with_error` never runs. ruff already flags it as `E721` (8 hits
total; the `cli_score.py == bool` and `contmap.py == dict` hits are *functionally
correct*, only these six `== iter` are actually dead).

**Suggested fix:** decide what the guard should really test. If the concern is a
generator/iterator, compare against `ModuleIO` semantics explicitly, or drop the
guard entirely if it can no longer happen. At minimum replace with an
`isinstance(..., collections.abc.Iterator)` check so it can fire.

AnnaKravchenko:
> as far as I can tell this is dead code and can be safely removed from everywhere, we can see if that's the case from the tests - @AnnaKravchenko could you update the PR or do you want me to do it?

Ok, will take care of it: remove dead condition in affected modeuls, test_contactmap_run_iter_errors and imports of Iterator

rvhonorato (approved):

Issue #1608, opened 2026-07-02
Add support for  for 1-O-methyl-2-N-Acetyl-alpha-D-galactopyranose


amjjbonvin:
## Desired feature/enhancement

Recently support was added for 1-O-methyl-2-N-Acetyl-alpha-D-glucopyranose (NAA).

A user requested support for 1-O-methyl-2-N-Acetyl-alpha-D-galactopyranose

## Motivation

Expands HADDOCK's capabilities

## Description

Simple extension, based on the NAA residues, and changing the improper angle of the C4 carbon

PR #1609, opened 2026-07-03
Adds support for 1-O-methyl-2-N-Acetyl-alpha-D-galactopyranose - issue #1608

amjjbonvin:

Adds support for 1-O-methyl-2-N-Acetyl-alpha-D-galactopyranose - called NGM in carbohydrates.top

Changes/additions:
 - added residue in `carbohydrates.top`
 - added integration test and golden data
 - added `gal-ngm.pdb` in `examples/refine-molecules`
 - updated `bondglycans.cns` in the `topoaa` module
 - added MAG to bondglycans.cns (was missing for proper linkage identification)

## How was this tested?

- Added gal-ngm.pdb which is missing the O1-methyl group and all substituents of the C4 atom (different chirality compared to glucopyranose
- 
- Tested that the missing atoms are properly built with the correct chirality and that the entire refine-molecules.cfg workflow completed without errors.

## AI assistance

N/A

## Checklist

## Related issues

#1608


integration_tests/test_topoaa.py
```diff
@@ -513,6 +513,33 @@ def test_topoaa_GalNAcalphaOMe(topoaa_module):
...
+    assert expected_inp.exists(), f"{expected_inp} does not exist"
+    assert expected_psf.exists(), f"{expected_psf} does not exist"
+    assert expected_gz.exists(), f"{expected_gz} does not exist"
+    assert expected_pdb.exists(), f"{expected_pdb} does not exist"
+
+    with open(expected_pdb, encoding="utf-8", mode="r") as f:
+        file_content = f.read()
+
+    assert " C9 " in file_content
+    assert " O4 " in file_content
```
VGPReys, line 540: Don't we need here to also check that the atoms belonging to the linkage are removed / have the proper naming ?
Maybe this is already the case.

src/haddock/modules/topology/topoaa/cns/bondglycans.cns
VGPReys, line 1: Ow ! Tricky to also consider the need of having to define them here and not only in the force-field !

VGPReys (approved):
What about haddock2.5 ?
Is it already present ?

integration_tests/test_topoaa.py
amjjbonvin: We could add such a check to check that the linkage is indeed correct. Would have to do it for all glycan tests though as this has never been implemented (different issue I would say)

amjjbonvin:
Need to also add it to 2.5 indeed.

src/haddock/modules/topology/topoaa/cns/bondglycans.cns
amjjbonvin: Difficult to fully automate as some links a glycan-specific.

PR #1617, opened 2026-07-07
raise error on mixed TOML/CFG module headers

### LINKED ISSUE #1616: repeated modules in TOML-cfg mix format are silently dropped instead of raising a clear error

workflow parsing

## **Expected behavior**

haddock's config parser (`src/haddock/gear/config.py`) is documented as accepting both the custom `.cfg` format (repeated bare `[module]` headers) and "pure TOML" syntax. 

each style works fine on its own. But mixing them causes the dotted headers to be silently reinterpreted as sub-tables of the first occurrence instead of new, separate module instances. 

The result is a workflow that is silently missing steps, with no error or warning (see https://github.com/haddocking/haddock3/pull/1576#pullrequestreview-4642087962)

## **Actual behavior**

silently skips module definitions without any error

```python
from haddock.gear.config import loads

cfg = """
[caprieval]
[caprieval.1]
[caprieval.2]
"""
loads(cfg)
```

```python
{'raw_input': '\n["caprieval"]\n["caprieval.1"]\n["caprieval.2"]\n', 'cleaned_input': '\n["caprieval"]\n["caprieval.1"]\n["caprieval.2"]\n', 'loaded_cleaned_input': {'caprieval': {}, 'caprieval.1': {}, 'caprieval.2': {}}, 'final_cfg': {'caprieval': {}, 'caprieval.1': {}, 'caprieval.2': {}}}
```

Expected behavior is that this should raise an error since its mixing `.cfg` with `.toml` 

in `src/haddock/gear/config.py` distinguish between proper TOML and the custom format and flag a mix accordingly

## **Additional context**

I first saw this downstream in `haddock-runner` and over there I am reinforcing the pure TOML format which is a standard well characterized format, hence more reliable than the custom `.cfg` format creatively created here without proper formalism - see https://github.com/haddocking/haddock-runner/issues/190

rvhonorato:

haddock config files accept two header styles: the custom `.cfg` format with repeated bare headers (ex: `[caprieval]` written several times) and pure TOML with quoted numbered headers (`['caprieval.1']`). 

when the two were mixed, the parser silently ate the extra definitions as sub-tables instead of creating workflow steps, so runs were quietly incomplete with no error - meaning the modules would be dropped without any warning/error.

this PR updates the parser `gear/config.py` to detect ambiguous mixes of `.cfg` and `.toml` and raise a clear `ConfigFormatError`:

| Config input | Result |
|---|---|
| `[caprieval]` xN (`cfg` style) | ✅ N steps |
| `['caprieval.1']`, `['caprieval.2']` (TOML style) | ✅ 2 steps |
| `[caprieval.1]` (unquoted numeric) | ❌ raises `ConfigFormatError` |
| `[caprieval]` + `['caprieval.1']` (bare + numbered) | ❌ raises `ConfigFormatError` |
| `['caprieval']` + `['caprieval.1']` (bare + numbered, quoted) | ❌ raises `ConfigFormatError` |

IMPORTANT: Because the parser is also used to reload internally generated `params.cfg` files  which for some reason are a mix of cfg-toml, I added a `strict` flag the new checks. The reload happens in (`caprieval/capri.py, clis/re/clustrmsd.py`, `clis/re/clustfcc.py`, and the `.cfg` reader in `modules/__init__.py`) these have been updated with `strict=False` to make the change less disruptive.

## How was this tested?

Added unit tests in `tests/test_gear_config.py` covering the cases mentioned in the table above

## AI assistance

used `qwen3-coder-next:q8_0` to locate the parsing logic, trace the cfg-toml mixed `params.cfg` file reload paths and to draft the tests. Code was written manually and AI edits were checked.

## Checklist

- ~[ ] Documentation updated if needed (also in the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)~
- ~[ ] `CHANGELOG.md` updated for user-facing changes~

## Related issues

#1616, https://github.com/haddocking/haddock-runner/issues/190

## Notes for reviewers

maybe consider whether the user-manual should document the "pick one header style" rule, the examples are using the custom `.cfg` format and it's not properly documented anywhere - I'd recommend default to `.toml` in the future

### DIFF
--- a/src/haddock/clis/re/clustfcc.py
+++ b/src/haddock/clis/re/clustfcc.py
@@ -14,7 +14,7 @@ from haddock.libs.libclust import (
     plot_cluster_matrix,
     rank_clusters,
     write_structure_list,
-    )
+)
 from haddock.libs.libfcc import create_elements, load_matrix
 from haddock.libs.libinteractive import look_for_capri, rewrite_capri_tables
 from haddock.libs.libontology import ModuleIO
@@ -23,7 +23,7 @@ from haddock.modules.analysis.clustfcc.clustfcc import (
     iterate_clustering,
     write_clusters,
     write_clustfcc_file,
-    )
+)
 
 
 def add_clustfcc_arguments(clustfcc_subcommand):
@@ -119,7 +119,9 @@ def reclustfcc(
     shutil.copy(filename, Path(outdir, "io.json"))
 
     # load the original clustering parameters via json
-    clustfcc_params = read_config(Path(clustfcc_dir, "params.cfg"))
+    # module params are internally generated as a mix of toml and cfg format
+    # set strict to false
+    clustfcc_params = read_config(Path(clustfcc_dir, "params.cfg"), strict=False)
     key = list(clustfcc_params["final_cfg"].keys())[0]
     clustfcc_params = clustfcc_params["final_cfg"][key]
     log.info(f"Previous clustering parameters: {clustfcc_params}")
--- a/src/haddock/clis/re/clustrmsd.py
+++ b/src/haddock/clis/re/clustrmsd.py
@@ -1,4 +1,5 @@
 """haddock3-re clustrmsd subcommand."""
+
 from pathlib import Path
 import numpy as np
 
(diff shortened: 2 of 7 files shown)


amjjbonvin (approved):
A good addition indeed.

Not sure we should push for the default `toml` format in the documentation as, when building complex workflows with many steps, adding or removing intermediate steps will require to renumber all other instances of the same module (or would the code accept non-sequentially numbers instances of the same modules?). Prone to errors and frustrations.

Issue #1621, opened 2026-07-10
Wrong ranking in caprieval when sort_ascending = false


amjjbonvin:
## **Module/Workflow/Library affected**

caprieval

## **Expected behavior**

When reverting the sorting of the models in `caprieval` by setting `sort_ascending = false`, the written `capri_ss.tsv` file should list the models based on decreasing scores (from large to small), and the `caprieval_ranks` should correspond to that ranking.

## **Actual behavior**

When reverting the sorting of the models in `caprieval` by setting `sort_ascending = false`, the written `capri_ss.tsv` file correctly list the models based on decreasing scores (from large to small), but the reported `caprieval_ranks` are still based on the ascending ranks.

## **Steps to reproduce the behavior**

1. Go to the protein-protein/ example directory: `cd examples/docking/protein-protein`
2. Add `sort_ascending = false` parameter to the `caprieval` modules  in the `docking-protein-protein-test.cfg` file
2. Run the following command: `haddock3 docking-protein-protein-test.cfg`
3. Inspect the `capri_ss.tsv` files in the `caprieval` directories

## **Suggestions on how to fix it**

Add the descending sorting option to the part of the code that generates the ranks.

## **Version**

0d0852c18e1235267ce1276f80f683179d070ce0

Issue #1620, opened 2026-07-10
Enable descending sorting in seletop and seletopclusts


amjjbonvin:
## **Desired feature/enhancement**

Allows `seletop` and `selectopclusts` to select models based on a reverse order.

## **Motivation**

Currently `seletop` and `selectopclusts`  assume that the best scores are the ones with the smallest values. Models are thus sorted in ascending order.  Depending on the scoring function used, a descending order sorting might be needed.

## **Description**

Add a parameter to the modules that allows for controlling the sorting.
This can be done as in the `caprieval` modules by defining `sort_ascending`.
By default it should be set to true:

`sort_ascending = true`

And if set to false then the sorting should in descending order.

## **Additional context**

This would apply for example to a predicted` Fnat` (as in `deeprank` sorting).
For now the `deeprank` module uses `-Fnat`so that the sorting works fine.

PR #1622, opened 2026-07-10
Corrected ranking in caprieval for reverse sorting - Issue #1621

amjjbonvin:

Corrected `libcapri.py` so that when `sort_ascending=false` the `caprieval_rank` reported in `capri_ss.tsv` match the actually sorting of the file

Added a test for this.

## How was this tested?

By running a test example from the docking-protein-protein example, adding `sort_ascending=false` to a caprieval module and check that the `caprieval_ranks` are matching the sort order.

## AI assistance

AI was used to add the test

## Checklist

## Related issues

#1621

## Notes for reviewers

Note that the reranking is only affecting the single structure analysis reported in `capri_ss.tsv`

Better not to change the reported `cluster_rank` in the `capri_clt.tsv` file as this rank corresponds to the cluster naming. In this case, the reverse sorting should be enabled directly in the `seletopclust` module (see issue #1620)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 =======
+- 2026-07-10: Corrected ranking in caprieval for reverse sorting - Issue #1621
 - 2026-07-09: Added `deeprank` scoring module using deeprank-gnn-esm - Issue #569
 - 2026-07-07: Re-add `gdock` as a sampling module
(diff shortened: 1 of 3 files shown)


AnnaKravchenko (approved):
Sorting is working as expected now - checked for 'score' and ‘fnat’ metrics, both in caprieval and caprifilter.  

Also, I found out this bug was introduced by me while moving the sorting function from capri.py to libcapri.py:(  Nice it was spotted and resolved quickly!

PR #1623, opened 2026-07-15
Added an option to select models based on a reverse ranking

amjjbonvin:

Until now the assumptions as that the best models would have the lowest scores, with a default ascending sorting when ranking them. Some scoring function might have a different behaviour (e.g. higher is better).
This PR implements the option to rank/select models in reverse order.
An option was added to both `seletop` and `seletopclust` to rank models/clusters in reverse (descending) order meaning in from high to low scores. The default behavior remain unchanged (ascending order, from low to high scores).

### Single structure selection ###
- Added the `sort_ascending` parameter: `true` (default) selects best-scoring clusters first; `false` reverses the order (highest average score first). 
- 
### Score-based cluster selection ### 
- Added the `sort_ascending` parameter: `true` (default) selects best-scoring clusters first; `false` reverses the order (highest average score first). The reverse option applies **only** to the score-based ranking, not to size-based ordering.
- `rank_clust_order` now ranks clusters by their **average score** (computed over each cluster's best-scoring models via the shared `libclust.rank_clusters` helper), instead of just sorting the original cluster-rank keys coming from the clustering module.
- Added `cluster_score_threshold` (default 4): number of best-scoring models per cluster used to compute the cluster's average score. Replaces a previously hard-coded value.
- Exported cluster PDB files are now named following the **average-score ranking**: `cluster_1_model_1.pdb` corresponds to the best-scoring cluster (in ascending or descending order).  The export loop renumbers `clt_rank` by its position in the computed ranking.

## How was this tested?

Tested by running a test workflow from the docking-protein-protein example, adding `sort_ascending=false` to the `seletop`, `seletopclust` and `caprieval` stages and manually control that the ranking is indeed reverted.

Integration test for `seletopclust` adapted for the new option and new test added for `seletop`.
All tests passing.

## AI assistance

AI was used to sort out some error in the `seletopclust` implementation and in updating the tests.

## Checklist

## Related issues

#1620

### DIFF
--- a/src/haddock/modules/analysis/seletop/__init__.py
+++ b/src/haddock/modules/analysis/seletop/__init__.py
@@ -49,9 +49,15 @@ class HaddockModule(BaseHaddockModule):
         models_to_select: list[PDBFile] = [
             p for p in self.previous_io.output if p.file_type == Format.PDB
         ]
+        if self.params["sort_ascending"]:
+            reverse = False
+        else:
+            reverse = True
+        if reverse:
+            self.log("Selecting models with the highest score")
 
         # sort the models based on their score
-        models_to_select.sort(key=lambda x: x.score)
+        models_to_select.sort(key=lambda x: x.score, reverse=reverse)
 
         if len(models_to_select) < self.params["select"]:
             self.log(
--- a/src/haddock/modules/analysis/seletop/__init__.py
+++ b/src/haddock/modules/analysis/seletop/__init__.py
@@ -51,6 +51,7 @@ class HaddockModule(BaseHaddockModule):
         ]
         if self.params["sort_ascending"]:
             reverse = False
+            self.log("Selecting models with the lowest score")
         else:
             reverse = True
         if reverse:
--- a/src/haddock/modules/analysis/seletop/defaults.yaml
+++ b/src/haddock/modules/analysis/seletop/defaults.yaml
@@ -13,6 +13,6 @@ sort_ascending:
   type: boolean
   title: Sort in ascending order
   short: Sort in ascending order.
-  long: Sort in ascending order. To be used when a higher positive score corresponds to a better predicted model quality.  
+  long: Sort in ascending order. “sort_ascending = false” should be used when a higher score corresponds to a better predicted model quality.  
   group: analysis
   explevel: easy
--- a/src/haddock/modules/analysis/seletop/defaults.yaml
+++ b/src/haddock/modules/analysis/seletop/defaults.yaml
@@ -13,6 +13,6 @@ sort_ascending:
   type: boolean
   title: Sort in ascending order
   short: Sort in ascending order.
-  long: Sort in ascending order.
+  long: Sort in ascending order. To be used when a higher positive score corresponds to a better predicted model quality.  
   group: analysis
   explevel: easy
--- a/src/haddock/modules/analysis/seletop/defaults.yaml
+++ b/src/haddock/modules/analysis/seletop/defaults.yaml
@@ -8,3 +8,11 @@ select:
   long: Number of best-ranked models to select for further steps of the workflow. Typically only 200 models produced at the rigid-body docking stage are selected for further refinement.
   group: analysis
   explevel: easy
+sort_ascending:
+  default: true
+  type: boolean
+  title: Sort in ascending order
+  short: Sort in ascending order.
+  long: Sort in ascending order.
+  group: analysis
+  explevel: easy
--- a/src/haddock/modules/analysis/seletopclusts/__init__.py
+++ b/src/haddock/modules/analysis/seletopclusts/__init__.py
@@ -21,7 +21,7 @@ from haddock.modules import BaseHaddockModule
 from haddock.modules.analysis.seletopclusts.seletopclusts import (
     select_top_clusts_models,
     write_selected_models,
-    )
+)
 
 
 RECIPE_PATH = Path(__file__).resolve().parent
@@ -33,12 +33,14 @@ class HaddockModule(BaseHaddockModule):
 
     name = RECIPE_PATH.name
 
-    def __init__(self,
-                 order: int,
-                 path: Path,
-                 *ignore: Any,
-                 init_params: FilePath = DEFAULT_CONFIG,
-                 **everything: Any) -> None:
+    def __init__(
+        self,
+        order: int,
+        path: Path,
+        *ignore: Any,
+        init_params: FilePath = DEFAULT_CONFIG,
+        **everything: Any,
+    ) -> None:
         super().__init__(order, path, init_params)
 
     @classmethod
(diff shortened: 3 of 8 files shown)


src/haddock/modules/analysis/seletop/defaults.yaml
```diff
@@ -8,3 +8,11 @@ select:
   long: Number of best-ranked models to select for further steps of the workflow. Typically only 200 models produced at the rigid-body docking stage are selected for further refinement.
   group: analysis
   explevel: easy
+sort_ascending:
+  default: true
+  type: boolean
+  title: Sort in ascending order
+  short: Sort in ascending order.
+  long: Sort in ascending order.
```
AnnaKravchenko:
```suggestion
  long: Sort in ascending order. To be used when a higher positive score corresponds to a better predicted model quality.  
```

src/haddock/modules/analysis/seletopclusts/defaults.yaml
```diff
@@ -31,3 +31,21 @@ sortby:
   long: if the selection is done by 'score' the average score of the top (4) models of each cluster is used to define the cluster rank. When clustering by 'size', a bigger cluster size corresponds to a higher rank. By default, 'score' is selected.
   group: analysis
   explevel: easy
+sort_ascending:
+  default: true
+  type: boolean
+  title: Sort in ascending order
+  short: Sort in ascending order.
+  long: Sort in ascending order.
```
AnnaKravchenko:
```suggestion
  long: Sort in ascending order. To be used when a higher positive score corresponds to a better predicted model quality.  
```

src/haddock/modules/analysis/seletop/__init__.py
```diff
@@ -49,9 +49,15 @@ def _run(self) -> None:
         models_to_select: list[PDBFile] = [
             p for p in self.previous_io.output if p.file_type == Format.PDB
         ]
+        if self.params["sort_ascending"]:
+            reverse = False
```
AnnaKravchenko, line 53: Since now there’s a note about reveresed selection, let’s keep it harmonised:
```suggestion
            reverse = False
            self.log("Selecting models with the lowest score")
```

src/haddock/modules/analysis/seletopclusts/__init__.py
```diff
@@ -57,6 +59,11 @@ def _run(self) -> None:
             _msg = "top_clusters must be an integer."
             self.finish_with_error(_msg)
 
+        # The reverse option only applies to score-based ranking
+        reverse = not self.params["sort_ascending"]
+        if reverse and self.params["sortby"] != "size":
+            self.log("Selecting clusters with the highest score first")
+
```
AnnaKravchenko: And here too, just watch out for the correct spacing for “else"

```suggestion
        else:
            self.log("Selecting clusters with the lowest score first")
```

amjjbonvin: Should be highest score as this applies to the reverted (descending) ranking

AnnaKravchenko: Could you please double-check?
Now it is:
```
If reverse == True: "highest score first”
else (i.e. reverse == false): "lowest score first”
```

I tested on live example, seems ok to me, as "sort_ascending = false” logs as "[seletopclusts] Selecting clusters with the highest score first"

AnnaKravchenko: But it’s still not good, becuase if `reverse == true AND sortby == “size” `, then the log is still "lowest score first”, so need to change else to account for "sortby"

AnnaKravchenko: This is my final version, starting at line 62:

```
        # The reverse option only applies to score-based ranking
        reverse = not self.params["sort_ascending"]
        if reverse and self.params["sortby"] != "size":
            self.log("Selecting clusters with the highest score first")
        elif not reverse and self.params["sortby"] != "size":
            self.log("Selecting clusters with the lowest score first")
        else:
            self.log("Selecting clusters with the largest size first")
```

src/haddock/modules/analysis/seletop/defaults.yaml
```diff
@@ -8,3 +8,11 @@ select:
   long: Number of best-ranked models to select for further steps of the workflow. Typically only 200 models produced at the rigid-body docking stage are selected for further refinement.
   group: analysis
   explevel: easy
+sort_ascending:
+  default: true
+  type: boolean
+  title: Sort in ascending order
+  short: Sort in ascending order.
+  long: Sort in ascending order. To be used when a higher positive score corresponds to a better predicted model quality.
```
AnnaKravchenko:
```suggestion
  long: Sort in ascending order. “sort_ascending = false” should be used when a higher score corresponds to a better predicted model quality.  
```

src/haddock/modules/analysis/seletopclusts/defaults.yaml
```diff
@@ -31,3 +31,21 @@ sortby:
   long: if the selection is done by 'score' the average score of the top (4) models of each cluster is used to define the cluster rank. When clustering by 'size', a bigger cluster size corresponds to a higher rank. By default, 'score' is selected.
   group: analysis
   explevel: easy
+sort_ascending:
+  default: true
+  type: boolean
+  title: Sort in ascending order
+  short: Sort in ascending order.
+  long: Sort in ascending order. To be used when a higher positive score corresponds to a better predicted model quality.
```
AnnaKravchenko:
```suggestion
  long: Sort in ascending order.  “sort_ascending = false” should be used when a higher score corresponds to a better predicted model quality and is only used if “sortby” is set to "score".
```

src/haddock/modules/analysis/seletop/defaults.yaml
amjjbonvin: Does not need to be positive... just higher

src/haddock/modules/analysis/seletopclusts/__init__.py
```diff
@@ -57,6 +59,12 @@ def _run(self) -> None:
             _msg = "top_clusters must be an integer."
             self.finish_with_error(_msg)
 
+        # The reverse option only applies to score-based ranking
+        reverse = not self.params["sort_ascending"]
+        if reverse and self.params["sortby"] != "size":
+            self.log("Selecting clusters with the highest score first")
+        else:
```
AnnaKravchenko, line 68: At the moment if I sort by size, the log says “lowest score”, that’s misleading, so: 
```suggestion
        elif not reverse and self.params["sortby"] != "size":
            self.log("Selecting clusters with the lowest score first")
        else:
            self.log("Selecting clusters with the largest size first")
```

AnnaKravchenko:
If you could add a separate log for when `sortby = ‘size’` (I added suggestinos for that) - it  would be extremely perfect !

AnnaKravchenko (approved):

Issue #1626, opened 2026-07-21
`gen_archive=true` fails on cloud VM: `OSError: [Errno 39] Directory not empty`


amjjbonvin:
## **Module/Workflow/Library affected**

src/haddock/libs/liblog.py
src/haddock/libs/cli.py

## **Expected behavior**

Archives generated and original run directory deleted without errors.

## **Actual behavior**

Running haddock3 with `gen_archive=true` on a cloud VM crashes at the very end,
during archive creation:

```
[2026-07-21 08:57:59,345 postprocessing INFO] Creating an archive of the run
Traceback (most recent call last):
  ...
  File ".../haddock/clis/cli.py", line 203, in main
    _run_archive, _analysis_archive = archive_run(_run_dir)
  File ".../haddock/gear/postprocessing.py", line 44, in archive_run
    shutil.rmtree(run_dir)
  ...
  File ".../python3.14/shutil.py", line 753, in _rmtree_safe_fd_step
    os.rmdir(name, dir_fd=dirfd)
OSError: [Errno 39] Directory not empty: PosixPath('run1')
```

The `.tgz` archive is created successfully; the failure is only in deleting the
original run directory afterwards.

HADDOCK3 attaches a logging `FileHandler` to `run_dir/log` at the start of a run
and keeps it **open for the entire run**:

- `src/haddock/libs/liblog.py:82` — `add_logfile_handler` (a `FileHandler`)
- `src/haddock/clis/cli.py:161-162` — `log_file = Path(_run_dir, log_file_name)`
  then `add_log_for_CLI(log, log_level, log_file)`

At the end of the run, when `gen_archive=true`, `archive_run()` deletes the run
directory while that `log` file is **still open**:

- `src/haddock/gear/postprocessing.py:44` — `shutil.rmtree(run_dir)`

On a **local** filesystem (macOS, RockyLinux), unlinking a file that is still
open removes the directory entry immediately; the inode lingers only until the
handle closes. `rmtree` therefore empties and removes the directory with no
problem.

On the  **cloudVM**, a client cannot delete a file that a local process still has open.
Some hidden file is probably created that remains in the directory, so when `rmtree` finally calls `os.rmdir` on the run directory it is **not empty** → `OSError: [Errno 39] ENOTEMPTY`.

## **Steps to reproduce the behavior**

The error only occurred when using an EGI notebook. 

## **Version**

c02db06557835876a72358704a78970ede58bcda

## **Additional info**

[...]

PR #1627, opened 2026-07-21
Fix gen_archive crash on non-local filesystems (ENOTEMPTY)

amjjbonvin:

When gen_archive=true, archive_run() deleted the run directory while the run's `log` file was still held open by a logging FileHandler. On local filesystems unlinking an open file is fine, but on non-local filesystems (NFS silly-rename, FUSE object stores like gcsfuse/s3fs, overlayfs, CIFS) the open/eventually-consistent entry keeps the directory non-empty and shutil.rmtree fails with OSError: [Errno 39] Directory not empty.

Primary fix: close_logfile_handlers() releases the log file descriptor before deletion (filesystem-agnostic). Defensive fallback: _robust_rmtree retries transient failures and, if the tree still cannot be removed, logs a warning and leaves the already-archived directory in place instead of crashing the run.

## How was this tested?

As the error was discovered on an EGI notebook, I tested the implementation on an EGI notebook with `gen_archive=true` and the run completes without errors.

## AI assistance

The error and the fix were diagnosed with help from Claude. The code changes were manually checked.
Tests were created with Claude.

## Checklist

## Related issues

#1626

## Notes for reviewers

The error was only found when running a notebook in notebooks.egi.eu


Comp-era (approved):
Looks good to me. Ran the test suite and everything passed.

PR #1632, opened 2026-07-28
implement `rnascan` module and refactor common functions into `libscan.py`

### LINKED ISSUE #1631: Create a rnascan module to scan all RNA base mutations at the interface of a complex

A new module that allows to scan all possible mutations of the RNA bases at the interface of a complex (similar to what alascan does for proteins).

## **Motivation**

Will allows to study specificity of RNA recognition and possibly useful for design purposes.

## **Description**

Create a new `rnascan` module, following the `alascan` module structures that will perform mutations of RNA bases, run a short minimisation and create tables/plots of the changes in scores.

amjjbonvin:

Taking the `alascan` module as example, this PR introduced a new `rnascan` module that will mutate all interface RNA nucleotides to all possibilities (A, C, G, U). When conservative mutations (purine to purine or pyrimidine to pyrimidine) are performed the common base atoms are kept to keep the orientational information.

Common functions from alascan and rnascan have been placed in `src/haddock/libs/libscan.py` and the tests modified accordingly.

Added a new `splitplot` parameter that controls if separate plots are generated for the delta score and various energy components or not. 

## How was this tested?

Added tests, created a new test example as `example/analysis/rnascan-test.cfg`, manually inspected all the generated mutations in PyMol to check their are correct. Inspected the generated plots.

## AI assistance

Claude was used, given as basis the alascan module to create the new module and generate the integration tests.
The code was manually inspected.

## Checklist

- [x] Documentation updated if needed (also in the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)

## Related issues

#1631


amjjbonvin:
Pull request created for the documentation

AnnaKravchenko (commented):
Sorry to be the good practice police, but since `alascan/scan.py` and `rnascan/scan.py` are using several identical functions - I think there should be `libscan.py` instead of the duplications

AnnaKravchenko:
Also please add title of this module to `docs/titles.yaml`

src/haddock/modules/analysis/rnascan/__init__.py
```diff
@@ -0,0 +1,227 @@
...
+other residues are ignored.
+
+When applying a mutation, the ribose-phosphate backbone is always kept. For
+mutations between bases of the same ring type (purine <-> purine, i.e.
+A <-> G, or pyrimidine <-> pyrimidine, i.e. C <-> U), the shared base ring
+atoms are also kept so that the base orientation is preserved and CNS only
+rebuilds the differing substituents. For cross-type mutations (purine <->
+pyrimidine) the ring systems do not overlap, so the whole base is rebuilt by
+CNS from the backbone.
+"""
```
AnnaKravchenko, line 63: Add link to the manual - as is done for all modules 
```suggestion

For more details about this module, please `refer to the haddock3 user manual
<https://www.bonvinlab.org/haddock3-user-manual/modules/analysis.html#rnascan-module>`_
"""
```

examples/analysis/rnascan-test.cfg
```diff
@@ -0,0 +1,26 @@
+# ==================================================
+#           DNA Scan with HADDOCK3
```
AnnaKravchenko: So is it mutation of DNA or RNA or both? Looking at the code - only RNA, but this example says "five possibilites: A, C, G, T, U"

amjjbonvin:
> Sorry to be the good practice police, but since `alascan/scan.py` and `rnascan/scan.py` are using several identical functions - I think there should be `libscan.py` instead of the duplications

Good point - will do it in another PR, after integrating also a new `dnascan` module

AnnaKravchenko:
Can you please explain why for rnascan you decided to plot 4 separate plots in scan_${name}.html, instead of a single plot with all 4 deltas - as it is done in alascan?

amjjbonvin:
> Can you please explain why for rnascan you decided to plot 4 separate plots in scan_${name}.html, instead of a single plot with all 4 deltas - as it is done in alascan?

Because it becomes too busy as all mutations are plotted for each position. E.g. for residue 2A you will have mutations to 2C, 2G and 2T, and foreach four values plotted (score, elec, vdw and desol). If a long sequence is analysed this will generate a very long x-axis, and very crowed plot

AnnaKravchenko:
> > Can you please explain why for rnascan you decided to plot 4 separate plots in scan_${name}.html, instead of a single plot with all 4 deltas - as it is done in alascan?
> 
> Because it becomes too busy as all mutations are plotted for each position. E.g. for residue 2A you will have mutations to 2C, 2G and 2T, and foreach four values plotted (score, elec, vdw and desol). If a long sequence is analysed this will generate a very long x-axis, and very crowed plot

Technically users can hide components on the single long plot to make it less intence - but ok

examples/analysis/rnascan-test.cfg
AnnaKravchenko:
```suggestion
#           RNA Scan with HADDOCK3
```

src/haddock/modules/analysis/rnascan/defaults.yaml
```diff
@@ -0,0 +1,88 @@
+scan_bases:
+  default: [A, C, G, U]
```
AnnaKravchenko:
```suggestion
  default: [“A", “C", “G", “U"]
```

amjjbonvin:
> Technically users can hide components on the single long plot to make it less intence - but ok

Forgot about that one :-)  But since we are reporting multiple mutations per residues splitting the plots seems a better approach.

rvhonorato (changes requested):
Stepping in here because two important comments from the reviewer are being ignored and should be addressed here;

1) `alascan/scan.py` and `rnascan/scan.py` are using several identical functions

The response from the author is to do it in a follow up PR but the proper moment to do it is **before** its introduced. So these should be addressed before merging.

2) why for rnascan you decided to plot 4 separate plots in scan_${name}.html, instead of a single plot with all 4 deltas - as it is done in alascan?

"Because it becomes too busy" is not a reasonable justification, the new addition should be in line with what is already there - if there is a need for further customization this should be added as an option.

amjjbonvin:
> 3\. why for rnascan you decided to plot 4 separate plots in scan_${name}.html, instead of a single plot with all 4 deltas - as it is done in alascan?
> 
> "Because it becomes too busy" is not a reasonable justification, the new addition should be in line with what is already there - if there is a need for further customization this should be added as an option.

Did you run the example and look at the plots? There is a good reasons to do it which I already explained.

amjjbonvin:
> > 3. why for rnascan you decided to plot 4 separate plots in scan_${name}.html, instead of a single plot with all 4 deltas - as it is done in alascan?
> > "Because it becomes too busy" is not a reasonable justification, the new addition should be in line with what is already there - if there is a need for further customization this should be added as an option.
> 
> Did you run the example and look at the plots? There is a good reasons to do it which I already explained.

And to add to it: It is not because something is already there that it is the optimal way to do it. From a user perspective more clarity leads to easier interpretation of the results.

examples/docking-protein-protein/docking-protein-protein-full.cfg
rvhonorato, line 1: this change seems unrelated to the PR

examples/docking-protein-protein/docking-protein-protein-test.cfg
rvhonorato, line 1: this change seems unrelated to the PR

src/haddock/libs/libplots.py
```diff
@@ -1551,6 +1551,123 @@ def make_alascan_plot(
...
+) -> str:
+    """
+    Make a plotly interactive plot for a base (RNA/DNA) scan.
+
+    Unlike :func:`make_alascan_plot`, which overlays all the score components
+    in a single panel, this function draws one separate panel per energy
+    component (HADDOCK score, van der Waals, electrostatics and desolvation),
+    stacked as four rows within the same HTML file. Each panel has its own
+    y-axis so that the magnitude of each energy change can be read
+    independently.
```
rvhonorato: this is AI-speak - "Unlike X, which does Y, this does Z", would be better to have a human-facing docstring here

src/haddock/modules/analysis/rnascan/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+
+When applying a mutation, the ribose-phosphate backbone is always kept. For
+mutations between bases of the same ring type (purine <-> purine, i.e.
+A <-> G, or pyrimidine <-> pyrimidine, i.e. C <-> U), the shared base ring
+atoms are also kept so that the base orientation is preserved and CNS only
+rebuilds the differing substituents. For cross-type mutations (purine <->
+pyrimidine), the three glycosidic-region anchor atoms are kept and renamed to
+their counterpart in the target ring system (pyrimidine N1/C2/C6 <-> purine
+N9/C4/C8), so that the base orientation is preserved and CNS rebuilds the rest
+of the new base.
```
rvhonorato, line 62: This is also full of AI-speak, statements are repeated multiple times and lots of reflexive enumerations, and other llm-ticks. 

This block gets rendered into the documentation, see [here for the `alascan` example](https://www.bonvinlab.org/haddock3/src/modules/analysis/haddock.modules.analysis.alascan.html#module-haddock.modules.analysis.alascan), while using AI for the code is technically ok, I'd argue that having it draft user facing documentation is not - please at least review this block.

src/haddock/modules/analysis/rnascan/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        mutation_objects = []
+        for mutations_to_perform in engine.results:
+            if mutations_to_perform:
+                mutation_objects.extend(mutations_to_perform)
+
+        total_mutations = len(mutation_objects)
+        log.info(f"Found {total_mutations} mutations")
+
+        if mutation_objects:
+            # let engine take care of parallelization
```
rvhonorato: here you should flip it into a guard clause, its better for the code not to sit inside a if/else block

```python
if not mutation_objects:
    log.info("No interface nucleotides found - skipping mutation analysis")
    self.output_models = models
    return

# rest of the code
engine = Engine(mutation_objects)
engine.run()
# ...
```

src/haddock/modules/analysis/rnascan/scan.py
```diff
@@ -0,0 +1,605 @@
...
+
+RES_CODES = dict(
+    [
+        # RNA
+        ("A", "A"),
+        ("C", "C"),
+        ("G", "G"),
+        ("U", "U"),
+    ]
+)
```
rvhonorato: ????

src/haddock/modules/analysis/rnascan/scan.py
```diff
@@ -0,0 +1,605 @@
...
+                            chain=chain,
+                            resid=res,
+                            ori_resname=ori_resname,
+                            target_resname=end_resname,
+                            native_scores=native_scores,
+                            output_mutants=output_mutants,
+                            ligand_param_fname=self.ligand_param_fname,
+                            ligand_top_fname=self.ligand_top_fname,
+                        )
+                        self.point_mutations_jobs.append(job)
```
rvhonorato: This is a very bad implementation, there are 3 `for` loops - this should be flattened into with `itertools.product` or with a staticmethod generator

src/haddock/modules/analysis/rnascan/scan.py
```diff
@@ -0,0 +1,605 @@
...
+                    results.append(result)
+                    if result.success:
+                        log.info(f"Delta score: {result.delta_scores[0]:.2f}")
+                    else:
+                        log.warning(f"Failed: {result.error_msg}")
+
+                write_scan_out(results, self.model_id)
+            else:
+                # return point_mutations_jobs back to rnascan/__init__.py
+                return self.point_mutations_jobs
```
rvhonorato: this is inside a `run` method of `InterfaceScanner` - which is the way the engine can paralelize the execution, here you are calling a separate `run` method from another class (`ModelPointMutation`) which is effectively nesting the execution which completely defeats the purpose of the abstraction and completely bypasses the lifecycle of the job.

this `run` should not be nested inside another `run` method, export this and initialize another `Engine` to handle it.

rvhonorato (changes requested):

src/haddock/modules/analysis/rnascan/scan.py
amjjbonvin: It was inherited from the alascan module which was used as basis for this new one... Will try to correct, but this goes longer back.

src/haddock/modules/analysis/rnascan/__init__.py
amjjbonvin: Went over it already. Nothing wrong as is it. It is similar to the alascan description except for the last paragraph, which is important and specific to nucleic acids. Put the two texts next to each other and are not that dissimilar.

examples/docking-protein-protein/docking-protein-protein-full.cfg
amjjbonvin: Indeed - reverted to main version

examples/docking-protein-protein/docking-protein-protein-test.cfg
amjjbonvin: Indeed - reverted to main version

src/haddock/libs/libplots.py
amjjbonvin: Edited in latest commit

src/haddock/modules/analysis/rnascan/scan.py
amjjbonvin: corrected now - do  check

src/haddock/modules/analysis/rnascan/scan.py
amjjbonvin: corrected now - do  check

src/haddock/modules/analysis/rnascan/__init__.py
amjjbonvin: Done

src/haddock/modules/analysis/rnascan/scan.py
amjjbonvin: Left over - cleaned

src/haddock/modules/analysis/rnascan/scan.py
```diff
@@ -0,0 +1,609 @@
...
+
+            # Library mode: run the mutation jobs through a haddock Engine
+            # (exactly as the module does in Step 2) instead of executing each
+            # job inline, so they go through the normal job lifecycle.
+            log.info(
+                f"Executing {len(self.point_mutations_jobs)} "
+                f"mutations for {self.model_id}"
+            )
+            results = execute_scan_jobs(self.point_mutations_jobs, self.params)
```
rvhonorato: this new addition is worse, it went further nested, its now spinning an engine inside another engine - what needs to be done is to remove it from the `run` completely, not encapsulate into another function.

the whole logic of `# Create mutations` needs to be done in the caller (`__init__.py`):

```python
        scan_objects = [
            InterfaceScanner(
                scan_bases=self.params["scan_bases"],
                model=model,
                params=self.params,
                library_mode=False,
            )
            for model in models
        ]

        log.info(f"Scanning {nmodels} models for possible mutations")
        exec_mode = get_analysis_exec_mode(self.params["mode"])
        Engine = get_engine(exec_mode, self.params)
        engine = Engine(scan_objects)
        engine.run()

        # some more code

        ### LOGIC GOES HERE ###
        # setup the `ModelPointMutation` objects
        # engine = get_engine
        # engine.run
        #########################
```

it needs a re-design. look carefully into the logic and simplify it, its now looking a bit convoluted and im not even sure you need this `library_mode`, better to redo this part.

since you are not doing it yourself, make sure to point the comment above to your AI agent - it should be able to detect the regression and rework this logic.

rvhonorato (changes requested):
looking better, I have a few nitpicks with the code but not blocking issues. the nesting `run` method still needs to be addressed.

src/haddock/modules/analysis/rnascan/scan.py
amjjbonvin: Done - please check

rvhonorato (approved):

Issue #1636, opened 2026-07-31
D-amino acid detection fails if only polar hydrogens are kept


amjjbonvin:
## **Module/Workflow/Library affected**

topoaa

## **Expected behavior**

D-amino acid detection should work. It uses the improper based on atoms HA, N, C CB.

## **Actual behavior**

Because in topoaa, the non-polar hydrogen are deleted before the check for D-amino acid, the CA improper is not defined because the HA atom is missing.

## **Suggestions on how to fix it**

When deleting the non-polar hydrogen (which happens before rebuilding all missing atoms to save time), the HA proton should be kept by default.

## **Version**

80e851f9c5f39f872d2fc39ecc521f9df6d105ba

## **Additional context**

Probably introduced when moving the non-polar hydrogen deleting before the rebuilding to save time.

PR #1637, opened 2026-07-31
Fixes issue #1636 - chirality detection

amjjbonvin:

The `generate_topology.cns` script from the `topoaa` module was modified to always keep the HA proton, even when non-polar hydrogens are deleted (default behaviour). In this way the improper angle used to check the chirality (L/D) is always defined.

Integration test add with a peptide containing two D-amino acids. Checks if the generated PDB contains D-amino acid detection statements.

## How was this tested?

By running: `haddock3-score --outputpsf PPCIVSMLPPDFLYAYS.pdb`

and checking that the generated psf file contains the D-amino acid detection statements:

```
; FILENAME="PPCIVSMLPPDFLYAYS7ozlundp_haddock.psf"
  D-amino acid detected for PRO 1 of chain A
  D-amino acid detected for PRO 9 of chain A
  DATE:31-Jul-2026  10:50:12       created by user: unknown
  VERSION:1.3U
;
```

## AI assistance

To generate the integration test

## Checklist

## Related issues

#1636


AnnaKravchenko:
Testing this PR on my 6 peptides - it’s working! But two issues:

1) Running `haddock3-score` on raw files did preserve correct chirality for all residues, except for a single THR in 1 peptide (“DL”) out of 6. I don’t have a full understanding of it yet, but it seems that because of the bad initial angles in raw peptide, `emscoring` happen to flatten CA in the same plane as HA-C-N-CB, and so while the HA-C-N-CB-improper remains the same +/-3 degrees, CA just happen to **move from D-side to L-side**. I will see if I can fix that using mdref/scoring instead on emscoring - otherwise will show it to you.

2) Another thing is that if there’s no need to rebuild atoms - then chirality check is skipped entirely, beacuse it lives in the `if ( $tobuild > 0 ) then` block of `build-missing.cns`. I saw it when running topoaa with default parameters on `haddock3-score` output D_${peptide}_hs.pdb; when forcing rebuilfing via `delenph=false` - detection is back. 
I don’t know how often "no need to rebuild atoms” occurs, but may-be better to put chirality check outside of the `build-missing.cns`?

amjjbonvin:
> 2\. Another thing is that if there’s no need to rebuild atoms - then chirality check is skipped entirely, beacuse it lives in the `if ( $tobuild > 0 ) then` block of `build-missing.cns`. I saw it when running topoaa with default parameters on `haddock3-score` output D_${peptide}_hs.pdb; when forcing rebuilfing via `delenph=false` - detection is back.
>     I don’t know how often "no need to rebuild atoms” occurs, but may-be better to put chirality check outside of the `build-missing.cns`?

Indeed - now updated (needed to keep one in the tobuild part for proper detection when rebuilding). Added a second one for the case where tobuild = 0 (added else statement)

src/haddock/modules/topology/topoaa/cns/build-missing.cns
```diff
@@ -283,14 +280,50 @@ if ( $tobuild > 0 ) then
...
+            buffer message
+            display D-amino acid detected for $resn $resid of chain $segid
+            end
+        end if
+
+    end loop ltod
+
+end if
 
 fix selection=( none ) end
```
AnnaKravchenko, line 329: This line (329 in new version) is not needed anymore, right? I think it was related to CG minimisation loop (line 294 in old verion), which was jsut deleted?

AnnaKravchenko:
Do you think that maybe there’s a point in keeping HA for the rest of the workflow ONLY for D-residues? 
And so deleting them after D-check in L-residues, smth like: 
" if ($delenph = true) then 
delete sele=(name HA and not [D residues])” ?

amjjbonvin:
> Do you think that maybe there’s a point in keeping HA for the rest of the workflow ONLY for D-residues?
> And so deleting them after D-check in L-residues, smth like:
> " if ($delenph = true) then
> delete sele=(name HA and not [D residues])” ?

Nope - we want all impropers to be defined.

src/haddock/modules/topology/topoaa/cns/build-missing.cns
amjjbonvin: Better to keep it to be safe and make sure nothing is fixed anymore

AnnaKravchenko (approved):
All 6 peptides, both raw and minimised, labelled correctly chirality-wise!
And that single stubborn residue, that was flipped from D to L with emref, remains D with mdref; I am very happy with this:)

AnnaKravchenko:
A bit late, but it just occurred to me that maybe the description of `delenph` in topoaa parameters should be changed to say that HA is always kept for the chirality?

amjjbonvin:
See https://github.com/haddocking/haddock3/pull/1649

Issue #1638, opened 2026-07-31
topocg module unnecessarily sanitise input molecules causing issues with ligands


amjjbonvin:
## **Module/Workflow/Library affected**

topogc

## **Expected behavior**

The module should take a PDB from the topoaa step and use it directly without sanitising it first (unneeded as this happened at the topoaa stage).

## **Actual behavior**

`topocg `currently sanitise the molecule, removing unknown ligands, which causes problems downstream in `cgtoaa`
Since ligands are not supported in the CG implementation, there is no need to provide the ligand top/param files to topocg, but if this is not provided, the sanitisation will remove any unknown ligand.

## **Suggestions on how to fix it**

Remove lines 267-274 from __init__.py in the topocg modules (the lines doing the sanitisation).

## **Version**

80e851f9c5f39f872d2fc39ecc521f9df6d105ba

PR #1639, opened 2026-07-31
Removed the molecule sanitisation and ensemble splitting from topocg

amjjbonvin:

The molecule sanitisation and ensemble splitting steps were removed from topocg as unnecessary (and problematic steps with ligands). topocg receives its molecules from topoaa which has already sanitised and split ensembles.

## How was this tested?

By running the CG workflow on BM5 entry 1AZS which was original causing problems.
Previously the topocg step was editing the PDB in the topoaa directory and removing the ligand.
Now the ligand is correctly preserved and the file in untouched.

## Checklist

## Related issues

#1638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 
+- 2026-07-31: Fixed topocg issue removing ligands - Issue #1638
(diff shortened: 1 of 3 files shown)


VGPReys (approved):
Looks good

PR #1644, opened 2026-08-01
Adding dnascan module - Issue #1634 - and code simplification

### LINKED ISSUE #1634: Create a dnacan module to scan all DNA base-pair mutations at the interface of a complex

A new module that allows to scan all possible mutations of DNA base-pairs at the interface of a complex (similar to what alascan does for proteins).

## **Motivation**

Will allows to study specificity of DNA recognition and possibly useful for design purposes.

## **Description**

Create a new `dnascan` module, following the `rnascan` module structures that will perform mutations of DNA base-pairs, run a short minimisation and create tables/plots of the changes in scores.

Different to RNA, mutations must be performed in pairs for the paired nucleotides as a single mutation would not make sense

amjjbonvin:

This PR add a new analysis module, `dnascan`, similar to the recently added `rnascan` module, but for DNA.
The main difference is that base pairs are mutated, scanning by default all possible combinations.
The mutations are done in two steps, first the purine to pyrimidine one, and then the pyrimidine to purine. This is required to ensure proper rebuilding of the base and base-pairing.

Tests are added and an example is provided in `examples/analysis`

### code simplification

Refactor scan modules: rename scan.py and extract shared code to libscan
Rename each scan module's scan.py to match its module name
(alascan.py, rnascan.py, dnascan.py) and factor the code the modules
duplicated into libscan.py:

- Nucleic-acid mutation helpers shared by rnascan/dnascan: ring/anchor
  atom constants, norm_atom_name, get_atoms_to_keep (parametrized by
  backbone/purines/pyrimidines) and validate_scan_bases (parametrized
  by allowed set).
- filter_interface / build_resname_dict: the identical interface
  user-filtering block used by all three modules' run().
- BaseInterfaceScanner: the shared InterfaceScanner.__init__ setup.
- ModelPointMutation: the identical point-mutation job + scoring flow
  shared by alascan/rnascan; each module keeps a thin run() that
  forwards its module-level mutate/calc_score (kept patchable in tests).

Also apply small cleanups to dnascan (defaultdict, .get, collapsed
branches, RES_CODES literal, cached get_atoms_to_keep per residue,
halved find_base_pairs distance computations) and rnascan (derive
DEFAULT_SCAN_BASES, correct atoms_to_keep annotation, stream file).

dnascan's ModelBasePairMutation (double/two-pass mutation) is genuinely
different and stays in the module. No behavior change; 83 unit tests
across the three modules pass.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

## How was this tested?

Using the newly added example, visually inspecting all mutations to ensure they are correct and the base-pairing is maintained.

## AI assistance

AI was use to build the module from the rnascan one and for generating the tests.
Code and results were manually checked.

## Checklist

- [X] Documentation updated if needed (also in the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)

## Related issues

#1634


examples/analysis/dnascan-test.cfg
```diff
@@ -0,0 +1,29 @@
...
+[topoaa]
+autohis = true
+
+[emscoring]
+
+[dnascan]
+output_mutants = true
+plot = true
+splitplot = false
+int_cutoff = 3.9
```
VGPReys, line 33:
```suggestion
# Output the PDB file containing tested mutations
output_mutants = true
# Generate a plot of the energetics
plot = true
# Do not split the plots for each energetical component
splitplot = false
# Manually define the interaction cutoff used to detect the interface
int_cutoff = 3.9
```

src/haddock/modules/analysis/dnascan/scan.py
```diff
@@ -0,0 +1,1139 @@
...
+
+RES_CODES = dict(
+    [
+        # DNA (two-letter name -> short one-letter code used in identifiers)
+        ("DA", "A"),
+        ("DC", "C"),
+        ("DG", "G"),
+        ("DT", "T"),
+    ]
+)
```
VGPReys: I would place this at the top of the script with all other global variables

src/haddock/modules/analysis/dnascan/scan.py
```diff
@@ -0,0 +1,1139 @@
...
+                add_resname=True,
+            )
+
+            # Detect Watson-Crick base pairs across the whole model
+            bp_cutoff = self.params.get("bp_cutoff", BP_CUTOFF)
+            base_pairs = find_base_pairs(coords, cutoff=bp_cutoff)
+
+            # Determine target nucleotides: get interface, then apply user filters
+            # Get all interface residues
+            cutoff = self.params.get("int_cutoff", 5.0)
```
VGPReys: I do not understand why `BP_CUTOFF` gets a global parameter but not `INT_CUTOFF`

VGPReys: Anyway, they both should always come from the user defined (or default) `self.params`.

src/haddock/modules/analysis/dnascan/scan.py
amjjbonvin: Done

src/haddock/modules/analysis/dnascan/scan.py
amjjbonvin: Updated

AnnaKravchenko (changes requested):
Please add a title to `docs/titles.yaml` and also link this module to the user manual - same that was done for `rnascan`

AnnaKravchenko:
> Please add a title to `docs/titles.yaml` and also link this module to the user manual - same that was done for `rnascan`

As a bonus, adding the title should fix broken link that is blocking merge of RNAscan documentation to the user manual:)

rvhonorato (changes requested):
this has roughly 90% overlap with `rnascan`, meaning its a huge code duplication.

the only functions that seem specific to this module are: `wc_atom`, `is_cross_type`, `find_base_pairs`, `_mutate_residues`, `_score_same_type`/`_score_cross_type`, `bp_cutoff` - everything else that is not boilerplate is repeated.

move the functions used both in `rnascan` and `dnascan` to `libscan.py` rename `scan.py` into `dnascan.py` and keep the specific functions there.

when making this AI changes please get used to making the following prompt: "make sure the contributions done in this PR are in line with AI-POLICY.md and CONTRIBUTING.md guidelines"

if you are using claude there is literally a `/simplify` command you should run before doing a 3.000 line PR full of duplicated code.

AnnaKravchenko:
The table and plot of the final mutatios are tough to read - it’s much more dence compared to protein and RNA. 
I think separating "chain-resid-resname” with dashes makes it easier to read - and more consistent with rnascan and alascan formats. Yes, it is a bit more bulky, but I think the tradeoff is worth it. 

So instead of `B35DT>DA/B4DA>DT` it can be `B-35-DT>DA/B-4-DA>DT`

I would go as far as adoping the same notation for rnascan, the use of “>” is great, it is very readable

src/haddock/modules/analysis/dnascan/dnascan.py
```diff
@@ -0,0 +1,1043 @@
...
+    best_dist: Dict[Tuple[str, int], float] = {key: cutoff for key in keys}
+    best_key: Dict[Tuple[str, int], Optional[Tuple[str, int]]] = {
+        key: None for key in keys
+    }
+    for i, key_i in enumerate(keys):
+        resname_i, xyz_i = nucleotides[key_i]
+        for key_j in keys[i + 1 :]:
+            resname_j, xyz_j = nucleotides[key_j]
+            # Partners must be of complementary ring type (purine <-> pyrimidine)
+            if (resname_i in PURINES) == (resname_j in PURINES):
```
AnnaKravchenko, line 364: Like this we are sure that only canonical base pairs are picked up for scan, and if someting mismatched happen to be in the interface - it is dropped from scannig:

```suggestion
            if COMPLEMENT[resname_i] != resname_j:
```

src/haddock/modules/analysis/dnascan/dnascan.py
```diff
@@ -0,0 +1,1043 @@
...
+            if dist < best_dist[key_i]:
+                best_dist[key_i] = dist
+                best_key[key_i] = key_j
+            if dist < best_dist[key_j]:
+                best_dist[key_j] = dist
+                best_key[key_j] = key_i
+
+    pairs: Dict[Tuple[str, int], Tuple[str, int, str]] = {}
+    for key, partner in best_key.items():
+        if partner is not None:
```
AnnaKravchenko, line 376: To make sure both nuleotides in pair did chose each other, i.e. that pair is symmetrical:
```suggestion
        if partner is not None and best_key[partner] == key:
```

AnnaKravchenko:
Also do you think it would be usefull to add an option to sort output table by various deltas? For dnascan and other two. In a separate PR that it.

AnnaKravchenko (approved):
I am happy with this PR!

PR #1646, opened 2026-08-02
Enabling automatic param/top generation for ligands embedded in a larger system and multiple ligands

### LINKED ISSUE #1645: autotoppar for ligand topo/param generation only works for isolated ligands

The autotoppar option of the `topoaa` module allows to automatically generate top/param files for ligands calling prodrg. This however only works when the ligand is an isolated molecule (which is fine for protein-ligand docking).

It however fails if the ligand, e.g. a co-factor, is part of a protein, because the entire PDB file is passed to prodrg, which causes a failure.  

The code should be change to extract the ligand coordinates and only pass those to prodrg.
It should also be able to handle multiple ligands.

## **Motivation**

This will extends haddock's capabilities to deal with ligands and co-factors.

## **Description**

- extract unknown ligands (only the 1st instance in case of multiple copies) and pass that to prodrg.
- if multiple different ligand are detected, each should be passed to prodrg and the resulting top/param files concatenated
- prodrg has a parameter that allows to define the character used to define the atom types. This should be unique one, not overlapping with what is in the cofactor.top file (which was created using prodrg) to avoid overlap in parameters definitions.

amjjbonvin:

topoaa's autotoppar path calls libligand.run_prodrg(model, ...) on the entire PDB. When an unknown ligand is embedded in a larger system (protein/ligand complex, or e.g. a protein in a membrane with many lipid copies), PRODRG received the whole system and failed to generate topology/parameter files.

`src/haddock/libs/libligand.py` was modified to deal with this cases.

  1. New extract_ligand() — strips out everything except the requested ligand residues, and keeps only the first copy of each residue name (a ligand present in N copies, e.g. 434× DPP, shares one topology, so only one copy is sent to PRODRG).
  2. Refactored run_prodrg() — now accepts ligand_resnames:
    - Extracted the single-molecule PRODRG execution into a _run_prodrg_single() helper.
    - When multiple distinct ligands are present (e.g. DPP + DPC), PRODRG runs once per ligand and the resulting .top files are concatenated into one, and the .param files into one.
    - De-duplicates resnames so a repeated ligand triggers a single run.
    - With ligand_resnames=None, behavior is unchanged (whole file → PRODRG).
  3. topoaa/__init__.py — passes the already-computed unknown resnames into run_prodrg(..., ligand_resnames=unknown).
  4. to avoid overlap in parameters, each ligand parametrised receives a unique one character identifier (also not overlapping with what is used in `cofactors.top`. This uses the CNSSEP option of PRODRG
  5. modified the haddock3-score cli to have autotoppar=true by default in order to be able to handle ligands.
  Verification

  - Added tests in tests/test_libligand.py (extraction, single-copy, complex success, multi-ligand concatenation, dedup) — 22 passed, ruff clean, topoaa tests still pass.
  - CHANGELOG.md updated.

## How was this tested?

Tested using examples of proteins contain a ligand (e.g. 1AZS_l_u.pdb from the BM5 benchmark).

## AI assistance

Claude was used to modify the ligand handling machinery and add tests.

## Checklist

## Related issues

#1645

## Notes for reviewers

You can test it using for example haddock3-score


chietake:
In v2026.8.0, I encountered exactly the scenario covered by `test_run_prodrg_on_complex_with_ligand_resnames`.
My use case is scoring/refinement of protein–ligand complexes predicted by Boltz or AF3, rather than docking protein and ligand structures supplied separately.
I plan to test this PR on some of these complexes using `haddock3-score` and will report back if I encounter any issues.

chietake:
I tested this PR with several pre-assembled protein–ligand complexes predicted by Boltz using `haddock3-score`.
In each case, the ligand was detected automatically, PRODRG generated the topology/parameter files during `topoaa`, and `emscoring` completed successfully.
Thanks!

Issue #1647, opened 2026-08-02
DEBUG logging info leaking even when debug is not set


amjjbonvin:
## **Module/Workflow/Library affected**

libio and some others.

## **Expected behavior**

DEBUG statements should only appear in debug mode

## **Actual behavior**

Some libraries still log DEBUG statements even when the debug flag is not set.

## **Steps to reproduce the behavior**

Run for example the docking-protein-protein-test example and look at the log file.
Note that this behaviour seems OS-dependent. On OSX DEBUG statements are showing.

## **Version**

80e851f9c5f39f872d2fc39ecc521f9df6d105ba## **Additional context**

PR #1648, opened 2026-08-02
Fix DEBUG logging leak and global warning suppression - Issue #1647

amjjbonvin:

DEBUG log records were leaking to the console even when the log level was not set to DEBUG (e.g. `[... libalign DEBUG] Sequence identity ...`, `[... libio DEBUG] removing: ...`).

Root cause: `haddock/__init__.py` sets the shared `haddock` logger to `DEBUG` and attaches a stdout `StreamHandler` **without a level**, so the handler emitted everything down to DEBUG. The main CLI hides this because `add_log_for_CLI()` clears and re-adds properly-levelled handlers, but any code path that never calls it — notably parallel worker subprocesses that re-import `haddock` — keeps the wide-open default handler and dumps DEBUG to stdout.

Fixes in this PR:

- **`haddock/__init__.py`**: cap the default stdout handler at `INFO`. The logger stays at `DEBUG` so `--log-level DEBUG` still works once `add_log_for_CLI()` reconfigures the handlers.
- **`libs/libaa2cg.py`**: the module ran `warnings.filterwarnings("ignore")` at import time, silencing **every** Python warning (`DeprecationWarning`, numpy `RuntimeWarning`, ...) process-wide as a side effect of import. Because `libaa2cg` is imported by `caprieval`/`caprifilter`/`topocg`, this affected most runs. Scoped the filter to `BiopythonWarning` (its original intent — quieting Biopython's PDB-parsing noise), matching the targeted pattern already used in `libgrid.py`.
- **`libs/libaa2cg.py`**: replaced a bare `except:` with `except Exception:` in the DSSP secondary-structure assignment (ruff E722); it no longer swallows `KeyboardInterrupt`/`SystemExit`.

## How was this tested?

- Verified the default handler now suppresses DEBUG while still showing INFO/WARNING/ERROR, and that `add_log_for_CLI(log, "DEBUG", ...)` still emits DEBUG when explicitly requested.
- Verified the scoped warnings filter suppresses `BiopythonWarning` while `DeprecationWarning`/`RuntimeWarning` are again visible.
- `ruff check` passes on the changed files; the affected unit tests pass (`pytest tests/ -k "aa2cg or capri or topocg or log"` → 95 passed).

## AI assistance

The leak was diagnosed and the fixes drafted with help from Claude. The code changes and verification steps were manually reviewed.

## Checklist

- [ ] Tests cover the new and/or changed code

## Related issues

#1647

## Notes for reviewers

No new tests were added — the changes are to import-time logging/warning state, which the existing suite already exercises. All three fixes share the same underlying anti-pattern: a module mutating global logging/warning state as an import side effect.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Comp-era (dismissed):
Looks good. Handler capped at INFO with the logger still at DEBUG (no leak — verified via handler inspection), add_log_for_CLI(log, 'DEBUG', ...) still emits DEBUG on request, and libaa2cg no longer mutes warnings globally. Approving.
Only the trivial CHANGELOG.md conflict is left to resolve.

Comp-era (approved):
Re-approving — 0f80dc6 just merges main and resolves the CHANGELOG conflict; the logging/warning fix is unchanged from my earlier tested review.

Issue #1530, opened 2026-04-21
refactor modularity-breaking implementations


rvhonorato:
Some modules were implementing following an approach that breaks modularity and adds a  dependency between them. 

For example for `clustrmsd` to be executed, `ilrmsdmatrix` or `rmsdmatrix` MUST have been executed before.

In the other hand `clustfcc` does not require anything to be executed before and thus follow proper modular implementation.

## List of modules that need to be refactored
- [ ] `clustrmsd` 
- [ ] `ilrmsdmatrix`
- [ ] `rmsdmatrix`

---

This can greatly benefit from the caching implementation at #1529

rvhonorato:
setting as `planned-long-term` because its held by #1528

rvhonorato:
as of today there are a growing number of decisions being made that break the initial modular design the project was conceptualized for. on top of the issues I listed above I have also identified a number of incompatibilities between the modules that are not properly described nor checked during the workflow constructions. I won't have more time to work on this issue, so I'm removing myself from the task but keeping it open as both a reminder and in case someone wants to pick it up if it remains relevant. if you need my input feel free to @ me and I'll reply to this issue.

PR #1651, opened 2026-08-04
Add workflow module ordering validation

amjjbonvin:

HADDOCK3 validates that each workflow module and its parameters are valid, but it does **not** check whether the *sequence* of modules makes sense. Some modules only work when another ran before them, some must not be chained to themselves, and a workflow must start by building topologies. Until now, such mistakes only surfaced at run time (or produced meaningless results).

This PR adds a data-driven workflow ordering check. Constraints are declared in `src/haddock/gear/workflow_rules.yaml` and enforced by `src/haddock/gear/workflow_ordering.py`, so new rules can be added without touching the validation logic. Four rule types are supported:

- **`disallowed_sequences`** — a module may not directly follow another (e.g. `clustrmsd → clustrmsd`, `clustfcc → clustfcc`, `seletop → seletop`, `seletopclusts → seletopclusts`).
- **`required_preceding`** — a module must be *immediately* preceded by one of a set (`topocg` after `topoaa`; `clustrmsd` after `rmsdmatrix`/`ilrmsdmatrix`).
- **`required_prior`** — a module must appear *somewhere earlier* (`cgtoaa` requires `topoaa`).
- **`required_first`** — the first module must be one of a set (`topoaa`).

The check is wired into `setup_run()` and raises a single `ConfigurationError` listing **all** violations (with step numbers and allowed alternatives). It is skipped for `--extend-run`, whose config only describes the appended modules, so the full order is not known there.

The `plot-finetune-clustfcc.cfg` example was adjusted so its `clustfcc` steps are no longer directly consecutive, matching the `clustfcc → clustfcc` rule.

## How was this tested?

- New unit tests in `tests/test_gear_workflow_ordering.py` (**100% coverage** of `workflow_ordering.py`) covering each rule type, multi-violation reporting, empty workflows, and custom rules files.
- Ran the check against all 110 real example workflows: only the two `docking-extend-run-*.cfg` configs "fail" standalone, and those are exclusively run with `--extend-run` (where the check is skipped by design).
- `ruff format --check` and `ruff check` pass on changed files; `pytest tests/test_gear_workflow_ordering.py tests/test_gear_prepare_run.py` → 93 passed.

## AI assistance

Drafted with Claude Code and reviewed manually. Claude wrote the validation logic, YAML rules, tests, and docs; verified the behaviour by running the full test suite and cross-checking the rules against every example workflow in the repo. This is config/ordering validation only — no scoring, energy, restraint, or CNS/topology logic is involved.

## Checklist

- [ ] Documentation updated if needed (also in the [haddock3 user-manual](https://github.com/haddocking/haddock3-user-manual)) — code docs updated (`architecture.md`, docstrings); user-manual update still to do

## Related issues

Related to #1530.

## Notes for reviewers

- Rules live in `gear/workflow_rules.yaml` — reviewers should sanity-check the rule set for scientific correctness (e.g. whether any other module pairs should be disallowed/required).
- The user-manual (separate repo) is not yet updated.

### DIFF
--- a/src/haddock/gear/workflow_rules.yaml
+++ b/src/haddock/gear/workflow_rules.yaml
@@ -1,0 +1,52 @@
+# =============================================================================
+# HADDOCK3 workflow ordering rules
+# =============================================================================
+#
+# Individual module- and parameter-level validation (see gear/prepare_run.py)
+# confirms that each requested module exists and that its parameters are valid.
+# It does *not* enforce constraints on the *sequence* in which modules appear.
+#
+# This file declares those sequence constraints as data, so that they can be
+# reviewed and extended without touching the validation logic in
+# `gear/workflow_ordering.py`. Module names below are the module folder names
+# (e.g. `topoaa`, `clustrmsd`), i.e. the keys used in a workflow config file
+# without their `.N` occurrence suffix.
+#
+# Four rule types are supported:
+#
+#   disallowed_sequences  a pair [A, B]: module B must NOT come directly
+#                         after module A (guards against meaningless repeats).
+#
+#   required_preceding    B: [A1, A2, ...]: module B must be immediately
+#                         preceded by one of A1, A2, ...
+#
+#   required_prior        B: [A1, A2, ...]: module B must be preceded
+#                         somewhere earlier in the workflow (not necessarily
+#                         directly) by at least one of A1, A2, ...
+#
+#   required_first        [M1, M2, ...]: the very first module of a workflow
+#                         must be one of M1, M2, ...
+#
+# All checks are skipped for entries whose "trigger" module is not present in
+# the workflow, so adding a rule never breaks unrelated workflows.
+# -----------------------------------------------------------------------------
+
+# Module B must not directly follow module A. Each entry is a [A, B] pair.
+disallowed_sequences:
+  - [clustrmsd, clustrmsd]
+  - [clustfcc, clustfcc]
+  - [seletop, seletop]
+  - [seletopclusts, seletopclusts]
+
+# Module (key) must be immediately preceded by one of the listed modules.
+required_preceding:
+  clustrmsd: [rmsdmatrix, ilrmsdmatrix]
+  topocg: [topoaa]
+
+# Module (key) must be preceded anywhere earlier by one of the listed modules.
+required_prior:
+  cgtoaa: [topoaa]
+
+# The first module of the workflow must be one of these.
+required_first:
+  - topoaa
--- a/src/haddock/gear/workflow_rules.yaml
+++ b/src/haddock/gear/workflow_rules.yaml
@@ -45,7 +45,7 @@ required_preceding:
 
 # Module (key) must be preceded anywhere earlier by one of the listed modules.
 required_prior:
-  cgtoaa: [topoaa]
+  cgtoaa: [topocg]
 
 # The first module of the workflow must be one of these.
 required_first:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 
+- 2026-08-04: Added workflow module ordering validation (disallowed sequences, required preceding/prior modules, required first module) driven by `gear/workflow_rules.yaml`
 - 2026-07-31: Fixed D-amino acid detection - Issue #1636
 - 2026-07-31: Fixed topocg issue removing ligands - Issue #1638
 - 2026-07-27: Added `rnascan` module for mutagenesis scanning of RNA bases (mutating interface nucleotides to A, C, G, U) - Issue #1631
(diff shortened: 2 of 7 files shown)


AnnaKravchenko:
At a glance this goes agains the [fine tuning of the clustfcc parameters](https://github.com/haddocking/haddock3/blob/main/examples/analysis/plot-finetune-clustfcc.cfg) from the user manual  [here, bottom of the page](https://www.bonvinlab.org/haddock3-user-manual/docking_scenarios/analyses.html)

amjjbonvin:
> At a glance this goes agains the [fine tuning of the clustfcc parameters](https://github.com/haddocking/haddock3/blob/main/examples/analysis/plot-finetune-clustfcc.cfg) from the user manual [here, bottom of the page](https://www.bonvinlab.org/haddock3-user-manual/docking_scenarios/analyses.html)

Indeed - this example was updated (but not the manual). This is intended to avoid similar problems as with clustrmsd.

But as said in the PR, the list of rules need to be checked.

src/haddock/gear/workflow_rules.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - [seletopclusts, seletopclusts]
+
+# Module (key) must be immediately preceded by one of the listed modules.
+required_preceding:
+  clustrmsd: [rmsdmatrix, ilrmsdmatrix]
+  topocg: [topoaa]
+
+# Module (key) must be preceded anywhere earlier by one of the listed modules.
+required_prior:
+  cgtoaa: [topoaa]
```
VGPReys:
```suggestion
  cgtoaa: [topocg]
```

src/haddock/gear/workflow_rules.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+# The first module of the workflow must be one of these.
+required_first:
+  - topoaa
```
VGPReys, line 52: simple comment / open question (after 3 year only): Why don't we just run `[topoaa]` by default at the start if it is required to be the first module ?

amjjbonvin: Because there are some options that can be given to it.

amjjbonvin: One of which is to provide your own param/top files for a ligand.  So it should remain exposed.

VGPReys (approved):

rvhonorato:
My two cents about this:

There is a creeping design problem in this codebase in which badly implemented modules break the modularity; in its current state its even arguable if this code is truly modular or if its transforming into a framework.

Given the problem that: some modules are breaking the modularity design, its unfortunate that instead of addressing the cause issue and re-working the bad implementations to uphold the true modular design, you went with this AI-generated low-effort solution to simply add a barrier in the pain points.

This is functional, no question about that, but considering this was offloaded to an AI agent that could as well have come up with a solution that would keep the project cohesive and modular, it is disappointing to see it going in this direction of patching the superficial issues instead of addressing the root causes.

I really hope that in the future you keep the modularity in mind, I know its not simple making the implementations modular, but a lot of groundwork was laid to make this possible and seeing this PR merged is a step backwards. 

Nothing to be done now but please consider the project's idea more carefully, it's not about code implementations, problems can be solved in multiple ways, in the same way that you decided to add a check, you could have fixed the cause issues, so its not really about code its about following the design.

src/haddock/gear/workflow_rules.yaml
rvhonorato: > Because there are some options that can be given to it.

this is not the reason, options could be passed to it without it being defined as a module - @VGPReys this module exist as an historical artifact, the very initial pre-alpha version had this logic of topology>rigidbody>selection>refinement (same as the legacy haddock version) and this was just propagated until now. so the reason it exists is because it always existed, but could indeed be absorbed into the initialization and not needed to be explicit.

amjjbonvin: No so simple. If you want to control the protonation state of say 10 histamine you will have to pass many options... Won't be user friendly. Same thing for providing custom param files. Now all this is defined in the config file (better for reproducibility). If you have to pass many options every time you call haddock3 it will be messy ("a la rosetta" with incredibly long command lines).  

But in the future we could lift the requirement a workflow starts with it.

amjjbonvin:
Full modularity is not simple. Take the cgtoaa module. If there is not prior topoaa and topocg steps it won't work.

rvhonorato:
> Take the cgtoaa module. If there is not prior topoaa and topocg steps it won't work.

exactly, this is a good example of a bad implementation that break modularity. 

as I said above, taking this pathway of bad implementations + superficial solutions is functional and it's also disappointing considering all the potential the modularity had to offer by putting more effort into fixing root causes, but anyway users will benefit from the code regardless and that is what matters in the end (:

amjjbonvin:
it’s not a bad implementation … There are simply things you can do and not do if you want to follow a given recipe using some specific tools (e.g. CNS within HADDOCK3)

It is not only about software design, but about what you can do with it.

amjjbonvin:
Last message was deleted because it violates our CODE_OF_CONDUCT rules

PR #1656, opened 2026-08-07
Add .dockerignore to reduce Docker build context

chernojagne:
Add a `.dockerignore` file so that `docker build` only copies what the image actually needs.

Without `.dockerignore`, every `COPY . .` in the Dockerfile sends the full working tree to the Docker daemon — including `.git/`, test suites, docs, IDE config, and build artifacts. This bloats the build context (and therefore build time and layer size) unnecessarily.

The file mirrors patterns already present in `.gitignore` (byte-compiled files, virtual envs, dist artefacts, notebooks checkpoints, etc.) and adds Docker-specific exclusions for development tooling, CI config, documentation, and test directories that have no place in a production image.

No Python source files are touched, so CI only triggers the `docker-build` job.

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,216 @@
+# Mirrored from .gitignore
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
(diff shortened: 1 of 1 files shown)


rvhonorato:
please see [AI-POLICY](https://github.com/haddocking/haddock3/blob/main/AI-POLICY.md)

PR #1661, opened 2026-08-08
fix dna nucleobase atom selection default for capri

### LINKED ISSUE #1658: CAPRI eval uses different atom sets for DNA in AA and CG mode

caprieval

## **Expected behavior**

CAPRI stats in case of DNA should be calculated in a consistent manner using heavy atoms (beads) of the bases (and not the backbone as it can be more flexible/variable).

## **Actual behavior**

Currently CAPRI statistics are calculated using base atoms in case of DNA in all atoms mode, but backbone beads in CG mode, which is inconsistent.

See the DNA_ATOM and DNA_ATOM_CG definitions in `libalign.py`

Replace the backbone beads by based beads for DNA_ATOM_CG

## **Version**

1f54a9e53aa6dbcf29960945b90a8fbb67ae4cb6

VGPReys:

This PR modifies the default CG atom selection during `caprieval` to select the nucelo-bases for DNA (l.316), matching the all-atom selection.

## AI assistance

Nope

## Checklist

- [ ] Tests cover the new and/or changed code

## Related issues

Closes #1658

## Notes for reviewers

Sorry, the automatic linting program did not allow me to touch only the appropriate part, and I had to change the linting all along the script...


amjjbonvin (dismissed):

AnnaKravchenko:
Something is not right with definition of U nucleotide (in all-atom): 
`[2026-08-10 12:01:15,701 libparallel WARNING] Exception in task execution: sequence contains letters not in the alphabet`

And then caprieval crushed the entire run

amjjbonvin:
Do we have any RNA-related tests/examples? If not probably good to add

AnnaKravchenko:
Technically that is separate issue that is not related to CG atom selection for DNA in capri, but while reading code in this PR, I spotted that in libalign, line 516 (definition of C in RNA_FULL_DICT): 
- O2 is missing; 
- N9, O6, N2 should not be there

AnnaKravchenko:
Another potential issue is that DNA_FULL_DICT lists “O1P” and “O2P”, while haddock writes pdbs using “OP1” and “OP2”. I don’t think there’s any way to match those atoms within libalign or caprieval?  
So they are probably just discarted when calculating all capri metrics with `allatoms=true` in `caprieval`

src/haddock/libs/libalign.py
```diff
@@ -292,8 +379,8 @@
     ],
     "DC": [
         "P",
-        "O1P",
-        "O2P",
+        "OP1",
```
AnnaKravchenko, line 382: This is nice, but I initially did not think about the reference file. If that one is not being produced by haddock -  then it can, in principle, contain O1P/O2P.

So maybe it’s better to keep all 4 options, same as `scan.py` in main is doing?

src/haddock/libs/libalign.py
```diff
@@ -436,13 +526,11 @@
         "C2'",
         "O2'",
         "C1'",
-        "N9",
         "C5",
-        "C6",
         "O6",
```
AnnaKravchenko, line 530: RNA_FULL_DICT[“C”] has “O6” - should be “C6”: 

```suggestion
        "C6",
```

amjjbonvin:
You are introducing wrong quotes… beware!

src/haddock/libs/libalign.py
AnnaKravchenko: I now copied quotes from the code directly - seems they are now not affected by this suggestion, but indeed beware!

AnnaKravchenko:
One more thing - can you please update caprieval/defaults.yaml of ‘allatoms’ to say that backbone is the case for proteins and RNAs, but it’s nucleobases for DNAs?

PR #1684, opened 2026-09-02
Add offset to seed per each sampling_factor replica 

### LINKED ISSUE #1685: Fixed seed per sampling_factor replica
If mdref, flexref or emref get input model with not Nan seed, this seed remains the same per each sampling_factor replica, and so all
n models ( sampling_factor = n ) with the same staring point end up as identiacl model. Effectivelly sampling_factor is not functional.  

flexref, mdref and emref - only when models already have seeds 

## **Expected behavior**
Different seeds, so that refinement produces different models from the same starting point.

## **Actual behavior**
Exactly the same coordinates of emref_1.pdb and emref_2.pdb with sampling_factor = 2, if rigidbody was executed before emref.

1. Go to the examples/docking-antibody-antigen/docking-antibody-antigen-CDR-NMR-CSP-test.cfg
2. Keep general parameters, topoaa, rigidbody and emref  
2. set sampling = 2 in rigidbody and sampling_factor = 2 in emref
3. Run in, gunzip emref models, and run ‘diff emref_1.pdb emref_2.pdb” - their coordinates are the same

cgtoaa fixes it by adding an offset to the seed

AnnaKravchenko:

Added offset to the seed (if seed is not Nan - otherwise random seed will be assigned later) for refinement job input. if sampling_factor set to 2, cns will get 2 identical staring models with seeds 918 and 919, so the resulting refined models with the same starting point will be different. 

## How was this tested?
Edited examples/docking-antibody-antigen/docking-antibody-antigen-CDR-NMR-CSP-test.cfg to [topoaa]->[rididbody]->[emref] with rididbody sampling = 2 and emref sampling_factor = 2. On main branch emref_1.pdb and emref_2.pdb are identical, on this branch - not anymore. (better description is the isse #1685 )

## Notes for reviewers
Copied this offset implementation from cgtoaa.


AnnaEngel98 (approved):
Works for me on all refinement modules

PR #1686, opened 2026-09-03
Caching stage1

### LINKED ISSUE #1689: We want caching!

Currently there is no caching in haddock3 runs. In the situation that a run gets interrupted and thus a module does not complete, the module will re-run from scratch when the run is restarted, with caching, the module will continue where it was

## **Motivation**

Saves a lot of time!

sjdv1982:

This is the first stage of the caching feature. The purpose is to fix several issues that make HADDOCK less reproducible (and therefore less cacheable) than it could be. For example, if we increase the sampling from 100 to 101, we want the first 100 structures to stay the same: as it is, the first 100 structures are often different because of different seeds.

It consists of 14 individual commits, each with a long commit message that describe what is being changed.

The first commit contains a design plan, essentially the long version of the description of the 4 stages.

## How was this tested?
Most commits add their own tests. The full test set and integration test set passes.

## AI assistance
AI (Claude Opus) wrote all the commit messages and AI (Codex and Claude) did the coding. The code changes were triggered by reproducibility errors identified by the AI when implementing stage 3 and testing it on stage 2.

## Checklist

- [ ] `CHANGELOG.md` updated for user-facing changes

## Notes for reviewers
This is preparatory work on the HADDOCK internals: user-facing work is in stage 2-3. See https://github.com/haddocking/haddock3/compare/caching-stage1?expand=1#diff-22967752f3bf7f43d4c2f7e748d12c8b2ff61c340e6fe499ff45023de5d8acba fot details.


sjdv1982:
Commit https://github.com/haddocking/haddock3/commit/4b066c8dc316593bbc5517ddb2a3e6ad9a5312c7 addresses (a subset of) the same issue as PR #1660 . Claude says that the PR #1660 code is strictly better, so 4b066c should be dropped from here.

